PR #82, opened 2021-10-15
Implement `Option<T>`

Desdaemon:
Rough implementation of `Option<T>` to Dart's `T?`.
I based this on the implementation of `Box<T>` where a null is equivalent to passing a nullptr down the wire.

### DIFF
--- a/frb_codegen/src/api_types.rs
+++ b/frb_codegen/src/api_types.rs
@@ -307,8 +307,10 @@ impl ApiTypeChild for ApiTypePrimitive {
 }
 
 impl ApiTypePrimitive {
-    /// Primitive pointers.
-    pub fn dart_marker_type(&self) -> &'static str {
+    /// Representations of primitives within Dart's pointers, e.g. `ffi.Pointer<ffi.Uint8>`.
+    /// This is enforced on Dart's side, and should be used instead of `dart_wire_type`
+    /// whenever primitives are put behind a pointer.
+    pub fn dart_native_type(&self) -> &'static str {
         match self {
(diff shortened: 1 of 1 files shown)


fzyzcjy:
Thank you!

fzyzcjy:
By the way, could you please add a line to CHANGELOG.md (only the one in the repo root)? I find I cannot do that on the PR.

Desdaemon:
@fzyzcjy I would like some advice on boxing primitives. I was thinking that this could go in two ways:
- We don't implement Options for primitives, and all wire API calls will only involve plain primitives.
- We allow Options of primitives, under the assumption that all Dart callers have to allocate memory for the primitive.
This could give some freedom in type design, but personally I don't have a use case for this. 

Perhaps this behavior could be defined by the user, which I think is the best compromise for now.

fzyzcjy:
> I was completely unaware of this solution, thank you for bringing it up! I think this will help out a lot.

You are welcome!

> Basically I was thinking that because the wire type is a pointer, it has to point to some underlying memory
and that being the case makes boxing primitives from Rust's side easier to deal with.

IMHO maybe no. For example, if wire type is `int32_t* value`, it can be null pointer, so it does not point to anywhere.

fzyzcjy:
> btw ignore the master's code update for a while (and do not merge into this pr). I am developing a new feature and will push some commits later.

features added. you may merge upstream now (or later, of course both ok)

fzyzcjy:
let me have a review

fzyzcjy:
Really great job! Only some code review above asking minor questions.

fzyzcjy:
@all-contributors please add @Desdaemon for code, test

fzyzcjy:
Great job! Let me have a review

PR #119, opened 2021-10-23
fix windows path handling

smw-wagnerma:
Well, quite tricky, my changes:

- to get the error message, I had to switch to stderr
- The path problem was, that the cononical functions on Windows get UNC [paths](https://github.com/rust-lang/rust/issues/42869), that the CMD can't handle
- after this, I got the next trouble, ffigen can't find the installed LLVM, so I have to introduce another llvm_path parameter

With these changes I can run the command:

`[2021-10-23T19:57:10Z INFO  flutter_rust_bridge_codegen] Success! Now go and use it :)`

My next problem ist, the generated bridge_generated dart and rust file contains error, this is another investigation...

### DIFF
--- a/frb_codegen/src/commands.rs
+++ b/frb_codegen/src/commands.rs
@@ -46,14 +46,20 @@ fn execute_command(arg: &str, current_dir: Option<&str>) {
     let result = cmd.output().unwrap();
 
     if result.status.success() {
-        debug!("command={:?} output={:?}", cmd, result);
+        debug!(
+            "command={:?} stdout={} stderr={}",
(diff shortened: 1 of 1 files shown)


fzyzcjy:
The CI seems to fail. Could you please check it? You may find this helpful: https://github.com/fzyzcjy/flutter_rust_bridge/blob/rust-command/CONTRIBUTING.md

fzyzcjy:
> My next problem ist, the generated bridge_generated dart and rust file contains error, this is another investigation...

You can create a new issue and I will also check it

PR #182, opened 2021-11-05
Copy Rust comments over to Dart generated file

Desdaemon:
Rust block comments are converted to Dart doc comments due to an issue with `dart format` needing to be run twice to get correct output. Only comments for structs, fields and functions are copied for now.


fzyzcjy:
Thanks! Let me have a look

fzyzcjy:
Just a tiny change and LGTM: We generate an abstract class `abstract class FlutterRustBridgeExample extends FlutterRustBridgeBase<FlutterRustBridgeExampleWire>`

https://github.com/fzyzcjy/flutter_rust_bridge/blob/81075427832dba15d06128a7b1ea3a6148761e33/frb_example/pure_dart/dart/lib/bridge_generated.dart#L11

, and a concrete implementation:

https://github.com/fzyzcjy/flutter_rust_bridge/blob/81075427832dba15d06128a7b1ea3a6148761e33/frb_example/pure_dart/dart/lib/bridge_generated.dart#L251

Maybe we should put the comments to the *abstract class* (that is to be used by humans) instead of the implementation (which normal humans do not need to care about).

fzyzcjy:
image

you may not be worried about this, since another iphone passes test. just commit and see whether tests pass the next time.

fzyzcjy:
🎉

fzyzcjy:
Let me release a new version.

fzyzcjy:
done

PR #186, opened 2021-11-08
Update README for trouble shooting.

bus710:
I was trying to reproduce the workflow to understand how it can be done from scratch.

I did cargo build first and then flutter_rust_bridge_codegen. Of course I got the error - undefined symbol: store_dart_post_cobject.

Finally I realized that the order is the other way around. The codegen should be done first and then cargo build (or make) so that the rust compiler can find the generated rust file and includes it to be a part of the end result.

I know this is solely my mistake, but I want to help others like me :eyeglasses:


fzyzcjy:
Thanks! @all-contributors bot please add @bus710 for doc

Issue #191, opened 2021-11-09
Implement `Vec<String>`


Desdaemon:
**Is your feature request related to a problem? Please describe.**

I came across this hiccup when designing my API, I think this is easy enough yet has high impact and should be implemented soon.

**Describe the solution you'd like**

Make `wire_list_String` hold a `*mut *mut wire_list_uint_8` and implement the missing methods (api_fill).

**Describe alternatives you've considered**

I haven't tried, but a `Vec<Vec<u8>>` might work but not very ergonomically.

I'm a bit busy right now, so I might not make a PR right away but I will eventually come around to it.

fzyzcjy:
Sounds wonderful and looking forward to the PR!

To add my two cents, maybe we can also write down a new `Delegate` type named `ListString` and manually do the things.

PR #193, opened 2021-11-11
Implement `Vec<String>`

Desdaemon:
Solves #191.

~~Blocker: Dart generator is creating unused code, I'm trying to find a workaround before declaring this mergeable.~~ Done!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -205,6 +205,7 @@ jobs:
       - name: Run Flutter integration tests
         working-directory: ./frb_example/with_flutter
         run: flutter test -d windows integration_test/main.dart --verbose
+
   flutter_linux_test:
     # CI on Linux does not work right now due to log reader errors
     # if: false
@@ -220,12 +221,13 @@ jobs:
           profile: minimal
           toolchain: stable
 
+      # From Flutter Gallery CI
       - uses: subosito/flutter-action@4389e6cbc6cb8a4b18c628ff96ff90be0e926aa8
         with:
           channel: master
 
       - name: Install Flutter build tools
-        run: sudo apt-get -y install ninja-build
+        run: sudo apt-get -y install clang cmake ninja-build pkg-config libgtk-3-dev libglu1-mesa
 
       - run: flutter config --enable-linux-desktop
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -207,8 +207,8 @@ jobs:
         run: flutter test -d windows integration_test/main.dart --verbose
 
   flutter_linux_test:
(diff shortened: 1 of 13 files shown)


fzyzcjy:
> Blocker: Dart generator is creating unused code, I'm trying to find a workaround before declaring this mergeable. Done!

Congrats!

frb_codegen/src/generator_dart.rs
```diff
@@ -361,7 +361,7 @@ fn generate_wire2api_func(ty: &ApiType, api_file: &ApiFile) -> String {
             }
         },
         Optional(opt) => format!(
-            "return raw == null ? null : _wire2api_{}(raw);",
+            "if (raw != null) return _wire2api_{}(raw);",
```
fzyzcjy, line 363: What about keeping the original code: it is explicit about what it returns and is thus more readable especially for new Dart users. Btw if you want to keep this change, maybe put it in a new PR, which will make the diff of PRs more readable.

.github/workflows/ci.yaml
```diff
@@ -207,27 +207,25 @@ jobs:
         run: flutter test -d windows integration_test/main.dart --verbose
   flutter_linux_test:
     # CI on Linux does not work right now due to log reader errors
-    if: false
```
fzyzcjy, line 210: After CI is ok I will merge it

frb_example/pure_dart/dart/lib/main.dart
```diff
@@ -97,6 +97,11 @@ void main(List<String> args) async {
       expect(listOfStructResp[3].width, 420);
     }
 
+    print('dart call handleStringList');
+    {
+      await api.handleStringList(names: ['Steve', 'Bob', 'Alex']);
```
fzyzcjy, line 102: What about making a very simple assert to its return value, e.g. `expect(..., ...)`

frb_example/pure_dart/dart/pubspec.lock
```diff
@@ -5,84 +5,84 @@ packages:
     dependency: transitive
     description:
       name: _fe_analyzer_shared
-      url: "https://pub.flutter-io.cn"
```
fzyzcjy, line 8: Agree. I should have done that

fzyzcjy:
Thank you for the PR! Only a few tiny suggestions and I am ready for merging ;)

frb_codegen/src/generator_dart.rs
Desdaemon: Good point. I think I will revert this.

fzyzcjy:
Just ping me when you are done and I will merge it

Desdaemon:
I re-disabled Linux integration tests for now so please merge after this run @fzyzcjy

Issue #234, opened 2021-12-01
Returning u32 from Rust results in wrong value received by Dart


surban:
**Describe the bug**
Returning u32 from Rust results in wrong values.

**To Reproduce**
Add the following to `frb_example/pure_dart/rust/src/api.rs`:
```
pub fn primitive_u32(my_u32: u32) -> Result<u32> {
    println!("primitive_u32({})", my_u32);
    assert_eq!(my_u32, 0xff112233);
    let ret = 0xfe112233;
    println!("returning {}", ret);
    Ok(ret)
}
```

And the following to `frb_example/pure_dart/dart/lib/main.dart`:
```
    print('dart call primitiveU32');
    {
      expect(await api.primitiveU32(myU32: 0xff112233), 0xfe112233);
    }
```

The test fails with the following output:
```
dart call primitiveU32

primitive_u32(4279312947)
returning 4262535731
00:00 +0 -1: main test [E]

  Expected: <4262535731>
    Actual: <-32431565>
```

fzyzcjy:
Thank you!

fzyzcjy:
Sounds like a bug in upstream. https://github.com/sunshine-protocol/allo-isolate/pull/17

fzyzcjy:
https://github.com/sunshine-protocol/allo-isolate/pull/17 is merged. let me update this lib.

fzyzcjy:
https://github.com/fzyzcjy/flutter_rust_bridge/runs/4394097901?check_suite_focus=true

valgrind tests passed. seems that it is fixed (by my pr to upstream)

fzyzcjy:
fixed https://github.com/fzyzcjy/flutter_rust_bridge/releases/tag/v1.8.2

PR #235, opened 2021-12-01
Add test for passing and returning u32.

surban:
The test currently fails and demonstrates bug #234.


fzyzcjy:
Thank you! I will merge once CI passes

PR #261, opened 2021-12-30
Marker attributes

### LINKED ISSUE #260: Marker attributes
As this project is growing in functionality, there are some places where Rust is not expressive enough to generate quality Dart code.
The idea is that bridging libraries e.g. wasm_bindgen also use attributes to guide code generation, for example in this case `#[wasm_bindgen(...)]`.

A simple use-case for our library would be declaring non-final fields: today, a struct in Rust corresponds to a class in Dart where
all fields are final with no way to configure. For example, this code in Rust:

```rust
struct Foo {
  pub foo: String
  pub bar: Option<String>
}
```

would be translated into this Dart code:

```dart
class Foo {
  final String foo;
  final String? bar;
  Foo({
    required this.foo,
    this.bar
  });
}
```

It would be impossible to update Foo.bar if it's null once it's sent to Dart, which might be undesirable for some usages.

With an attribute e.g. `#[frb]`, we can signal that some field actually needs to be non-final in Dart:

```rust
#[frb]
struct Foo {
  pub foo: String
  #[frb(mut)]
  pub bar: Option<String>
}
```

so that this Dart code is generated:

```dart
class Foo {
  final String foo;
  String? bar;
  Foo({
    required this.foo,
    this.bar
  });
}
```

These attributes will be beneficial for other features in the future as well, so there are many advantages to implementing attributes first.

Desdaemon:
Resolves #260.

Only the attribute macro itself is implemented, not any mechanism to retrieve information from the attributes.

Some bike-shedding:
- [x] Name of the attribute, currently `frb`.
- [x] Should it be visible to users by default? (currently **no**). This boils down to convenience vs. compile times, due to the added *compile-time* dependency on `syn`.


fzyzcjy:
>  Name of the attribute, currently frb.

Agree ;)

> Should it be visible to users by default? (currently yes). This boils down to convenience vs. compile times, due to the added dependency on syn.

Why do we add dependency on syn? Since the proc macro itself does *nothing* in it, what about implementing a "raw" proc macro?

frb_example/pure_dart/dart/lib/bridge_generated.dart
```diff
@@ -84,6 +84,8 @@ abstract class FlutterRustBridgeExample extends FlutterRustBridgeBase<FlutterRus
   Future<Foobar> handleEnumStruct({required Foobar val, dynamic hint});
 
   Future<String> handleComplexEnum({required KitchenSink val, dynamic hint});
+
+  Future<void> handleCustomiedStruct({required Customized val, dynamic hint});
```
fzyzcjy, line 88: typo

fzyzcjy:
What about this: let the `frb` macro does **nothing** inside that proc macro function body. Instead, **simply output the input**. Then no need for syn, and that `fn frb() {}` will almost be empty function.

Desdaemon:
> Why do we add dependency on syn? Since the proc macro itself does nothing in it, what about implementing a "raw" proc macro?

It's needed to remove the attributes from the AST, as rustc does not accept attribute macros on fields. (Only derive macros are allowed to have field attributes.)

Desdaemon:
> and that fn frb() {} will almost be empty function.

Also yes, the macro does zero work if it's not used on a struct/enum/union.

fzyzcjy:
Hmm interesting... Can we do it without syn dependency?

Desdaemon:
> Hmm interesting... Can we do it without syn dependency?

I've never tried this, it does seem a lot more effort upfront though. What we *could* do is re-implement a subset of syn's features, but of course we'd still be depending on the parsing functionality of syn which is only a fraction of the code.

fzyzcjy:
> > Hmm interesting... Can we do it without syn dependency?
> 
> I've never tried this, it does seem a lot more effort upfront though. What we _could_ do is re-implement a subset of syn's features, but of course we'd still be depending on the parsing functionality of syn which is only a fraction of the code.

Interesting... Then we should not do that.

> Should it be visible to users by default? (currently yes). 

Then this becomes a question

Desdaemon:
Related: rust-lang/rust#65823

Desdaemon:
Given that this is a new feature with no real use cases yet, I think we can keep it gated behind a feature for now (already done), and we can reassess this need later on.

fzyzcjy:
btw is there a crate that we can use directly, instead of manually writing down those calls to remove_marker_attr? seems like this should be a commonly used situation

fzyzcjy:
going to merge it and publish it and change CI soon

Desdaemon:
FWIW the last commit shaved off about 2 seconds on a debug clean build, so I'm hopeful it's not too much of an impact.

Desdaemon:
> btw is there a crate that we can use directly, instead of manually writing down those calls to remove_marker_attr? seems like this should be a commonly used situation

I'm quite sure frb_codegen has no use for these, as they are only there to satisfy rustc? Anyways I can move them later if needed.

fzyzcjy:
> I'm quite sure frb_codegen has no use for these, as they are only there to satisfy rustc? Anyways I can move them later if needed.

I mean these https://github.com/fzyzcjy/flutter_rust_bridge/blob/409acd9482c3478c4d1fe9607eba7da1ba3f7a15/frb_macros/src/lib.rs#L14-L72

is there someone already write down a crate called, e.g. "create_proc_macro_that_does_nothing" on crates.io and we can use that

fzyzcjy:
> FWIW the last commit shaved off about 2 seconds on a debug clean build, so I'm hopeful it's not too much of an impact.

Agree. not a big problem

Desdaemon:
> is there someone already write down a crate called, e.g. "create_proc_macro_that_does_nothing" on crates.io and we can use that

Strangely enough, I couldn't find anything like that, the closest thing was wasm_bindgen itself which uses more features of syn.

fzyzcjy:
interesting. so let me merge it

I may not have time immediately and will publish tomorrow (ping me if I forget)

PR #267, opened 2021-12-31
Enum struct improvements

### LINKED ISSUE #265: Enum struct wire2api does not convert raw data
Missed calling wire2api on raw data received from Rust on enum structs. For example, this code doesn't work:

```rust
pub enum Foo { Bar, Baz }
pub enum Letters {
  Alpha,
  Beta(Foo)
}
```

Foo is represented as an integer when passed to Dart, which is used before being re-converted to a Dart enum.

### LINKED ISSUE #266: Boolean fields cannot be used
In structs and enum structs, if one of the fields is of boolean type, that field is mapped as a `u8` field in the corresponding wire struct.
This leads to the issue where ffigen generates wire functions that takes in boolean parameters but structs with integer fields (ffigen does its own bool-to-int conversions in the wire functions case), therefore wrong signatures of api2wire functions.

Desdaemon:
Several small fixes and improvements:
- Resolve #265
- Resolve #266
- Fix attributes not working on enum variants
- Add comments on enum variants and fields
- (Internal) unify tuple and normal enum structs

### DIFF
--- a/frb_codegen/src/api_types.rs
+++ b/frb_codegen/src/api_types.rs
@@ -297,7 +297,10 @@ impl ApiTypeChild for ApiTypePrimitive {
     }
 
     fn dart_wire_type(&self) -> String {
-        self.dart_api_type()
+        match self {
+            ApiTypePrimitive::Bool => "int".to_owned(),
+            _ => self.dart_api_type(),
+        }
     }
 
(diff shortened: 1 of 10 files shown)


fzyzcjy:
Wonderful!

fzyzcjy:
Will merge when CI passes

PR #277, opened 2022-01-03
Remove `syn` dependency from macros

Desdaemon:


fzyzcjy:
Excellent!

fzyzcjy:
btw this pr introduced a little bug about dependency:

https://github.com/fzyzcjy/flutter_rust_bridge/runs/4697344168?check_suite_focus=true

https://github.com/fzyzcjy/flutter_rust_bridge/commit/3e20f269e3c210dea736924b11e9084c2666502e

antonok-edm:
@fzyzcjy @Desdaemon just curious, what's the reason to avoid `syn`? As I see it:
- it's already a complete and well-tested implementation of Rust syntax parsing, which saves effort otherwise spent on building custom parsers
- it stays up to date whenever Rust starts supporting new syntactical structures
- it's well-documented and widely adopted by Rust codegen tools, so external contributors may be able to help more easily

Also curious about the rationale behind leaving codegen out of procedural macros:
- `flutter_rust_bridge` can just be a dependency in Cargo.toml without requiring a binary installation
- codegen is handled in `cargo build` with no extra steps required
- great support for error reporting, i.e. you can have custom error messages that point to problematic spans of code the same way normal compilation errors from `cargo` are reported
- no need for users to check in an intermediate `bridge_generated.rs` file into version control
- still super easy to inspect generated code using [`cargo-expand`](https://github.com/dtolnay/cargo-expand)

Having built a few procedural macros myself, this crate seems like it could benefit a lot from both, and I'd be happy to help support that effort.

fzyzcjy:
@antonok-edm Hi thanks for the suggestions!

This pr avoids syn in the proc macro, since that proc macro does *nothing* except for used as a marker. So we do not want to waste time for the users.

As for the frb_codegen, it uses syn and does the heavy job.

> flutter_rust_bridge can just be a dependency in Cargo.toml without requiring a binary installation

I guess you mean flutter_rust_bridge_codegen?

I agree, your suggestion seem also promising, and thanks for your interest in helping.

It generates not only .rs files but also .dart files which the dart/flutter code requires to compile. So maybe a manual triggering is needed?

@Desdaemon what do you think?

Desdaemon:
I think there are two primary advantages to using proc-macros: error reporting with span information, and keeping the bridge module in sync with the API file. Outside of these use-cases, this library is more focused in generating non-Rust files and therefore a codegen step is always required, not unlike wasm_bindgen (which already uses proc-macros.)

What is more interesting is whether we report errors from flutter_rust_bridge_codegen, or we let rustc emit compiler errors. Both choices have their pros and cons. What do you think @antonok-edm?

antonok-edm:
> As for the frb_codegen, it uses syn and does the heavy job.

Ah, I see it now. What confused me is that many of the functions in `frb_codegen/src/parser.rs` operate on string representations of code rather than `syn` types.

> I guess you mean flutter_rust_bridge_codegen?
> 
> It generates not only .rs files but also .dart files which the dart/flutter code requires to compile. So maybe a manual triggering is needed?

Yeah, Dart code and C headers can't be generated by a procedural macro but it should definitely be possible to do this with a build script. `flutter_rust_bridge` can expose a build dependency library crate for that, then output paths can be configured in a `build.rs` file. The build dependency library should be able to read over the source files and work with signatures from functions with the relevant procedural macro attribute.

> What is more interesting is whether we report errors from flutter_rust_bridge_codegen, or we let rustc emit compiler errors. Both choices have their pros and cons.

IMO `rustc` has really awesome error reporting infrastructure already, so the more it's possible to lean on that, the better.

fzyzcjy:
> What confused me is that many of the functions in frb_codegen/src/parser.rs operate on string representations of code rather than syn types.

@antonok-edm I agree it is a bit hacky (you know, originally I do not know this repo will be so popular!). You are welcomed to use syn to actually parse it and that would be great and much more robust!

fzyzcjy:
@antonok-edm The suggestion about not using string manipulation is now in #340. Feel free to create other issues/PRs!

PR #286, opened 2022-01-04
When writing files in frb_codegen, create directories if they don't exist

SecondFlight:
This patch adds `fs::create_dir_all` calls to `frb_codegen/src/main.rs` to ensure directories exist before trying to write files in them.

## Justification

Currently, running `frb_codegen` with `--rust-output folder_does_not_exist/code.rs` causes the program to exit with a filesystem error (tested on Windows 10).

I'm creating a bindgen script, in which I run `frb_codegen` and tell it to write my generated dart code to to a folder called `generated`. I'd like to gitignore my auto-generated code. However, my `generated` folder only contains auto-generated files, which means the folder itself isn't created when the repo is checked out for the first time. As a result, my bindgen script encounters the error above. I can create the folder using my shell script, but I think it's reasonable for `frb_codegen` to create the directories if they don't exist since it already creates nonexistent files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* When running codegen, create folders for output paths if they don't exist
+
 ## 1.12.0
 
 * Redesign documentation and make it a mdBook #272
--- a/frb_codegen/src/main.rs
+++ b/frb_codegen/src/main.rs
@@ -1,4 +1,5 @@
 use std::fs;
+use std::path::Path;
 
 use env_logger::Env;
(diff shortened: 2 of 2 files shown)


fzyzcjy:
Sure. I will run CI and merge it.

btw you can use a `.gitkeep` for your own use case as well

fzyzcjy:
CI fails. could you plz check it

fzyzcjy:
CI error caused by other things, so I will merge it and fix.

fzyzcjy:
@all-contributors please add @SecondFlight for code

Issue #241, opened 2021-12-09
[Vote to make it open-source] Object pools


fzyzcjy:
Avoid transfering large objects or even non-serializable objects (e.g. with a mutex) between Rust and Dart. Instead, let it always live in Rust, and Dart only gets an integer as a handle.

Remark: I have written and used it internally in my project (in production environment). If you need to use it, I can make it open-source.

fzyzcjy:
If #243 is done, this is no longer needed

fzyzcjy:
bump since it is #242's dependency

PR #293, opened 2022-01-04
Opaque types

### LINKED ISSUE #68: How to return objects of an opaque class? Or even let Dart to own and manage Rust's objects?
This is the code I used before, as follows
use flutter_rust_bridge Is there an equivalent implementation ?

```
#[derive_ReprC]
#[ReprC::opaque]
pub struct Blake3Hasher {
  pub h: Hasher,
}

#[ffi_export]
impl Blake3Hasher {
  pub fn new() -> Self {
    Self { h: Hasher::new() }
  }
}

#[ffi_export]
pub fn blake3_hasher_new() -> repr_c::Box<Blake3Hasher> {
  let peer = repr_c::Box::new(Blake3Hasher::new());
  peer
}

#[ffi_export]
pub fn blake3_hasher_update(hasher: &mut Blake3Hasher, data: *const u8, len: usize) {
  let v = unsafe { slice::from_raw_parts(data, len) };
  hasher.h.update(v);
}

#[ffi_export]
pub fn blake3_hasher_end(hasher: repr_c::Box<Blake3Hasher>) -> *const u8 {
  return const_u8(*hasher.h.finalize().as_bytes());
}
```

### LINKED ISSUE #243: [Draft] `flutter_rust_bridge` 2.0: Allow Dart to *own* and manage Rust's opaque objects
Related discussions: https://github.com/fzyzcjy/flutter_rust_bridge/issues/68 (thanks @gcxfd)

## Why

Powerful!

### What data structure to pass around?

Attempt 1: Maybe `Box<Arc<RwLock<T>>>` as a raw pointer?

However, can we do that? When raw pointer is passed from dart to multiple rust functions concurrently, multiple rust functions will all rehydrate the *same* `Arc` object from the raw pointer, so have problems?

Attempt 2: Maybe `Box<(AtomicUsize, RwLock<T>)>` as a raw pointer? the AtomicUsize is like a self-managed Arc.

Attempt 3: Maybe `Box<RwLock<T>>` is enough? Since when it is used in a Rust function, the Rust function always *immutably borrow* the `RwLock<T>`, and it is Dart who indeed owns it. --> See "How to ensure that the Dart opaque object is still alive before" below. **This method is WRONG**.

### How can we create a Dart object with a finalizer, and then pass that object from Rust to Dart?

use `as_native_pointer` of Dart_CObject https://github.com/dart-lang/sdk/blob/5f32f4d7b678dc08d6079664637ad061902c5b69/runtime/include/dart_native_api.h#L88

Interestingly, it was added only 4months ago! https://github.com/dart-lang/sdk/commit/bc387837c4f1162d23a1434a8715ec4175fed9be#diff-587868a9cf3a0dd424f04d4d4d30efef5f00661441e094b1e993d20d3847199d

In other words, in Dart 2.14 (Flutter 2.5), https://github.com/dart-lang/sdk/blob/2.14.0/runtime/include/dart_native_api.h does NOT have it.
In Dart 2.15 (Flutter 2.8 - released a few days ago!) https://github.com/dart-lang/sdk/blob/2.15.0/runtime/include/dart_native_api.h It is there now.

### How to update the real "external size" of the Rust object to Dart, when using the `Dart_CObject` instead of `DartHandle`?

see: https://github.com/dart-lang/sdk/issues/47901

[TODO]

### How to know the real "external size" of the Rust object?

It is needed by Dart VM, otherwise vm may not know the memory pressure (I guess?).

For example, if the user's data structure is `struct MyObject {a: i32, b: String, c: Box<Something>}`, how can we know the actual heap size used? Notice `String` and `Box`...

Attempt 1: Maybe create a trait and enforce user to implement that trait. Say, `pub trait Sizeable { fn memory_size(&self) -> usize }`.

However, how to know that size has changed in Rust, and tell Dart the new size whenever it has changed (maybe using Dart_UpdateExternalSize/Dart_UpdateFinalizableExternalSize)?

Attempt 2: Not only have the trait `Sizeable`, but also require the user to *manually* call flutter_rust_bridge when its memory size changes.

Not that friendly...

Attempt 3: Can we automatically generate some code?

Attempt 4: Is there a library to do this?

https://crates.io/crates/deepsize

Related issue: https://github.com/Aeledfyr/deepsize/issues/28

Problem: Can it accurately know things like ndarray? https://github.com/Aeledfyr/deepsize/issues/29

Attempt 5: Can we just use `Attempt 2` and tolerate that the users may not update the size frequently enough?

see https://github.com/dart-lang/sdk/issues/47900

### Will we face performance problems?

Look at how ObjC makes ref counting, sounds a bit similar.

May not be a problem, since we have to have interior mutability, and have to be thread safe.

### How to ensure that the Dart opaque object is still alive when Rust is using that object? Especially, when in async

Serious problem!

So we really need `Arc` to have reference counting, instead of blindly let Dart to own the object. After doing so, we simply do not care when is Dart object deallocated.

### Do we have concurrecy problems?

Sounds no, since we have `RwLock`.

### What if, when Dart calls Rust, at the same time, Dart makes a GC and make that pointer invalid?

Code example:

```
// Dart
var opaque_pointer = ...;
my_ffi_function(opaque_pointer);

// Rust
fn my_ffi_function(opaque_pointer) {
  opaque_pointer.convert_raw_pointer_into_rust_object().increase_arc_count(); // (a)
}
```

When we are executing (a), what if Dart make a GC and the finalizer of `opaque_pointer` is called, then we are trying to do `increase_arc_count` on an invalid object? (Notice if the finalizer is called and the pointer has rc=1, then it is dropped immediately)

Attempt 1: Maybe force the opaque_pointer to be used *after* the ffi call is done. For example:

```
var opaque_pointer = ...;
await my_ffi_function(opaque_pointer);
pretend_we_use_it_to_avoid_gc(opaque_pointer); // <--- NOTE
```

[TODO] Seems to have seen some discussions about this before?

Desdaemon:
- Resolves #243 
- Resolves #68

There are a few kinks to work out before we can stabilize it:
- [x] Dart API
- [x] Safety
- [x] Data structure
- [ ] Safety documentation

Other tasks:
- [x] ~~Copy use statements from API file~~ Require users to declare `pub use`
- [x] ~~Find a way to safely convert between pointers of unrelated types~~ Cast to `*const c_void`

More other tasks (added by @fzyzcjy):

- [x] more simple tests (since this can cause very hard-to-debug bugs)
- [x] more concurrent tests - many bugs may only happen under extreme conditions
- [x] fix the dangerous bug: https://github.com/fzyzcjy/flutter_rust_bridge/pull/293#issuecomment-1006191364
- [x] fix another possibly bug https://github.com/fzyzcjy/flutter_rust_bridge/pull/293#issuecomment-1006204527
- [ ] enhance valgrind checker: currently all invalid memory access are simply ignored, because even a literally hello-world simple dart program will cause tons of invalid memory access. But since this feature is tricky we may need stronger checks.
- [ ] add changelog
- [ ] publish new version

## Dart API
An opaque pointer to e.g. `[u8; 2000]` would look like this:

```dart
class U82000 extends Opaque {
  U8200._(int ptr, int drop) : super(Pointer.fromAddress(ptr), Pointer.fromAddress(drop));

  void dispose(); // implemented by super, safe to call many times
  bool isStale(); // true if dispose has been called
}
```

I don't know if there's a lint that enforces calling `dispose`, but it would be really helpful for this situation.
Otherwise, I think this is an acceptable API to use. ~~Then in the future when weak references are implemented, calls to `dispose` will become no-op and be deprecated.~~ It is most likely that `dispose` will continue to be necessary in the foreseeable future.

## Safety

*Resolved: Opaque may not obtain a mutable reference. Here's the original text:*
<details>
The Rust side however has the issue of whether we should allow Rust functions to obtain a mutable reference to T:

```rust
impl<T> Opaque<T> {
  ..
  pub unsafe fn as_mut(&mut self) -> Option<&mut T>;
}
```

then in a function:

```rust
pub fn handle_opaque(mut arr: Opaque<[u8; 2000]>) -> Result<()> {
  unsafe {
    if let Some(arr) = arr.as_mut() {
      std::mem::drop_in_place(arr); // instant UB
    }
  }
  Ok(())
}
```

In line with other Rust API, since we do not have ownership of T we can't really hand out mutable references, but not having this functionality also precludes a lot of potential usage out of opaque pointers. I'm not quite sure how to proceed here.
</details>


fzyzcjy:
>  don't know if there's a lint that enforces calling dispose, but it would be really helpful for this situation.

I guess no, or at least the lint will never be strong enough. We know in Flutter it is common to have a dispose, e.g. for a Stream, a ChangeNotifier, etc. There do exist lints to warn you when you *never* dispose them, but nobody can ensure we call them *in the right place*. 

> Then in the future when weak references are implemented, calls to dispose will become no-op and be deprecated.

I guess it will still be used and never deprecated? b/c from the comments of dart team https://github.com/dart-lang/language/issues/1847#issuecomment-1002860171, fully automatic memory management is strongly discouraged.

Even worse, by doing this way we can never tell Dart VM that this "int ptr" may occupy huge memory indeed. Then Dart VM will even do not bother to cause a GC and dispose objects since it wrongly thinks nobody is using huge memory.

fzyzcjy:
> extends Opaque

btw I guess dart:ffi has a class called Opaque so we may rename it, e.g. FrbOpaque

fzyzcjy:
## Question: When the dart object is `dispose()`ed, how can we ensure nobody will ever touch invalid memory?

fzyzcjy:
### The bridge may look like (very roughly):

```
fn convert_dart_wire_object_to_rust_object<T>(wire_ptr: *const isize) {
  // NOTE manuallydrop
  let dart_owned_obj: Box<RwLock<T>> = ManuallyDrop::new(Box::from_raw_ptr(wire_ptr));
  dart_owned_obj
}
```

### Problem: Are we safe? Notice the `convert_dart_wire_object_to_rust_object` can be called concurrently by multiple calls! So we may be holding the raw pointer as a `Box<RwLock<T>>` concurrently.

What do you think?

Desdaemon:
> However, the memory may still be used - for example, a long running Rust method may still be using this. This may be acceptable since it is the direct result of reference counting.

The only way I think this can be safe is that our hypothetical FrbOpaque is a newtype over Arc<T> (then users can specify that they want an `Opaque<RwLock<T>>` instead). Maybe something like a model of *shared ownership* would help, if not reference counting?

> `let dart_owned_obj: Box<RwLock<T>> = ManuallyDrop::new(Box::from_raw_ptr(wire_ptr));`

I think there may be some issue with that, I've been seeing it as early as trying to send a `*mut [u8; _]` only for rustc to tell me it is not `Send`. So yeah, I think Arc<T> is the best compromise for now.

fzyzcjy:
## My attempt on data structures

reference: https://github.com/fzyzcjy/flutter_rust_bridge/issues/243

Dart side: `int ptr`.

Rust side: `Box<Arc<RwLock<T>>>`.

Rust usage:

```
fn convert_dart_wire_object_to_rust_object<T>(wire_ptr: *const isize) {
  // NOTE manuallydrop
  let dart_owned_obj: Box<Arc<RwLock<T>>> = ManuallyDrop::new(Box::from_raw_ptr(wire_ptr));
  // NOTE we clone the Arc and return that clone
  dart_owned_obj.clone()
}
```

### Question: Is it safe?

When raw pointer is passed from dart to multiple rust functions concurrently, multiple rust functions will all rehydrate the same Arc object from the raw pointer, so have problems?

fzyzcjy:
> The only way I think this can be safe is that our hypothetical FrbOpaque is a newtype over Arc<T> (then users can specify that they want an Opaque<RwLock<T>> instead). Maybe something like a model of shared ownership would help, if not reference counting?

Same idea here ;) (See my #243 root post as well)

> let dart_owned_obj: Box<RwLock<T>> = ManuallyDrop::new(Box::from_raw_ptr(wire_ptr));

plz ignore it, I realize it has problems

Desdaemon:
> When raw pointer is passed from dart to multiple rust functions concurrently, multiple rust functions will all rehydrate the same Arc object from the raw pointer, so have problems?

Haven't tested yet, but I think it might go something like this?

```rust
// begin, refcount = 1 as owned by Dart
let obj1: Arc = convert_dart_wire_object_to_rust_object(wire_ptr); // clone, refcount = 2
let obj2: Arc = convert_dart_wire_object_to_rust_object(wire_ptr); // clone, refcount = 3
normal_rust_function_that_accepts_an_object(obj1); // does refcount decrease here?
normal_rust_function_that_accepts_an_object(obj2); // or here?
drop(obj1); // it must not, because refcount must be 3 here
drop(obj2); // similarly, refcount = 2
// refcount = 1, safe
```

The wildcard is the two normal fns in the middle, if they accept a `&mut Arc` then they can call `Arc::decrement_strong_count` and invalidate this assumption (which is unsafe anyways). As I was writing this, I realized you can turn an Arc into a Weak reference, which would be quite hard but not impossible to force-decrement the refcount, so would this make FrbObject a Weak instead?

fzyzcjy:
> Arc::decrement_strong_count

maybe we can assume users never do that? that is a *unsafe* function.

> The wildcard is the two normal fns in the middle, if they accept a &mut Arc then they

oh, moreover, I realize they should accept `Arc` not `&mut Arc` since indeed those functions can run for a long time.

So it is like:

```
let obj1: Arc = convert_dart_wire_object_to_rust_object(wire_ptr); 
let obj2: Arc = convert_dart_wire_object_to_rust_object(wire_ptr); 
normal_rust_function_that_accepts_an_object(obj1); 
normal_rust_function_that_accepts_an_object(obj2); 

fn normal_rust_function_that_accepts_an_object(o: Arc) {...}
```

fzyzcjy:
### Second attempt on data structures

When reading https://doc.rust-lang.org/std/sync/struct.Arc.html I realize they have:

* into_raw
* as_ptr
* from_raw

can we exploit them, then we do not need `Box<Arc<...>>` but only need `Arc<...>`

Desdaemon:
> accept Arc not &mut Arc

Oh, that hypothetical is fine because if it takes ownership then you cannot drop them anyway. But the same worry presents if that inner function somehow decreases the refcount more than once.

> we can assume users never do that

Maybe not, but there is still an off-chance one of that function's dependencies will, still that is probably a 99.9% hypothetical that might turn out impossible to debug once present in production code.

fzyzcjy:
post above edited, with a full example

fzyzcjy:
related https://stackoverflow.com/questions/70587003/safely-call-arcfrom-raw-multiple-times-for-a-raw-pointer-in-other-words-ho

fzyzcjy:
> Maybe not, but there is still an off-chance one of that function's dependencies will, still that is probably a 99.9% hypothetical that might turn out impossible to debug once present in production code.

Well I agree with you: If the refcount decreases to zero accidentally, the raw pointer immediately become a rubbish, and **we have NO WAY to detect it!**

fzyzcjy:
Then, what about the object pool https://github.com/fzyzcjy/flutter_rust_bridge/issues/241 approach? If anyway we are using `dispose` manually, the object pool does not cause any extra complexity. Moreover, it is much, much safer, since we almost never touch a `unsafe` operation with object pool. We just use indices (like pointers) to index into the pool (a HashMap, for example). Then, invalid indices (read: invalid pointers in this PR) just causes the HashMap to panic, instead of terrible UB as in this PR.

I have already had a object pool implemented in my production environment so I can open source it (and then we make some improvements to it).

fzyzcjy:
If we can do automatic memory management (i.e. no manual `dispose`) then this pr is wonderful; but since the Dart team says "strongly discourage" such, we should manually call `dispose` forever,

Desdaemon:
> we should manually call dispose forever

I think that might be the case, seeing how dispose is already part of Flutter culture.

fzyzcjy:
https://stackoverflow.com/questions/70587003/safely-call-arcfrom-raw-multiple-times-for-a-raw-pointer-in-other-words-how/70587029?noredirect=1#comment124780745_70587029 gives an improvement:

```
let x_ptr = Arc::into_raw(Arc::new(some_big_object));

for i in 0..10 {
    thread::spawn(|| {
        unsafe {
            Arc::increment_strong_count(x_ptr);
            let x_recovered = Arc::from_raw(x_ptr);
            make_use_of_it(x_recovered);
        }
    });
}

unsafe {
    Arc::decrement_strong_count(x_ptr);
    // Could be also
    // drop(Arc::from_raw(x_ptr));
}

File: frb_dart/lib/src/basic.dart [hunk 1]
```diff
@@ -209,3 +209,21 @@ class WireSyncReturnStruct extends ffi.Struct {
   @ffi.Uint8()
   external int success;
 }
+
+/// An opaque pointer to a native C or Rust type.
+/// Recipients of this type should call [dispose] at some point during runtime.
+class Opaque {
```
raphaelrobert, line 215: This name might clash with https://api.flutter.dev/flutter/dart-ffi/Opaque-class.html

raphaelrobert (commented):
Great you're tackling this!

See also my PoC that uses `Arc<RwLock<T>>`, maybe you can reuse some of that: https://gist.github.com/raphaelrobert/f7e517a7d4d02cd4b6d24cb270207791

Desdaemon:
Changes up until last commit:
- Opaque is now FrbOpaque
- opaque::Opaque is now a transparent newtype over `Option<Arc<T>>`
- `opaque::Opaque::<T>::as_ref(&self) -> Option<&T>` for immutable references
- Extensions for `Opaque<RwLock<T>>`

fzyzcjy:
Excellent! A few things to check before merging

- [ ] more simple tests (since this can cause very hard-to-debug bugs)
- [ ] some concurrent tests
- [ ] discuss the question below

fzyzcjy:
### Bug: What if a user calls ffi and call dispose before the rust code executes and increases the reference counting? Currently it will be undefined behavior and cause all troubles!

In other words:

```
// dart
var object = ... an opaque pointer ...;
unawaited(my_rust_function(object.ptr)); // (1)
object.dispose(); // (2)

// rust
fn my_rust_function(object_ptr: isize) {
  let object: Arc<...> = convert_to_object(object_ptr);
  ... use it ...
}
fn convert_to_object(object_ptr: isize) -> Arc<...> {
  Arc::increment_strong_count(object_ptr); // (3)
  return Arc::from_raw(object_ptr); // (4)
}
```

It can be the case that, the code is executed in the sequence of (1) - (2) - (3) - (4). 

Then awful things happen: At (1) things are still good. After executing (2), the `object.dispose` will decrease refcount of the Arc. So it has refcount of **zero**, and the data of Arc is dropped! Then (3) and (4) happens - we are trying to access an invalid (deallocated) pointer. That is TERRIBLE and we are touching invalid memory and can cause UB.

fzyzcjy:
posts above are edited.

Desdaemon:
> What if a user calls ffi and call dispose before the rust code executes and increases the reference counting?

Unfortunately this is a real bug right now, I'll get to fixing this.

Desdaemon:
The bug turned out to be similar, but not exactly the same as you described. Given the following situation:
```dart
var opaque = await createOpaque();
var future = handleOpaque(val: opaque); // Future<Opaque>
opaque.dispose() // refcount = 1
opaque = await future; // refcount = 0!!
expect(opaque.isStale(), false); // passes
opaque.dispose(); // segfault
```
It was essentially a double free, but otherwise it seemed to be able to increase the refcount before Dart can dispose it.
There is no 100% guarantee that this is the case, but I don't not how to write tests that can challenge that assumption.

fzyzcjy:
Well can you remove 

```
opaque = await future;
```

since by doing this, you are reassigning the variable and it is harder for people to understand what is happening

maybe 

```
var opaque_a = await createOpaque();
var future = handleOpaque(val: opaque_a); // Future<Opaque>
opaque_a.dispose() // refcount = 1
var opaque_b = await future; // refcount = 0!!
expect(opaque_b.isStale(), false); // passes
opaque_b.dispose(); // segfault
```

Desdaemon:
Fair point, I've edited the example to clarify that. This is also part of the test suite btw.

fzyzcjy:
> There is no 100% guarantee that this is the case,

May depend on your rust code and the auto generated code as well, since you are indeed disposing the *returned object* instead of considering the *input object*.

And, my proposed bug is not that easy to reproduce - it does not happen frequently, since with high probability it is (1)-(3)-(4)-(2) instead of (1)-(2)-(3)-(4). But, since we are developing a library, and concurrency bugs are VERY hard to debug for the users, we should tackle it.

Desdaemon:
Maybe we should separate the example library from actual testing, it is getting kind of cluttered.

fzyzcjy:
@Desdaemon Agree!

Desdaemon:
Not a solution, but since this is a data race we're trying to solve, the provably correct solution is some kind of synchronization between Dart and Rust, though this does add a fair amount of overhead. I'm doing some testing in GDB and already saw some fascinating results, but it will take me a while to have something substantial.

Desdaemon:
> How can a Weak still be happily held even when Arc is already disposed an no memory exist?

From std documentation:
> Since a Weak reference does not count towards ownership, it will not prevent the value stored in the allocation from being dropped, and Weak itself makes no guarantees about the value still being present. Thus it may return None when upgraded. Note however that a Weak reference does prevent the allocation itself (the backing store) from being deallocated.

This is one possible solution, but it's not clear how we can use it yet.

fzyzcjy:
> This is one possible solution, but it's not clear how we can use it yet. 

My question is: In my naive guess, when `Arc` has ref counting down to zero, it frees the data, and there is *nothing* left in memory (is it correct?). However, some rust code may still be holding a `Weak` to that location in memory. Then, the `Weak` is literally holding a pointer to a invalid (i.e. freed) memory. Then how can Weak accurately detect "this is invalid memory"?
I guess my thoughts above must have something wrong. Have not dig very deep into src code yet.

Desdaemon:
What I ended up with is using a `parking_lot::Mutex<()>` to wrap the critical sections. I think this is the easiest solution for now, unless anyone has an alternative.

fzyzcjy:
> What I ended up with is using a parking_lot::Mutex<()> to wrap the critical sections. I think this is the easiest solution for now, unless anyone has an alternative.

I guess you mean a *global* lock that is shared among everything?

Sounds like a feasible solution. The performance may be reduced, but hope few people is having this as a bottleneck.

fzyzcjy:
**Wait a bit...** I get confused. How is this bug https://github.com/fzyzcjy/flutter_rust_bridge/pull/293#issuecomment-1006191364 solved with locking? Cannot see how locking prevents this so could you plz give a bit more hints?

Desdaemon:
On second thought, that doesn't really solve it. I'll get back to this once I have something better.

fzyzcjy:
@Desdaemon Take you time and thanks for your contribution!

Desdaemon:
Can't fully comment on this, but would everyone agree on this solution? <https://docs.rs/semaphore/0.4.0/semaphore/>

fzyzcjy:
Some remarks on `Weak` and `Arc`:

> Note however that a Weak reference does prevent the allocation itself (the backing store) from being deallocated. (by std::rc::Weak, but arc should be same)

Thus this does not inspire anything to us. Originally I wrongly thought Weak has some magic to detect dangling pointer.

fzyzcjy:
If this PR is too challenging to implement, we may consider the "pool" approach, which is even equivalent to this PR but seems to be safer (less strange corner cases).

My current implementation of pool is: https://gist.github.com/fzyzcjy/ff84c4a4cc37950275505a5fdd99039c (directly copied from my private codebase). However, for more general and easy-to-use pools, we may consider the following (slight modification):

There is one giant pool that holds all objects. You can get objects, put objects, etc, into it, and it will use a integer (the "handle") to represent the object. The handle is analogy to the pointer.

The "subpool" in my current implementation can also be taken into account. I use subpools to make a "scoped" thing. In other words, I can call `subpool.dispose()` and everything in it will be disposed - then if you forget to dispose something, it can safe your life.

Desdaemon:
The solution was quite simple in retrospect: if dispose is called non-deterministically, then the cloning of Arc must occur as part invoking the wire functions. This ensures that Rust already holds the allocation and a call to dispose won't drop the memory. Also new is the ability to make opaque pointers to trait objects.

I'm thinking about tapping into the Dart C API for attaching a [finalizer](https://github.com/dart-lang/sdk/issues/35770) instead of forcing users to manually call dispose, but it seems like there are problems with that approach too. Any thoughts?

fzyzcjy:
> Also new is the ability to make opaque pointers to trait objects (!)

Interesting! Do not forget to mention it in changelog/doc later ;)

> I'm thinking about tapping into the Dart C API for attaching a finalizer instead of forcing users to manually call dispose, but it seems like there are problems with that approach too. Any thoughts?

Well recall that "strongly discourage" comment from Dart official team. IMHO applies to it too. What's worse, with dart GC finalizer, we even do not tell Dart VM how huge the external object is. So Dart GC can happily think that there is no memory pressure, while Rust code allocates, say, 10GB memory and really need a GC.

About the solving of the problem: Let me think twice.

fzyzcjy:
> The solution was quite simple in retrospect: if dispose is called non-deterministically, then the cloning of Arc must occur as part invoking the wire functions. This ensures that Rust already holds the allocation and a call to dispose won't drop the memory.

Well may still need some care: every single `await` in Dart code is indeed some asynchronous gap. In other words, e.g.

```
func1();
func2();
await func3();
func4();
```

in the example above, we are sure no other code can execute between func1() and func2(). But, a ton of code can execute between func3() and func4().

Thus, we must be **very careful** about not introducing **any async gap**. Even one async gap will break the safety guarantee.

What's worse, frb allows users to customize the executors and handlers. For example, you can customize and wrap dart side's `executeAsync`, as well as Rust side. So we must be very sure this cannot be violated even with customizations.

raphaelrobert:
This is amazing progress 🚀!

Regarding the finalizers: I think the takeaway right now is clearly that Dart cannot and should not manage the native memory. Dart's GC only kicks in when new objects are allocated on the Dart heap, not when the the memory pressure grows from allocating objects on the native heap. Once https://github.com/dart-lang/language/issues/1847 has landed, it should be possible to automatically call `.dispose()`, but it won't be a full replacement for calling it manually.

Maybe going forward there is a way to implement some additional GC on the Dart side that tracks the allocation on the native heap with weak references, but that's something that I haven't fully investigated yet.

sagudev:
👀

sagudev:
With the release of flutter 3 and dart 2.17 the [`NativeFinalizer`](https://api.dart.dev/stable/2.17.0/dart-ffi/NativeFinalizer-class.html) is available as per https://medium.com/dartlang/dart-2-17-b216bfc80c5d

lattice0:
🙏🏻

lattice0:
Just tested it, and currently, `DartSafe` is not being included in `bridge_generated.rs`. Is it something to be done or will the users have to import `frb_rust` to use `Opaque` types?

book/src/feature/opaque.md
```diff
@@ -0,0 +1,32 @@
...
+`DartSafe` is automatically implemented for types that implement all of these
+traits.
+
+This requirement has one consequence: trait objects e.g. `Box<dyn Trait>` must
+also implement `DartSafe` to be compatible with Opaque. Here is a common way to
+define opaque pointers to trait objects:
+
+```rust,ignore
+// first define a trait that exhibits the desired traits
+pub trait MyTrait: DartSafe + Clone + .. {}
```
lattice0, line 27: for me, `Clone` prevents `MyTrait` from building a vtable

lattice0 (approved):

lattice0:
I had to add 
``` 
import 'package:meta/meta.dart';
```
to the generated code otherwise the `@sealed` wouldn't work.

fzyzcjy:
Related: 

* #775

fzyzcjy:
Btw, given that we already have opaque types implemented, shall we close this PR?

PR #328, opened 2022-02-11
Don't panic on import renames

### LINKED ISSUE #326: [Regression] Rust code can no longer contain use renames, even though we don't try to parse them
Any Rust code that contains an import rename will cause `source_graph.rs` to panic. This is bad, since imports are not currently used by codegen and so shouldn't matter.

I'm working on a fix for this, but just leaving this issue here in case someone gets to releasing before I open a PR.

SecondFlight:
Fixes #326.

### DIFF
--- a/frb_codegen/src/source_graph.rs
+++ b/frb_codegen/src/source_graph.rs
@@ -1,8 +1,8 @@
 /*
     Things this doesn't currently support that it might need to later:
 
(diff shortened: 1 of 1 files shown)


fzyzcjy:
I will merge it after CI passes (excepting for those failures unrelated to this pr)

fzyzcjy:
Looks great, :tada:

PR #335, opened 2022-02-13
Support bridging functions with return types other than `Result<_>`

### LINKED ISSUE #332: Return types other than `Result<_>`
I tried to get started with this library and instantly ran into some problematic function signatures. I had to dig into the code to find out that... [only `Result<_>` is supported as a return type](https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_codegen/src/parser.rs#L132-L136).

Some functions should never fail; is there a reason to enforce that restriction everywhere? If it's an intentional design decision, it'd be nice to have it listed in the documentation somewhere with an explanation (I'm aware of https://github.com/fzyzcjy/flutter_rust_bridge/issues/318). But, either way, the error messages could provide much more useful guidance to users starting out. Currently a signature like the following:

```rust
pub fn concat_strings(a: String, b: String) -> String {
```

gives the following error:

```
thread 'main' panicked at 'unsupported type_string: String', /home/antonok/.config/cargo/registry/src/github.com-1ecc6299db9ec823/flutter_rust_bridge_codegen-1.16.0/src/parser.rs:135:21
```

...whereas something like `` Functions must return `Result<_>` `` would be immensely more helpful for identifying the issue and suggesting a solution.

antonok-edm:
Fixes #332

I implemented this by wrapping the returned type in `bridge_generated.rs` with `Ok(...)` whenever there's a non-`Result<_>` return type. So under the hood, the top-level return type is still always a special-cased `Result<_>`.

The _ideal_ solution would probably be to add support for parsing `Result<_>` as a type and giving it a generalized Dart<->Rust representation, so it could be supported inside nested structures. In the meantime this is a nice convenience for bridging infallible functions.

As an aside: I'm not really sure what `StreamSink` is used for, so I didn't touch anything related to it. `Result<_>` still seems to be required there as a return type, although there's also no error when it's not specified.

### DIFF
--- a/frb_example/pure_dart/rust/src/api.rs
+++ b/frb_example/pure_dart/rust/src/api.rs
@@ -46,7 +46,6 @@ pub fn handle_string(s: String) -> String {
 
 pub fn handle_return_unit() -> () {
     println!("handle_return_unit()");
-    ()
 }
 
 // to check that `Vec<u8>` can be used as return type
@@ -381,9 +380,8 @@ pub struct Customized {
     pub non_final_field: Option<String>,
 }
 
-pub fn handle_customized_struct(val: Customized) -> () {
+pub fn handle_customized_struct(val: Customized) {
     println!("{:#?}", val);
-    ()
 }
 
 #[frb]
--- a/frb_example/pure_dart/rust/src/api.rs
+++ b/frb_example/pure_dart/rust/src/api.rs
@@ -12,8 +12,8 @@ use flutter_rust_bridge::*;
 use crate::data::{MyEnum, MyStruct};
 
 /// Documentation on a simple adder function.
-pub fn simple_adder(a: i32, b: i32) -> Result<i32> {
-    Ok(a + b)
+pub fn simple_adder(a: i32, b: i32) -> i32 {
+    a + b
 }
 
 /**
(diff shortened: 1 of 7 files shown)


frb_example/pure_dart/rust/src/api.rs
```diff
@@ -22,37 +22,37 @@ pub fn simple_adder(a: i32, b: i32) -> Result<i32> {
...
     let s2 = s.clone();
-    Ok(s + &s2)
+    s + &s2
 }
 
-pub fn handle_return_unit() -> Result<()> {
+pub fn handle_return_unit() -> () {
     println!("handle_return_unit()");
-    Ok(())
+    ()
```
fzyzcjy, line 49: Seems that Rust grammar allows us to write *nothing* for this case

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -22,37 +22,37 @@ pub fn simple_adder(a: i32, b: i32) -> Result<i32> {
...
-pub fn handle_string(s: String) -> Result<String> {
+pub fn handle_string(s: String) -> String {
     println!("handle_string({})", &s);
     let s2 = s.clone();
-    Ok(s + &s2)
+    s + &s2
 }
 
-pub fn handle_return_unit() -> Result<()> {
+pub fn handle_return_unit() -> () {
```
fzyzcjy, line 47: Btw if you are interested, this case can also be improved: Use may write `pub fn something() {}` without `-> ()`

fzyzcjy:
Thanks for the PR! Only some tiny problems above and LGTM

frb_example/pure_dart/rust/src/api.rs
antonok-edm: Good call. Actually, this didn't work without it! But I've just pushed an extra fix to handle it.

I left it this way to specifically test the `-> ()` case, but I did remove `-> ()` from `handle_customized_struct` which also had the same situation.

fzyzcjy:
@all-contributors bot please add @antonok-edm for code

fzyzcjy:
Hmm CI fails...

PR #346, opened 2022-02-17
Avoid converting syn types to strings before parsing

### LINKED ISSUE #340: Code beautification: When parsing types, fully use `syn` instead of partially string manipulation
Thanks @antonok-edm suggesting this in https://github.com/fzyzcjy/flutter_rust_bridge/pull/277#issuecomment-1038595982

p.s. I personally suggest wait for a few hours until #338 is landed, since that refactoration can move code around.

antonok-edm:
Closes #340 

This is already a good number of changes to start with, but there's a few followups that might be useful here, including:
- Better error handling. Right now everything either panics or returns an `Option` as an output type. `Result`s would be more friendly to work with.
- Unifying the `IrFunc` model with `IrFuncOutput` and `IrFuncArg`. The way I've implemented it, just the type from each `IrFuncArg` and `IrFuncOutput` is extracted and stored in `IrFunc` along with the `fallible` and `mode` information. This would also be a good opportunity to forbid more combinations in function signatures, like `StreamSink` along with another return type.

I tried to keep functionality as similar as possible; all the existing test cases are interpreted without any changes. There is one significant difference in that types with leading paths are _always allowed_ now, whereas it used to only work for "special" types (e.g. `anyhow::Result<_>`) before. Everything before the last path segment is ignored. I left it this way for simplicity/consistency, but it's easy enough to change now if necessary.

### DIFF
--- a/frb_codegen/src/ir/func.rs
+++ b/frb_codegen/src/ir/func.rs
@@ -16,6 +16,20 @@ impl IrFunc {
     }
 }
 
+/// Represents a function's output type
+#[derive(Debug, Clone)]
+pub enum IrFuncOutput {
+    ResultType(IrType),
+    Type(IrType),
+}
+
+/// Represents the type of an argument to a function
+#[derive(Debug, Clone)]
+pub enum IrFuncArg {
+    StreamSinkType(IrType),
+    Type(IrType),
+}
+
 #[derive(Debug, Clone, PartialOrd, PartialEq)]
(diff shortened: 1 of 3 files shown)


fzyzcjy:
Excellent job and LGTM!

> This is already a good number of changes to start with, but there's a few followups that might be useful here, including:

Agree, especially the error handling part. Looking forward to the PR!

PR #352, opened 2022-02-19
Add struct mirrors to use external types

Unoqwy:
This PR adds an attribute marker (called mirror) to use external structs/enums. Since frb needs to know the fields in a struct, external types need to be redefined in the crate, with the marker. Once frb has read the placeholder types to generate the glue code, they are not used (they are really just used to know the "model" of an external type), so they should not even make it to runtime. Despite this minor inconvenience, it still makes the experience of dealing with external types way better.
Usage is demonstrated in the `pure_dart` example.

Mirrored structs/enums need to be wrapped in a newtype (invisible to users) to work around Rust's orphan rule. That's what "wrapper" refers to in the codegen.

Enum struct variants are not supported, I may work on adding them in a future PR.


fzyzcjy:
Good job! Just a question: What happens if the user accidentally write down the *wrong* definition, i.e. a definition different from the actual mirrored type?

For example, for the external type:

```
#[derive(Debug, Clone)]
pub struct ApplicationSettings {
    pub name: String,
    pub version: String,
    pub mode: ApplicationMode,
}
```

What if we have the following problems, e.g.:

```
#[frb(mirror(ApplicationSettings))]
pub struct _ApplicationSettings {
    pub name: String,
    pub version: i32, // 1. wrong type? (can be even more sophisticated)
    // pub mode: ApplicationMode, // 2. missing field
    pub a: i32, // 3. extra field
}
```

This can happen frequently, when, for example, the external library *upgrades* and add/remove/change a field.

fzyzcjy:
Can we guarantee that this is a compile-time error? If so then we are safe.

fzyzcjy:
Can we guarantee that this is a compile-time error? If so then we are safe.

frb_codegen/src/markers.rs
```diff
@@ -0,0 +1,24 @@
...
+        .iter()
+        .filter(|attr| attr.path.is_ident("frb"))
+        .find_map(|attr| match attr.parse_meta() {
+            Ok(Meta::List(MetaList { nested, .. })) => nested.iter().find_map(|meta| match meta {
+                NestedMeta::Meta(Meta::List(MetaList {
+                    path,
+                    nested: mirror,
+                    ..
+                })) if path.is_ident("mirror") && mirror.len() == 1 => {
+                    match mirror.first().unwrap() {
```
Desdaemon, line 15: Perhaps match with a Some(_) or add an `expect` description?

fzyzcjy:
@all-contributors bot please add @Unoqwy for code

Desdaemon (commented):
LGTM

Unoqwy:
> What happens if the user accidentally write down the wrong definition, i.e. a definition different from the actual mirrored type?
Hm I think this should be considered undefined behavior because it may act differently depending on the types.

> A quick search reveals https://docs.rs/static_assertions/latest/static_assertions/macro.assert_type_eq_all.html . Can we use that to assert each and every field?
Well, this would assert static types, it would still require to figure out how to get a field's type at compile time.

frb_codegen/src/markers.rs
Unoqwy: The guard `mirror.len() == 1` already assert the length so this cannot panic

fzyzcjy:
> So this sounds quite challenging given we'd need to get the type of fields at compile time but if we can do that then there is no need to have mirror in the first place since we could infer the field types.

Seems not that difficult. We do *not* need to "get the type of fields at compile time". We only need to generate some assertion code for that.

I get an idea: What about generate the following code in the `generated_bridge.rs`. Notice we generate it inside `flutter_rust_bridge_codegen`, so it is quite easy to do following the existing code.

```
const _: () = {
  // this function is NEVER called, and we NEVER need to really construct a `ApplicationSettings` object. But it is sufficient for static checker!
  fn dummy(a: ApplicationSettings) {
    let _: String = a.name; // we ensure a.name field is a string!
    let _: String = a.version;
    let _: ApplicationMode = a.mode;
  }
}
```

Unoqwy:
Yeah, that would indeed work. 

I was just testing something similar:
```
const _: fn() = || {
        fn field_0() -> i32 {
            None::<Test>.unwrap().name
        }
};```
Your way of doing it is cleaner though, let's go with it.

**Comment by Unoqwy** [comment]
Ok so this is working great. By using the type name as var the errors are actually meaningful. (e.g. https://imgur.com/IVaxOCo)

However, I was trying this out in an actual scenario but nested structs are not a thing apparently? The examples seem to all use Vec<> when showcasing struct nesting, is there any technical reason behind it or could it be implemented?

**Comment by fzyzcjy** [comment]
> However, I was trying this out in an actual scenario but nested structs are not a thing apparently? The examples seem to all use Vec<> when showcasing struct nesting, is there any technical reason behind it or could it be implemented?

```
struct A {
  b: Box<B>
}
```

need a `Box` and it should work.

maybe we should add it to doc?

fzyzcjy:
Oh I did not realize this pr is ready for review (again)... Please ping me if I do not reply after one or two hour!

fzyzcjy:
LTGM

fzyzcjy:
Oops a problem: https://github.com/fzyzcjy/flutter_rust_bridge/issues/357

fzyzcjy:
doc added

PR #358, opened 2022-02-19
Code comments and CI related to Corrosion

### LINKED ISSUE #354: Document edge cases regarding Flutter Linux
Thanks to @tmpfs's [comment](https://github.com/fzyzcjy/flutter_rust_bridge/issues/318#issuecomment-1045614165), an issue with running Corrosion on ARM Linux has been found.

Detail: Flutter ARM only ships with Snaps which bundle CMake 3.10 and cannot use FetchContent.

Needs to investigate whether Corrosion itself is compatible with older versions of CMake. I will raise this issue with upstream as needed.

Desdaemon:
Resolves #354.

Opening a new PR since #356 was closed prematurely.


fzyzcjy:
Wonderful job!

PR #359, opened 2022-02-19
Make mirroring work for more use cases

Unoqwy:
Add support for tuple structs, enum variants, wrapping in Vec and Optional.

There are still cases where it's not working, especially with Box (and sending back Vec).
It would also be nice to have it working for StreamSink but I'm not sure how to go about it.

### DIFF
--- a/frb_codegen/src/generator/rust/mod.rs
+++ b/frb_codegen/src/generator/rust/mod.rs
@@ -260,18 +260,8 @@ impl Generator {
             .collect::<Vec<_>>()
             .join("");
 
-        let wrapper = match &func.output {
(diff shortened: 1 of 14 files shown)


fzyzcjy:
Good job! Looking forward to merging the PR

Unoqwy:
I won't be working on this in the next few days, if you want to merge it to add what's already been done you can go ahead.
Either way, I'll work on supporting wrapping for more use cases in the near future since it should still be relevant whenever external crates can be resolved (since it would enable to read struct fields without mirroring but would still face the same issue against orphan rule).

fzyzcjy:
Sure, I will merge it. Looking forward to your further PRs!

PR #363, opened 2022-02-20
CLI improvements

### LINKED ISSUE #344: The `.freezed.dart` is not automatically generated, making the users confused
The freezed declaration file doesn't get generated. The declaration file does include `part 'bridge_generated.freezed.dart';` in the header, but the file doesn't exist. There is no error in the codegen output so I assume this is a bug, probably a side effect of the heavy refactoring going on.
Also, there is no flag to customize the name of this file, it would be a nice addition for consistency (well, once the file actually generates).

Desdaemon:
Resolves #344.

- Added `--verbose` flag
- Added `--dart-root` flag to specify root of Dart library, by default inferred
- Runs `dart run build_runner build` at the Dart root if freezed code is detected, also lint generated files
- Added `--no-build-runner` to skip running build_runner
- Fixed "bridge_generated.freezed.dart" being hardcoded
- Allow specifying multiple `--c-output`

### DIFF
--- a/frb_codegen/src/commands.rs
+++ b/frb_codegen/src/commands.rs
@@ -3,7 +3,7 @@ use std::path::Path;
 use std::process::Command;
 use std::process::Output;
 
-use log::{debug, error, warn};
+use log::{debug, error, warn, info};
 
 /// Known failures that occur from external commands.
 /// If an error occurs frequently enough, consider adding it here and use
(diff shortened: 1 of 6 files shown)


fzyzcjy:
🎉

fzyzcjy:
related: https://github.com/fzyzcjy/flutter_rust_bridge/issues/365

Issue #360, opened 2022-02-19
[Bug] Unable to parse use statements when having `use a::b as c`?


Poly-Pixel:
### Describe the bug

I am working on a Flutter project that uses [libp2p](https://libp2p.io/), and I came across this tool to bind [rust-libp2p](https://github.com/libp2p/rust-libp2p) to the Flutter project. I followed the "Integrating with existing projects" instructions on the docs, using the cloned repository as the crate. I got through the instructions until the last step of running the tool. The logs are shown below. After some looking around, the issue appears to be the `use` statements in the Rust code.

### Codegen logs with `RUST_LOG=debug` env variable

```shell
user@ubuntu:~/Documents/programs/project$ RUST_LOG=debug flutter_rust_bridge_codegen -r rust-libp2p/src/lib.rs -d lib/bridge_generated.dart -c ios/Runner/bridge_generated.h
[2022-02-19T19:56:10Z DEBUG flutter_rust_bridge_codegen::commands] execute command: bin=sh args=["-c", "dart pub global list"] current_dir=None cmd="sh" "-c" "dart pub global list"
[2022-02-19T19:56:10Z DEBUG flutter_rust_bridge_codegen::commands] command="sh" "-c" "dart pub global list" stdout=dartdoc 4.1.0
    ffigen 4.1.3
    protoc_plugin 20.0.0
     stderr=
[2022-02-19T19:56:10Z DEBUG flutter_rust_bridge_codegen::commands] execute command: bin=sh args=["-c", "test -x \"$(which cbindgen)\""] current_dir=None cmd="sh" "-c" "test -x \"$(which cbindgen)\""
[2022-02-19T19:56:10Z DEBUG flutter_rust_bridge_codegen::commands] command="sh" "-c" "test -x \"$(which cbindgen)\"" stdout= stderr=
[2022-02-19T19:56:10Z DEBUG flutter_rust_bridge_codegen::commands] 
[...]

PR #364, opened 2022-02-21
Speculative fix for #360

SecondFlight:
This patch should fix the issues described in #360. It disables import parsing, since the code itself is unfinished and the result isn't being used.


fzyzcjy:
🎉

PR #401, opened 2022-04-13
Ignore prefer_const_constructors in generated code

surban:
This is necessary to avoid warnings in generated Dart code for Rust
enums that have variants with no fields.


fzyzcjy:
Maybe run codegen (https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/CONTRIBUTING.md#appendix-sample-commands-to-run-code-generator) to make CI happy

surban:
Done.

PR #413, opened 2022-04-23
Add brackets to boxed variable

Syndim:
Should fix #412.

PR #414, opened 2022-04-23
Add brackets to boxed variable

Syndim:
My bad, wrong change from the previous PR.

Should fix #413


fzyzcjy:
Wait for the ci

Syndim:
Looks like clippy is complaining when there is only 1 boxed value in the function parameter: https://github.com/fzyzcjy/flutter_rust_bridge/runs/6139942673?check_suite_focus=true

fzyzcjy:
Just add ignores. We have already added tons of ignores for the generated rs file. Just follow it

fzyzcjy:
@all-contributors please add @syndim for code

sagudev:
There are some cases where this causes an error at least in my private project, that's why I have some dirty hack lines in gen part of just file:
```
sed -i -e 's/mirror_SomeStruct((\*v))/mirror_SomeStruct(v)/g' native/src/bridge_generated.rs
```
Will try to create test case later today.

PR #431, opened 2022-04-30
Update config.rs

sagudev:
I'm pretty sure that llvm path should not contain lib as per https://pub.dev/packages/ffigen.

On my computer I used `--llvm-path /usr/lib/llvm-14` in justfile, when changing to `/usr/lib/llvm-14/lib` it does not work.


sagudev:
Can confirm that it is working now.

PR #434, opened 2022-05-01
Make code generator a `lib` to be used in `build.rs`; add error types for codegen; depend on `cbindgen` directly; update docs

### LINKED ISSUE #433: frb_codegen as lib for build.rs
Now codegen needs to be manually launched after changing the code. I believe it would be much simpler to provide fbr_codegen as library too, to allow direct usage in build.rs. This way the regen could be triggered only on api.rs file changed and before checking, so users get right errors from rust-analyzer. Codegen binary could still work as it's working now as it would leverage same codegen lib, meaning that workflow of current users would not be broken.

sagudev:
Fixes #433.
And because it was needed for this PR this changes were also made:
- split codegen into library (lib.rs to be used in build.rs) and binary part (main.rs to allow creating good old frb_codegen binary)
- adding error handling in codegen (using own error type)
- depend on cbindgen directly, instead of depending on it installed by user
- updated docs


fzyzcjy:
By the way, a subsection can be added in http://cjycode.com/flutter_rust_bridge/feature.html doc

sagudev:
@fzyzcjy Should I port with_flutter example too?

And maybe it would be the best to depend on cbindgen from codegen directly, instead of depending on it installed by user

fzyzcjy:
> Should I port with_flutter example too?

No need :) Let users know the raw methods are also good

fzyzcjy:
> And maybe it would be the best to depend on cbindgen from codegen directly, instead of depending on it installed by user

Sounds reasonable! I am not sure why I did not do that before.

sagudev:
There are only two blockers: CI and docs.

sagudev:
CI is done, meanwhile I found this in docs
> llvm_path := if os() == "macos" {
    "--llvm-path /opt/homebrew/opt/llvm"
} else {
    ""
}

as a solution to problem with ffigen missing llvm on macos, but it appears that this same path is already in frb_codegen https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_codegen/src/config.rs#L137 so is this note in docs redundant?
Unfortunately I do not have mac to check it it.

fzyzcjy:
> so is this note in docs redundant?

I guess yes judging from your description. 

/cc @Desdaemon who wrote this part

book/src/tutorial_with_flutter.md
```diff
@@ -27,9 +27,8 @@ git clone https://github.com/fzyzcjy/flutter_rust_bridge && cd flutter_rust_brid
 
 This step is optional, since I have generated the source code already (in quickstart). Even if you do it, you should not see anything changed.
 
-Firstly, you may need to install dependencies of this library:
+Firstly, you may need to install dependencie of this library:
```
fzyzcjy, line 30: typo

frb_codegen/src/main.rs
```diff
@@ -1,196 +1,19 @@
...
-mod config;
-mod generator;
-mod ir;
-mod markers;
-mod others;
-mod parser;
-mod source_graph;
-mod transformer;
-mod utils;
+use lib_lutter_rust_bridge_codegen::{frb_codegen, Opts};
```
fzyzcjy, line 5: typo: "lutter" -> "flutter" (maybe other places also have this typo?)

fzyzcjy:
LGTM except for a bit of typos

fzyzcjy:
https://github.com/all-contributors please add @sagudev for doc

fzyzcjy:
@all-contributors please add @sagudev for doc

Desdaemon:
> > so is this note in docs redundant?
> 
> I guess yes judging from your description. I personally do not need that flag though I am on macos, maybe because my llvm is installed in another way (not very sure).
> 
> /cc @Desdaemon who wrote this part

We did partially fix it by manually supplying the paths, but the documentation for it should stay in my opinion.

PR #441, opened 2022-05-04
update all deps in locks

sagudev:
NOTE: https://github.com/sagudev/flutter_rust_bridge/runs/6277947055?check_suite_focus=true failed two times on my CI

### DIFF
--- a/frb_example/pure_dart/rust/Cargo.lock
+++ b/frb_example/pure_dart/rust/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.16.0"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
+checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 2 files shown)


fzyzcjy:
Sounds great. But the CI now seems to be failing, what about fixing it before merging

fzyzcjy:
In addition, I wonder why you update those deps in locks? How is it related to fixing the failure you mentioned in your ci?

sagudev:
I think current CI problems are from master. Its about `/pure_dart/dart/lib/bridge_generated.freezed.dart` which has nothing to do with my PR but its from previous commit:
slika
So its probably regression from: https://github.com/fzyzcjy/flutter_rust_bridge/pull/438.

As for my CI error, it's gone. Maybe it was just lottery.

sagudev:
> In addition, I wonder why you update those deps in locks? How is it related to fixing the failure you mentioned in your ci?

I updated deps in lock because I noticed (by mistake) that they are outdated. No special reason. It's not related to fixing error on my CI but its rather causing it. But as I said before the failures were probably just random noise.

fzyzcjy:
All right. So shall I merge it or not?

sagudev:
It was really just a noise, it's working now on my fork: https://github.com/sagudev/flutter_rust_bridge/runs/6300502762?check_suite_focus=true

So I think you should merge it.

fzyzcjy:
let me rerun it

image

sagudev:
Did you fix errors in master?

fzyzcjy:
Oh you are right, I will do that

fzyzcjy:
Sorry, provided #446 I have to revert this...

sagudev:
> Sorry, provided #446 I have to revert this...

Ah, no problem.

This locks are just for examples so they can be outdated without a harm.

PR #442, opened 2022-05-05
Add support for usize and [T;N]

trobanga:
I've added support for usize and [T;N] since it's often needed for mirroring external types.
It's still a work in progress and it's not tested properly, yet. But the generated output looks good so far.

Also, I'm not sure if SupportedInnerType::Array is needed or if there's a more elegant way.


fzyzcjy:
Sounds great!

> It's still a work in progress

So I convert it to a "draft pr". Feel free to convert back when you are done

fzyzcjy:
About testing: Please have a look at frb_example/pure_dart and add something into the Rust code as well as some assertions in the dart code.

trobanga:
Ok, I've added some tests and from my side it's ready. However, it still relies on my version of allo-isolate...

trobanga:
This might be a little bit off topic:
I've notices that there are still problems with external types because usually their fields are private. Is there a workaround?

frb_example/pure_dart/dart/lib/main.dart
```diff
@@ -354,6 +354,18 @@ void main(List<String> args) async {
...
+    expect(await api.getArray(), [1, 2, 3, 4, 5]);
+
+    print('dart call getComplexArray()');
+    final points = await api.getComplexArray();
+
+    expect(points[0].x, 1.0);
+    expect(points[1].x, 2.0);
+
+    print('dart call getUsize');
+    expect((await api.getUsize(u: 2)), 2);
```
fzyzcjy, line 367: btw the redundant `()` can be removed

fzyzcjy:
> However, it still relies on my version of allo-isolate...

Make a PR to allo-isolate, and upgrade the dependency version of frb here.

trobanga:
> 

I already did a PR but still wait for the reply

trobanga:
> > I've notices that there are still problems with external types because usually their fields are private. Is there a workaround?
> 
> Do you mean "mirroring"? /cc @Unoqwy who implements that wonderful feature

Yes, for example I'd like to use uuid::Uuid but get: `field `0` of struct `uuid::Uuid` is private`
AFAIK the struct fields have to be pub to enable to create the objects in Dart, but if I only want to read them this requirement could be omitted.

trobanga:
Ok, allo-isolate is ready

fzyzcjy:
1. resolve conflict here
2. https://github.com/sunshine-protocol/allo-isolate/pull/21#discussion_r867269016

then ready to merge :)

trobanga:
>     2. [Add usize and [T; N] sunshine-protocol/allo-isolate#21 (comment)](https://github.com/sunshine-protocol/allo-isolate/pull/21#discussion_r867269016)
I think it's a good idea (and I made already a pull request), however, in my opinion it's unrelated to this pull request. So, can we merge this and change the allo-isolate version later?

fzyzcjy:
> So, can we merge this and change the allo-isolate version later?

Sure. Just do not forget it :)

frb_codegen/src/parser/ty.rs
```diff
@@ -42,11 +42,14 @@ impl<'a> TypeParser<'a> {
...
 
 /// Generic intermediate representation of a type that can appear inside a function signature.
+#[derive(Debug)]
 pub enum SupportedInnerType {
     /// Path types with up to 1 generic type argument on the final segment. All segments before
     /// the last segment are ignored. The generic type argument must also be a valid
     /// `SupportedInnerType`.
     Path(SupportedPathType),
+    /// Array type
+    Array(Box<Self>, usize),
```
fzyzcjy, line 52: When reviewing the code, I am wondering, can we do *not* create SupportedInnerType::Array, but instead do something in the SupportedPathType?

PR #449, opened 2022-05-08
Implement Wire2Api<[T; N]> & refactor dart tests

trobanga:
I've noticed that `[T; N]` is only half way implemented. Rust functions that take `[T; N]` as an argument don't generate and functions that take a struct with an array field don't build. 
The first fails with `cannot parse argument...` and the latter with `the trait Wire2Api<> is not implemented.
I works if I manually add
```rust
impl<const N: usize> Wire2Api<[u8; N]> for *mut wire_uint_8_list {
   fn wire2api(self) -> [u8; N] {
       unsafe {
           let wrap = support::box_from_leak_ptr(self);
           std::slice::from_raw_parts(wrap.ptr, wrap.len as usize)
               .try_into()
               .unwrap()
       }
   }
}
``` 
into bridge_generated.rs. 

To my knowlegde there don't exist fixed length lists in Dart, only lists with `growable = false`, so I'm not sure what's the best approach here...any ideas?


fzyzcjy:
> To my knowlegde there don't exist fixed length lists in Dart, only lists with growable = false, so I'm not sure what's the best approach here...any ideas?

Maybe throw runtime exception if the user provides the wrong length?

By the way, I guess you can use "delegate" design pattern, and `[T;N]` just delegates to `Vec<T>`.

trobanga:
@fzyzcjy it works so far...but valgrind explodes and I don't understand what's happening...

fzyzcjy:
> error: valgrind_util find this line is bad: ==5753==    definitely lost: 28 bytes in 2 blocks

You have memory leak.

fzyzcjy:
```
==5753== 12 bytes in 1 blocks are definitely lost in loss record 3 of 185
==5753==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==5753==    by 0x57B45EB: alloc::alloc::alloc_zeroed (alloc.rs:158)
==5753==    by 0x57B468C: alloc::alloc::Global::alloc_impl (alloc.rs:169)
==5753==    by 0x57B47DC: <alloc::alloc::Global as core::alloc::Allocator>::allocate_zeroed (alloc.rs:234)
==5753==    by 0x581FBA9: alloc::raw_vec::RawVec<T,A>::allocate_in (raw_vec.rs:186)
==5753==    by 0x581DC6B: alloc::raw_vec::RawVec<T,A>::with_capacity_zeroed_in (raw_vec.rs:140)
==5753==    by 0x583E24E: <T as alloc::vec::spec_from_elem::SpecFromElem>::from_elem (spec_from_elem.rs:54)
==5753==    by 0x5855A18: alloc::vec::from_elem (mod.rs:2363)
==5753==    by 0x583A0EC: flutter_rust_bridge::support::new_leak_vec_ptr (support.rs:15)
==5753==    by 0x57E02D1: new_uint_32_list (bridge_generated.rs:1144)
==5753==    by 0x50DC1C1: ???
==5753==    by 0x50DC164: ???
==5753== 
==5753== 16 bytes in 1 blocks are definitely lost in loss record 4 of 185
==5753==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==5753==    by 0x57B44DB: alloc::alloc::alloc (alloc.rs:87)
==5753==    by 0x57B4676: alloc::alloc::Global::alloc_impl (alloc.rs:169)
==5753==    by 0x57B4799: <alloc::alloc::Global as core::alloc::Allocator>::allocate (alloc.rs:229)
==5753==    by 0x5821853: alloc::raw_vec::RawVec<T,A>::allocate_in (raw_vec.rs:185)
==5753==    by 0x581DA6B: alloc::raw_vec::RawVec<T,A>::with_capacity_in (raw_vec.rs:132)
==5753==    by 0x584F972: alloc::vec::Vec<T,A>::with_capacity_in (mod.rs:609)
==5753==    by 0x57FEFE0: <T as alloc::vec::spec_from_elem::SpecFromElem>::from_elem (spec_from_elem.rs:14)
==5753==    by 0x5855930: alloc::vec::from_elem (mod.rs:2363)
==5753==    by 0x583A360: flutter_rust_bridge::support::new_leak_vec_ptr (support.rs:15)
==5753==    by 0x57E0211: new_list_point (bridge_generated.rs:1126)
==5753==    by 0x50DC5E9: ???
```

They are allocated from these places, and never de-allocated.

fzyzcjy:
What about looking at `bridge_generated.rs:1144` and `bridge_generated.rs:1126` and consider what is happening - why it allocates memory without free?

trobanga:
Those are functions that are generated by the Vec<> implementation which I now delegate to.
I'm a little bit lost here...I can look into it but any help would be much appreciated :)

fzyzcjy:
Currently I do not have much time in digging deep into it (I am busy with https://github.com/fzyzcjy/flutter_convenient_test and my project and research etc). The general suggestion is that, look at the generated code. There is no magic - no hidden code behind macro or something, so only look at bridge_generated.rs and a few dependencies are enough.

trobanga:
This seems to work. The only issue I've found is that `just test-pure` does not close after it's finished...let's see what the integration tests say...

fzyzcjy:
Great!

valgrind runs for 4hours... let me try to cancel&rerun

trobanga:
Ok, panicking in Rust does not work because then the other function arguments are not handled and leak memory.
Throwing exceptions in Dart prevents - for whatever reason - the test from returning, even tearDown is called, so I've no clue what's happening here

trobanga:
So, I've found a way to have the tests finish by using dart test. However, I wasn't able to feed the library name via command line, thus, I've put it directly into `main.dart`. On my terminal I can clearly see All tests passed! I think the valgrind command would need an update to work with that.

trobanga:
Then there's a new problem that my machine generated u64 where UnsignedLong should be...any idea what's happening there?

trobanga:
@fzyzcjy I'd like to finish this pull request soon. The code works and the tests pass. If run with `dart test` the tests even finish correctly. 
That my machine generates `ffi.Uint64` instead of `ffi.UnsignedLong` is not really an issue, I'd say.

fzyzcjy:
Oh I forget to reply to your comments above...

> That my machine generates ffi.Uint64 instead of ffi.UnsignedLong is not really an issue, I'd say.

Well... Then what about manually edit it so CI passes...

trobanga:
> Well... Then what about manually edit it so CI passes...

Yes, I was thinking about that too.

trobanga:
But what about the test issue? Would using `dart test` be an option?

fzyzcjy:
`dart test` runs successfully != no memory leak. valgrind checks whether there are memory problems. so it is possible that your code really has bug that leaks memory...

```
==6234== HEAP SUMMARY:
==6234==     in use at exit: 9,377 bytes in 171 blocks
==6234==   total heap usage: 952 allocs, 781 frees, 798,436 bytes allocated
==6234== 
==6234== 288 bytes in 1 blocks are possibly lost in loss record 170 of 171
==6234==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==6234==    by 0x40149DA: allocate_dtv (dl-tls.c:286)
==6234==    by 0x40149DA: _dl_allocate_tls (dl-tls.c:532)
==6234==    by 0x4869322: allocate_stack (allocatestack.c:622)
==6234==    by 0x4869322: pthread_create@@GLIBC_2.2.5 (pthread_create.c:660)
==6234==    by 0x40E9E2: ??? (in /home/runner/work/flutter_rust_bridge/flutter_rust_bridge/frb_example/pure_dart/dart/main.exe)
==6234==    by 0x2D8E2D: ??? (in /home/runner/work/flutter_rust_bridge/flutter_rust_bridge/frb_example/pure_dart/dart/main.exe)
==6234==    by 0x2D7CF0: ??? (in /home/runner/work/flutter_rust_bridge/flutter_rust_bridge/frb_example/pure_dart/dart/main.exe)
==6234==    by 0x2C615C: ??? (in /home/runner/work/flutter_rust_bridge/flutter_rust_bridge/frb_example/pure_dart/dart/main.exe)
==6234==    by 0x2CA2F2: ??? (in /home/runner/work/flutter_rust_bridge/flutter_rust_bridge/frb_example/pure_dart/dart/main.exe)
==6234==    by 0x2CAA18: ??? (in /home/runner/work/flutter_rust_bridge/flutter_rust_bridge/frb_example/pure_dart/dart/main.exe)
==6234==    by 0x49F6082: (below main) (libc-start.c:308)


==6234== LEAK SUMMARY:
==6234==    definitely lost: 0 bytes in 0 blocks
==6234==    indirectly lost: 0 bytes in 0 blocks
==6234==      possibly lost: 288 bytes in 1 blocks
==6234==    still reachable: 9,089 bytes in 170 blocks
==6234==         suppressed: 0 bytes in 0 blocks
```

trobanga:
The problem here is this: `Invalid argument(s): Failed to load dynamic library '../rust/target/debug/libflutter_rust_bridge_example.so': ../rust/target/debug/libflutter_rust_bridge_example.so: cannot open shared object file: No such file or directory`

I wasn't able to add a command line argument to  `dart test` so I hard-coded it into `main.dart`. I guess for this the valgrind command in CI needs to be adapted...

We could also start a new test directory which would be called with `dart test`, then the previous workflow can stay the way it is.

fzyzcjy:
> The problem here is this: Invalid argument(s): Failed to load dynamic library '../rust/target/debug/libflutter_rust_bridge_example.so': ../rust/target/debug/libflutter_rust_bridge_example.so: cannot open shared object file: No such file or directory

I am confused... Why this fails, since it passes CI before?

fzyzcjy:
Btw I will fix CI failure on master https://github.com/fzyzcjy/flutter_rust_bridge/issues/467

trobanga:
The problem here is this: `Invalid argument(s): Failed to load dynamic library '../rust/target/debug/libflutter_rust_bridge_example.so': ../rust/target/debug/libflutter_rust_bridge_example.so: cannot open shared object file: No such file or directory`

I wasn't able to add a command line argument to  `dart test` so I hard-coded it into `main.dart`. I guess for this the valgrind command in CI needs to be adapted...

We could also start a new test directory which would be called with `dart test`, then the previous workflow can stay the way it is.



> > The problem here is this: Invalid argument(s): Failed to load dynamic library '../rust/target/debug/libflutter_rust_bridge_example.so': ../rust/target/debug/libflutter_rust_bridge_example.so: cannot open shared object file: No such file or directory
> 
> I am confused... Why this fails, since it passes CI before?

It's because I hard-coded the path into `main.dart`. I'm sorry if I wasn't clear about this. I had to do it because I can't add command line args to `dart test`. Maybe it's possible somehow but I don't know how. That's why I'm suggesting to start a new test directory so that the previous workflow is not disrupted.

fzyzcjy:
> cannot open shared object file: No such file or directory

Why cannot open this .so file ...?

trobanga:
I think it has to be `/home/runner/debug/libflutter_rust_bridge_example.so` for CI

fzyzcjy:
> I think it has to be /home/runner/debug/libflutter_rust_bridge_example.so for CI

Why...? Seems that it worked fine before (?)

trobanga:
Because in my current version the library path is hard-coded `main.dart`. The CI passes the above path into `main.dart` but there it's replaced/overridden. Sorry, for the confusion...

fzyzcjy:
I mean, just look at the master branch: It passes valgrind CI (https://github.com/fzyzcjy/flutter_rust_bridge/runs/6518709784?check_suite_focus=true), and it does not hardcode path(https://github.com/fzyzcjy/flutter_rust_bridge/blob/4881928400ccabf9a87b188340aa1e7afdac5309/frb_example/pure_dart/dart/lib/main.dart#L10). "Implement the `[T;N]`" should not need to change anything about that.

trobanga:
The problem is that `dart main.dart` does not finish when running with a test that throws an exception if an array has the wrong size (most likely a bug in Dart - see the beginning of this thread). `dart test main.dart` does but I cannot pass command line args to `dart test`. That's why I've hard coded it into main.dart which is of course not the final solution. 

So the options are: 
1. write tests that run with `dart test` in the CI
2. don't run the tests in CI 
3. use an environment variable to pass the dylib path into main.dart
4. don't check in Dart code for correct array length. Then the Rust code might panic which results in an FfiException. Sounds best but might leak memory if there are other function arguments because due to panicking those aren't processed...

fzyzcjy:
> does not finish when running with a test that throws an exception if an array has the wrong size (most likely a bug in Dart - see the beginning of this thread)

maybe `catch` it?

trobanga:
This of coure I do. One of the offending tests is:
```rust
    print('dart call returnStructWithArray() with wrong sized list');
    {
      try {
        var a = MyArray(a: Uint32List.fromList([1, 2]), b: Uint16List.fromList([1]));
        await api.takeAndUnpackArray(a: a);
        fail("exception not thrown");
      } catch (e) {
        print(e);
        expect(e, isA<Exception>());
      }
    }
```
It passes and at the end I get `All tests passes` but the program does not quit.

fzyzcjy:
> It passes and at the end I get All tests passes but the program does not quit.

That is weird. Could you please show logs of that?

In addition, try:

```
expect(() => yourFunction(), throws(...));
```

trobanga:
> > It passes and at the end I get All tests passes but the program does not quit.
> 
> That is weird, and is the problem that should be fixed. Could you please show logs of that?
> 
> In addition, try:
> 
> ```
> expect(() => yourFunction(), throws(...));
> ```
I tried that...same problem

What logs do you mean? The current version is one that never stops...

fzyzcjy:
I also cannot understand... Without a deep dig into your code, I am not sure why it runs forever. Maybe make a minimal reproducible sample and ask in the Dart repository or StackOverflow?

A wild guess: there is some Future that is pending and never finished.

trobanga:
If it runs with `dart test` it finishes. I've read about similar problems with dart test if `--coverage` is enabled.
I think it's a bug in Dart and we should definately tell the Dart people but at the same time I'd like to finish this pull request.
In general I think it'd be good style, if the tests would be in tests/, called with `dart test`, and each test would have it's own test() function.

fzyzcjy:
> if the tests would be in tests/, called with dart test, and each test would have it's own test() function.

I guess you mean modify https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_example/pure_dart/dart/lib/main.dart, such that:

1. `lib/main.dart` goes to `test/main.dart`
2. each cluster of API calls become a `test`, instead of currently everything is in one giant `test`

Agree!

fzyzcjy:
However, ensure valgrind scripts pass.

trobanga:
It seems I was very much wrong and it had nothing to to with Dart test but it got stuck in executeNormal's callFfi. However, I've found a solution but I had to dig very deep and it's not yet working for struct fields.
The main problem is that the variables have to be evaluated before they are passed into the ffi call. I think I took a wrong turn in checking the array length in wire2api  and returning a `Result` which has to be checked again...instead I should do the length check in Dart.

fzyzcjy:
Great work and congratulations on finding it! Now we see the value of enforcing CI to pass before merging the PR: there is really a bug.

trobanga:
Indeed! 
But I am questioning if it's worth to finish implementing [T; N] because there is still some work to be done before it's ready and I don't really see the benefits of it anymore . In pure Rust [T; N] is checked during compilation but if runtime checks are necessary one can use Vec as well.

fzyzcjy:
Agree. Indeed in order to implement that "implement [T;N] using Vec", you may be interested in this proposal as well: https://github.com/fzyzcjy/flutter_rust_bridge/issues/475#issuecomment-1140099008.

IMHO your refactor for test dart file can be quite helpful :)

trobanga:
> Agree. Indeed in order to implement that "implement [T;N] using Vec", you may be interested in this proposal as well: [#475 (comment)](https://github.com/fzyzcjy/flutter_rust_bridge/issues/475#issuecomment-1140099008).
> 
Sounds interesting, maybe I can look into that but I'll need a break for a few days^^

> IMHO your refactor for test dart file can be quite helpful :)
I'll make a new PR

fzyzcjy:
Take your time :) Looking forward to your contributions!

PR #452, opened 2022-05-11
Non-final fields in Dart structs.

surban:
Use the #[frb(non_final)] attribute on a field in a Rust struct to
make the corresponding field in the Dart struct non-final.

### DIFF
--- a/frb_codegen/src/ir/field.rs
+++ b/frb_codegen/src/ir/field.rs
@@ -4,6 +4,6 @@ use crate::ir::*;
 pub struct IrField {
     pub ty: IrType,
     pub name: IrIdent,
-    pub non_final: bool,
+    pub is_final: bool,
     pub comments: Vec<IrComment>,
 }
--- a/frb_codegen/src/ir/field.rs
+++ b/frb_codegen/src/ir/field.rs
@@ -4,5 +4,6 @@ use crate::ir::*;
 pub struct IrField {
     pub ty: IrType,
     pub name: IrIdent,
+    pub non_final: bool,
     pub comments: Vec<IrComment>,
 }
--- a/frb_example/pure_dart/dart/lib/bridge_generated.dart
+++ b/frb_example/pure_dart/dart/lib/bridge_generated.dart
@@ -25,11 +25,7 @@ abstract class FlutterRustBridgeExample {
   /// Newlines are preserved.
   ///
   Future<int> primitiveTypes(
-      {required int myI32,
-      required int myI64,
-      required double myF64,
-      required bool myBool,
-      dynamic hint});
+      {required int myI32, required int myI64, required double myF64, required bool myBool, dynamic hint});
 
   Future<int> primitiveU32({required int myU32, dynamic hint});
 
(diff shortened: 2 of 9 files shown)


frb_codegen/src/ir/field.rs
```diff
@@ -4,5 +4,6 @@ use crate::ir::*;
 pub struct IrField {
     pub ty: IrType,
     pub name: IrIdent,
+    pub non_final: bool,
```
fzyzcjy, line 7: Shall we name it `final` instead of `non_final` to avoid double negation :)

frb_example/pure_dart/dart/lib/bridge_generated.dart
```diff
@@ -25,11 +25,7 @@ abstract class FlutterRustBridgeExample {
   /// Newlines are preserved.
   ///
   Future<int> primitiveTypes(
-      {required int myI32,
-      required int myI64,
-      required double myF64,
-      required bool myBool,
-      dynamic hint});
+      {required int myI32, required int myI64, required double myF64, required bool myBool, dynamic hint});
```
fzyzcjy, line 28: We now use line-length 80, so need to run code formatter

frb_example/pure_dart/dart/lib/bridge_generated.dart
```diff
@@ -39,22 +35,17 @@ abstract class FlutterRustBridgeExample {
 
   Future<Uint8List> handleVecU8({required Uint8List v, dynamic hint});
 
-  Future<VecOfPrimitivePack> handleVecOfPrimitive(
-      {required int n, dynamic hint});
+  Future<VecOfPrimitivePack> handleVecOfPrimitive({required int n, dynamic hint});
```
fzyzcjy, line 38: Btw please add a test in frb_example/pure_dart/rust as well as frb_example/pure_dart/dart

fzyzcjy (changes requested):

fzyzcjy:
Looks good! Except for a bit of change

frb_codegen/src/ir/field.rs
surban: final is a reserved keyword in Rust. This is why I stuck with non_final.

frb_example/pure_dart/dart/lib/bridge_generated.dart
surban: The Rust part was already there: https://github.com/surban/flutter_rust_bridge/blob/master/frb_example/pure_dart/rust/src/api.rs#L389

I've added a test in the Dart code.

frb_codegen/src/ir/field.rs
fzyzcjy: what about `is_final`?

surban: Ok, I'll change it.

fzyzcjy:
By the way I am confused: why no code is added to frb_example/pure_dart/rust, i.e. no rust test code added?

image

fzyzcjy:
LGTM. Will merge when CI passes

fzyzcjy (approved):

surban:
Somebody already added a test case, but the feature was never implemented.

image

See pure_dart/rust/src/api.rs line 389.

surban:
CI seems to be failing due to other reasons, i.e. `lib/main.dart:50:20: Error: 'Size' is imported from both 'dart:ffi' and 'dart:ui'.`

fzyzcjy:
🎉

PR #463, opened 2022-05-17
Enhance: metadata annotation for structs

### LINKED ISSUE #459: Allow rust struct generating freezed class in dart
In rust we often define:
```rust
struct UserId(value: u64);
```
And use UserId in HashMap as keys.

Since tuple is not well supported in dart, we can define:
```rust
struct UserId { value: u64 }
```
And use it in maps/dicts as keys

But frb currently doesn't generate `==` and `hashCode` method for UserId in dart/flutter, so we can't use UserId as keys in dart, we had to unwrap it and use u64 as key type, it's not quit convient.

If we support generate @freezed (or even better any annotation) on struct, then we got `==` and `hashCode`, and can use it directly in maps.

alanlzhang:
close #459


alanlzhang:
CHANGELOG.md updated.

alanlzhang:
@fzyzcjy Please approve the workflow since this is my first PR for frb ^_^

frb_codegen/src/ir/import.rs
```diff
@@ -0,0 +1,5 @@
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct IrImport {
```
fzyzcjy, line 2: After reading example, I realized it is not "import" in Rust, but instead in Dart. Shall we rename `IrImport -> IrDartImport` and `IrAnnotation -> IrDartAnnotation`?

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -537,3 +537,15 @@ pub fn get_complex_array() -> [Point; 2] {
 pub fn get_usize(u: usize) -> usize {
     u
 }
+
+/// Example for @freezed and @meta.immutable
+#[frb(metadata=("freezed", "immutable" import "package:meta/meta.dart" as meta))]
```
fzyzcjy, line 542: I guess this means *annotations* for dart classes? If so, what about remaming "metadata=..." to "dart_annotation=..."

frb_example/pure_dart/dart/lib/main.dart
```diff
@@ -372,6 +368,10 @@ void main(List<String> args) async {
     customized.nonFinalField = "changed";
     expect(customized.nonFinalField, "changed");
 
+    print('dart call next_user_id');
```
fzyzcjy, line 371:
```suggestion
    print('dart call next_user_id to test metadata annotations');
```

frb_codegen/src/parser/mod.rs
```diff
@@ -213,6 +215,137 @@ fn extract_comments(attrs: &[Attribute]) -> Vec<IrComment> {
...
+            .collect();
+        Ok(Self(imports))
+    }
+}
+
+enum FrbOption {
+    Mirror(MirrorOption),
+    NonFinal,
+    Metadata(NamedOption<frb_keyword::metadata, MetadataAnnotations>),
+}
```
fzyzcjy, line 316: If I understand correctly, you successfully parsed *all* rust macros? That sounds great!

Seems that, the mirror/nonfinal/... already have their own parsers to parse them. Thus are we duplicating works here? Maybe we can refactor those?

fzyzcjy:
Great work! Only some small changes and I am ready to merge it.

frb_codegen/src/parser/mod.rs
alanlzhang: Yes, I try to make an uniform way to parse parameters in "[frb]" macros. I think we can refactor code to use it. But I am new here, so I think it's better to move things slowly and do it later. ^_^

frb_codegen/src/ir/import.rs
alanlzhang: Sure. :-)

fzyzcjy:
Oh, one more thing: Would you mind add a subsection in the documentation :) For example, sibling to: http://cjycode.com/flutter_rust_bridge/feature/handler.html

frb_example/pure_dart/rust/src/api.rs
fzyzcjy: Good point! I did not realized they call the `int` in `int a = ...` a type annotation before. So `dart_metadata` is quite reasonable.

frb_codegen/src/parser/mod.rs
fzyzcjy: Sure. Look forward to your future PRs!

alanlzhang:
Ok. And I think just add some description in "struct" feature page because we can only add dart metadata annotations on rust struct.

fzyzcjy:
Agree. Btw you can "resolve conversation" above when done.

alanlzhang:
Add some line to the lang_struct.md.

fzyzcjy:
Wait for the CI and I will merge :)

fzyzcjy:
@all-contributors please add @alanlzhang for doc

PR #476, opened 2022-05-28
Separate dart tests

trobanga:


frb_example/pure_dart/dart/lib/main.dart
```diff
@@ -6,204 +6,179 @@ import 'package:flutter_rust_bridge_example/bridge_generated.dart';
...
-    print('dart call primitiveTypes');
-    {
-      expect(await api.primitiveTypes(myI32: 123, myI64: 10000000000000, myF64: 12345678901234567890.123, myBool: true),
-          42);
-    }
+  String dylibPath = args[0];
+  print('flutter_rust_bridge example program start (dylibPath=$dylibPath)');
+  print('construct api');
+  final dylib = DynamicLibrary.open(dylibPath);
+  final api = FlutterRustBridgeExampleImpl(dylib);
```
fzyzcjy, line 13: Shall we make these a `setUp` or `setUpAll`?

fzyzcjy (commented):
LGTM and wait for CI

frb_example/pure_dart/dart/lib/main.dart
trobanga: I would leave it as it is. setUp is called before each function and setUpAll would require to use `late api;` which just adds more lines with no more functionality...

fzyzcjy: Well `setUpAll ` not exactly the same - it will be run at different time than this. But anyway this is also good.

trobanga: It might be different if called with `dart test` or what's the difference?

fzyzcjy: setUpAll is called when tests are starting, i.e. after `test()` is executed.

Issue #496, opened 2022-06-08
[Bug] Runtime error calling functions across bridge in iOS release mode


banool:
### Describe the bug

When my app calls into the bridge functions in the debug mode of my app on iOS, it works correctly (simulator or real device), however in release mode it fails with errors like these:
```
Invalid argument(s): Failed to lookup symbol 'store_dart_post_cobject': dlsym(RTLD_DEFAULT, store_dart_post_cobject): symbol not found
```
```
Exception: Subclasses of `FlutterRustBridgeBase` should be singletons - there should not be two instances (runtimeType=NativeImpl)
```
The first one only ever appears once, then the rest are always the second error. Sometimes I only see the second error.

To get to this point I've followed all the steps in the docs. I've also tried to look through all the issues in this repo.

### Codegen logs with `RUST_LOG=debug` env variable

```shell
cd native && cargo build && cargo build --release && cd .. && RUST_LOG=debug just >& /tmp/eh
```

Output: https://gist.github.com/banool/3d1143f9acb02bb79f9cff2011bcc323.

### To Reproduce

```
git clone git@github.com:banool/aclip.git
cd frontend
cd native
cargo build
cargo build --release
cd ..
just
flutter run -d 'My Real iPhone' --release
```

Once in the app, go to `Settings -> Offline download errors` to see the errors pasted above.

### Expected behavior

You can see the expected behavior by running the above repro but without `--release`.

### Generated binding code

See https://github.com/banool/aclip/.

### OS

MacOS 12.4

### Version of `flutter_rust_bridge_codegen`

1.34.1

### Flutter info

```shell
https://gist.github.com/5ecf66d7f10d20932a4db7072464c232
```

### Version of `clang++`

13.1.6

### Version of `ffigen`

5.0.1

### Additional context

Linking a possibly (but probably not) related issue that I've also been struggling with a lot: https://github.com/bbqsrc/cargo-ndk/issues/65.

fzyzcjy:
> Invalid argument(s): Failed to lookup symbol 'store_dart_post_cobject': dlsym(RTLD_DEFAULT, store_dart_post_cobject): symbol not found

Guess ios just think the symbol is unused and thus strip it. Try http://cjycode.com/flutter_rust_bridge/integrate/ios_headers.html.

In addition, try this template project http://cjycode.com/flutter_rust_bridge/template.html.

[...]

PR #502, opened 2022-06-08
Update iOS headers section of book with code that ensures symbols do not get stripped

banool:
For context, see https://github.com/fzyzcjy/flutter_rust_bridge/issues/496.


fzyzcjy:
@all-contributors please add @banool for doc

PR #510, opened 2022-06-12
cleaned up `generate()` in `dart/mod.rs`

w-ensink:
- split the big function up into some smaller functions with descriptive names

Feedback is welcome 🙂

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,10 +72,10 @@ jobs:
           echo UDID=$UDID
           xcrun simctl boot "${UDID:?No Simulator with this name found}"
 
-      - uses: subosito/flutter-action@v1
+      - uses: subosito/flutter-action@v2
         with:
           channel: "stable"
-          architecture: arm64
+          architecture: x64
 
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


fzyzcjy:
LGTM, good work :)

Btw CI is failing (though not due to this PR). Do you have interest in fixing it, such that after the CI passes I will merge this PR.

w-ensink:
I would love to, but I can't really tell what went wrong on the CI, the errors are really vague... If I look at the commit history it started failing at a point when you just updated some docs, so it isn't really clear what is causing it to fail.

fzyzcjy:
Yes. I suspect it is because, either Flutter upgrades, or GitHub Actions runtime upgrades.

Anyway, maybe fire an issue at proper places - StackOverflow / github issues of flutter.

> /Users/runner/hostedtoolcache/flutter/3.0.2-stable/x64/bin/internal/shared.sh: line 225: /Users/runner/hostedtoolcache/flutter/3.0.2-stable/x64/bin/cache/dart-sdk/bin/dart: Bad CPU type in executable
Error: Process completed with exit code 1.

w-ensink:
Yeah, maybe GitHub started using M1 based Macs. I'll try installing rosetta before running the tests.

fzyzcjy:
Btw you may use a shell command to print out "is this a x86 or M1 macos", so we will know exactly

w-ensink:
Good point, but it seems like they are using `x86_64` still, since it says rosetta is not available on the system.

fzyzcjy:
Great job!

PR #520, opened 2022-06-16
Added event listener tests

w-ensink:
- with an event listener that outlives its creation function by being saved in a global
- test creating events after registering a listener and checking if listener is called
    - expected to fail with no waiting
    - expected to fail with (sync) `sleep()` waiting
    - expected to succeed with async `Future.delayed()` waiting

Also added a `run_macos.sh` script to run the `pure_dart` example on macOS, mainly targeted at devs working on that part of the code base.


fzyzcjy:
> Also added a run_macos.sh script to run the pure_dart example on macOS, mainly targeted at devs working on that part of the code base.

Maybe just modify the `justfile`?

w-ensink:
> Maybe just modify the justfile?
Maybe. I haven't used the justfile, I haven't really looked in to what that is.

> Hmm CI fails...
Yeah.. mostly due to formatting and a restriction on changing files? I doesn't really make a lot of sense to me.
The Valgrind is even more confusing...

w-ensink:
Thanks for the tips, I'll try fixing it. 
Valgrind was mainly confusing to me because of the immense log. I just found out that only 1 test is failing. Which is still a bit of a weird one...

The `sleep()` version doesn't register the listener in time, but the "no wait at all" version does register in time on Linux apparently...

fzyzcjy:
> Valgrind was mainly confusing to me because of the immense log. I just found out that only 1 test is failing. Which is still a bit of a weird one...

Try to also look at previous successful CI runs and compare them. 

You can skip some logs about "invalid address access" b/c seemingly caused by Dart not our lib. But look at "memory leaked".

fzyzcjy:
I strongly suspect the "memory leak" error is caused by the `lazy_static` you introduced in the test code. As we know, lazy_static memory is never de-allocated so look like a leak.

w-ensink:
Hmm Valgrind seems to have passed without me even modifying it...

fzyzcjy:
Interesting... If it always succeed from now on then it is ok :)

w-ensink:
It does make it look like it is kind of undefined... Maybe the test where I expect the callback not to be called are not very deterministic...

fzyzcjy:
Btw I am considering whether it is suitable to assert "the callback will not be called"? That does not sound like a *feature*.

w-ensink:
I agree, maybe these two should be removed. The intention was to describe the behavior to potential users running into the same problem I was, but the behavior seems to not be so consistent in these cases...
I'll remove them and update the PR.

dbsxdbsx:
@fzyzcjy @w-ensink, I have a side question here. When using the sync delay in dart , say `sleep(const Duration(milliseconds: 100));`, does it intend to stick the backend (the rust part)? Aas far as I know, the backend(rust side) is using a threadpool, which should not be stuck by the frontend thread(dart side). reference from [here](http://cjycode.com/flutter_rust_bridge/feature/concurrency.html).  Am I wrong?

fzyzcjy:
Dart's sleep will not affect Rust's running code, but will affect other Dart code such as registering a callback etc.

Anyway the merged part of this PR does not test such sync delay.

PR #516, opened 2022-06-15
feature: multiple blocks of files in one command

dbsxdbsx:
## what's new in this pr
In this pr, I add a feature to just use one `flutter_rust_bridge_codegen ` command to generate ALL essential FFI codes, whatever how many rust Api blocks users take.

For example, if there are 2 rust Api blocks, say `rule_api` and `ucci_api` . Now the FFI codes can be generated from this:
```
gen:
  # api for chess rule
    export REPO_DIR="$PWD"; cd /; flutter_rust_bridge_codegen {{llvm_path}} \
        --rust-input "$REPO_DIR/native/src/rule_api.rs" \
        --dart-output "$REPO_DIR/lib/gened_rule_api.dart" \
        --rust-output "$REPO_DIR/native/src/gened_rule_api.rs" \
        --class-name RuleApi \

    # api for connecting ucci engine
    export REPO_DIR="$PWD"; cd /; flutter_rust_bridge_codegen {{llvm_path}} \
        --rust-input "$REPO_DIR/native/src/ucci_api.rs" \
        --dart-output "$REPO_DIR/lib/gened_ucci_api.dart" \
        --rust-output "$REPO_DIR/native/src/gened_ucci_api.rs" \
        --class-name UcciApi \
        --exclude-sync-execution-mode-utility true
```
to this:
```
gen:
    # union version
    export REPO_DIR="$PWD"; cd /; flutter_rust_bridge_codegen {{llvm_path}} \
        --rust-input "$REPO_DIR/native/src/rule_api.rs" "$REPO_DIR/native/src/ucci_api.rs" \
        --dart-output "$REPO_DIR/lib/gened_rule_api.dart" "$REPO_DIR/lib/gened_ucci_api.dart" \
        --rust-output "$REPO_DIR/native/src/gened_rule_api.rs" "$REPO_DIR/native/src/gened_ucci_api.rs" \
        --class-name RuleApi UcciApi \
        # --exclude-sync-execution-mode-utility false true
```

## why need this

1. For further check on Api signature duplication (for both rust and dart side) when using multiple rust blocks.
2. For less code, which makes the `justfile`  more easy and readable for users to manipulate.

## need to mention

1. I changed several fields of `RawOpts` from `Option<x>` to `Option<Vec<x>>` . The reason is for logic simplicity among those non-essential flags. That is, **there are only 2 options for these flags**,  (say: `class-name`) either submitting all data for each api block, or submitting no data at all --- **No way in the middle**.
[...]

PR #527, opened 2022-06-22
Avoid conflicted and redundant FFI with generation over multiple blocks of rust Api

dbsxdbsx:
## What's new 

1. For Api defined by users explicitly, now frb could check confliction over Api defined in different blocks.
2. Flag `exclude_sync_execution_mode_utility` is needless now, all related code is deleted. And in rust, only the first block would contain  Api `free_WireSyncReturnStruct`.
3. In dart, after introducing https://github.com/fzyzcjy/flutter_rust_bridge/pull/516, ALL rust Api are generated in EVERY generated dart file, within class like `class xxxWire implements FlutterRustBridgeWireBase {`, even some of them are not used for the block.  Now this is fixed to a great degree.
4. For implicit Api, like `new_uint_8_list`, which is automatically generated when there is an Api with `String` as a parameter, it would raise conflicts with multiple blocks. Now it is fixed over blocks, by making each conflicted Api with a number suffix. For example, If there are 2 blocks both needing this Api, then the Api would be defined as `new_uint_8_list_1` and `new_uint_8_list_2` respectivley. 

## Thoughts on avoiding FFI conflicts over blocks
There are generally two approaches to avoid this:

1. Put the conflicted Api into another shared file.
2. Give each conflicted Api a  suffix to make them not the same among all blocks.

The 1st approach needs to add new file(s) and raise complex in building logic. Also, this way seems to be not a good practice especially in a big project with different teams in charge of different blocks--- it is not decoupling.
The only advantage of the 1st is that it has less code.  And the 2ed approach just have the opposite features.
Therefore, for similar usage as before, for decoupling,  and for not too much logic modification in frb, I chosen the 2ed approach.

## Todos
for feature 1, now it can't do checking over conflicts among the same block. May do it later.
for feature 3, implicitly defined Api, like `new_uint_8_list_1`,`new_uint_8_list_2` would still be written to ALL generated dart files, it is not a big deal, but not perfect.
Some test examples over multiple blocks are needed, like that in `flutter_rust_bridge\frb_example\pure_dart\rust`. And I do think some "should panic" examples are also needed--- How to make it, does it need some modification over GitHub CI test?


fzyzcjy:
Good job! Will review once everything is finished and CI passes

fzyzcjy:
> And I do think some "should panic" examples are also needed--- How to make it, does it need some modification over GitHub CI test?

Agree. Maybe need some modification I guess. Feel free to modify `.github` folder, and your change to CI will also be run and you can see it in the PR's panel

fzyzcjy:
> like new_uint_8_list_1,new_uint_8_list_2 would still be written to ALL generated dart files, it is not a big deal, but not perfect.

This will result in bigger generated code size. As we know, when building android/ios app, app size is very important.

dbsxdbsx:
> > like new_uint_8_list_1,new_uint_8_list_2 would still be written to ALL generated dart files, it is not a big deal, but not perfect.
> 
> This will result in bigger generated code size. As we know, when building android/ios app, app size is very important.

Frankly, I didn't find a good way to excluding the implicit symbols, seems needing another extraction over `generator.extern_func_collector.names`. Or is there a better idea?

fzyzcjy:
Maybe yes, at least need to create an intermediate representation instead of directly output some string

dbsxdbsx:
@fzyzcjy, the left CI failures are the same --- difference  between `typedef uintptr_t = ffi.UnsignedLong;` and `typedef uintptr_t = ffi.UnsignedLongLong;`  My pr should not change on this, is this introduced by bump test?  How to deal with it?

fzyzcjy:
That is because when you run codegen on different platforms that uintptr is different. I have indeed hacked it: https://github.com/fzyzcjy/flutter_rust_bridge/blob/f156ea6174c4dce1dbe5ea5a4f1d82eed780156b/.github/workflows/ci.yaml#L569

So, the solution (or, hack...) is: After running codegen on your computer, before committing you code, please *manually* (or write a script) modify that file and replace it with `UnsignedLong`

dbsxdbsx:
@fzyzcjy, I added an example for testing Apis consisting of multiple blocks  in folder `pure_dart_multi` and modify the `ci.yaml` accordingly. Could you please manually check them? 
In addition, I still wonder how to build a "should panic" example---Is there an example for `ci.yaml` ?

fzyzcjy:
> In addition, I still wonder how to build a "should panic" example---Is there an example for ci.yaml ?

Not yet. But I guess you can assert non-zero exit code in shell

frb_example/pure_dart_multi/README.md
```diff
@@ -0,0 +1,8 @@
+## Pure-Dart example for `flutter_rust_bridge (multi blocks of rust Api)`
```
fzyzcjy, line 1:
```suggestion
## Pure-Dart example for `flutter_rust_bridge` with multi blocks of rust Api
```

frb_example/pure_dart_multi/dart/lib/main.dart
```diff
@@ -0,0 +1,435 @@
...
+import 'package:test/test.dart';
+
+void main(List<String> args) async {
+  String dylibPath1 = args[0];
+  String dylibPath2 = args[1];
+  print(
+      'flutter_rust_bridge example program start (dylibPath=$dylibPath1,$dylibPath2)');
+  print('construct api');
+  final dylib1 = DynamicLibrary.open(dylibPath1);
+  final api1 = ApiClass1Impl(dylib1);
```
fzyzcjy, line 16: Btw are you using api2?

frb_example/pure_dart_multi/dart/lib/main.dart
```diff
@@ -0,0 +1,435 @@
...
+
+  test('dart call simpleAdder', () async {
```
fzyzcjy, line 18: No need to repeat *all* tests in `pure_dart`. Only run two tests here may be sufficient, one for api1, one for api2

frb_example/pure_dart_multi/dart/valgrind_util.py
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python3
```
fzyzcjy, line 1: Hey, do not copy and paste blindly :) Follow the DRY rule

Moreover, do we really need to run valgrind for pure-dart-multi? or is a simple `dart test` sufficient?

If the latter, remove run.sh and valgrind_util.py

frb_example/pure_dart_multi/rust/external/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[package]
+name = "flutter_rust_bridge_example_external"
```
fzyzcjy, line 2: Again, no need for this "external" subpackage I guess?

fzyzcjy:
Good job! Except for a few points that we can discuss

frb_example/pure_dart_multi/README.md
dbsxdbsx: Fixed in 3917234ba483bcd16ba702640a8d4b806d27d912.

frb_example/pure_dart_multi/dart/lib/main.dart
dbsxdbsx: I used, but at the very end of this function. Fixed in 3917234ba483bcd16ba702640a8d4b806d27d912.

frb_example/pure_dart_multi/dart/lib/main.dart
dbsxdbsx: Fixed in 3917234ba483bcd16ba702640a8d4b806d27d912.

frb_example/pure_dart_multi/rust/external/Cargo.toml
dbsxdbsx: Fixed in https://github.com/fzyzcjy/flutter_rust_bridge/commit/3917234ba483bcd16ba702640a8d4b806d27d912.

fzyzcjy:
Great! Btw your ci is broken https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/2566477865

dbsxdbsx:
@fzyzcjy, now almost everything is ready. If no other issue, I think it time to merge. I would refine doc correspondingly in another pr --- I need link to example modified here, so this pr should be published before writing doc.

fzyzcjy:
Sure. Just fix the CI otherwise we will have a broken CI on master which is terrible

frb_example/pure_dart_multi/dart/lib/main.dart
dbsxdbsx: > Sure. Just fix the CI otherwise we will have a broken CI on master which is terrible

I don't know why it failed? What does `checkout@v2` refer to?

fzyzcjy:
Use a yaml verifier to check it.

frb_example/pure_dart_multi/dart/lib/main.dart
dbsxdbsx: > Use a yaml verifier to check it.

Now everything is ok.

frb_codegen/src/utils.rs
```diff
@@ -12,15 +12,69 @@ pub fn mod_from_rust_path(code_path: &str, crate_path: &str) -> String {
         .replace('/', "::")
 }
 
-pub fn with_changed_file<F: FnOnce() -> anyhow::Result<()>>(
+pub fn with_changed_file(
     path: &str,
     append_content: &str,
-    f: F,
+    generate_c_from_rust_func: impl FnOnce() -> anyhow::Result<()>,
```
fzyzcjy, line 18: Why change `with_changed_file` and let it be aware of what is happening inside the callback? 

"with_changed_file" was designed to be a very general utility function, which changes some file content (while do not know what is that content) and run callback and change back content.

frb_codegen/src/utils.rs
```diff
@@ -12,15 +12,69 @@ pub fn mod_from_rust_path(code_path: &str, crate_path: &str) -> String {
...
     fs::write(&path, content_original.clone() + append_content)?;
 
-    f()?;
+    generate_c_from_rust_func()?;
+    generate_dart_from_c_func()?;
 
     Ok(fs::write(&path, content_original)?)
 }
+
+pub fn find_all_duplicates(mut vec: Vec<String>) -> anyhow::Result<()> {
```
fzyzcjy, line 30: I guess no need for so long code. As for duplicates, what about utilizing `Set`s? https://stackoverflow.com/questions/46766560/how-to-check-if-there-are-duplicates-in-a-slice

Sth like:

```
let mut uniq = HashSet::new();
let mut duplicates = vec![];
iter.into_iter().all(move |x| {
 let already_exists = uniq.insert(x);
 if already_exists { duplicates.insert(x) }
})
if duplicates.isnotempty() { panic (...) }

```

Btw, try to make this function generic as it is so general

frb_codegen/src/utils.rs
```diff
@@ -12,15 +12,69 @@ pub fn mod_from_rust_path(code_path: &str, crate_path: &str) -> String {
...
+            "{} [{}] {} already been defined",
+            symbol_str, conflict_symbols, verb_str
+        );
+    }
+
+    Ok(())
+}
+
+/// check api defined by users, if no duplicates, then generate all symbols (api function name)
+/// ,including those generated implicitily by frb
```
fzyzcjy, line 65: "," maybe at end of line, not begin of line

frb_codegen/src/utils.rs
```diff
@@ -12,15 +12,69 @@ pub fn mod_from_rust_path(code_path: &str, crate_path: &str) -> String {
...
+    let mut explicit_raw_symbols = Vec::new();
+    let mut all_symbols = Vec::new();
+    for (i, config) in configs.iter().enumerate() {
+        let raw_ir_file = config.get_ir_file();
+
+        // for checking explicit api duplication
+        explicit_raw_symbols.extend(raw_ir_file.funcs.iter().map(|f| f.name.clone()));
+
+        // for avoiding redundant generation in dart
+        all_symbols.extend(raw_ir_file.get_all_symbols(config, i + 1));
```
fzyzcjy, line 76: We can see that, if we have `BlockIndex` type, this code can be cleaner

frb_example/pure_dart/rust/build.rs
```diff
@@ -18,9 +20,11 @@ fn main() {
...
-    let opts = config_parse(raw_opts);
-    // run flutter_rust_bridge_codegen
-    for opt in &opts {
-        frb_codegen(opt).unwrap();
+    let configs = config_parse(raw_opts);
+
+    // generation of rust api for ffi
+    let all_symbols = get_symbols_if_no_duplicates(&configs).unwrap();
+    for (i, config) in configs.iter().enumerate() {
+        frb_codegen(config, &all_symbols, i + 1).unwrap();
```
fzyzcjy, line 28: Can we remove the block_index argument here, and integrate it into the config and auto gen it in config_parse? Otherwise it is quite cumbersome to let end users to specify such index manually.

fzyzcjy (commented):
Good job! Except for a few improvements - guess that you still have some improvement spaces when programming rust :)

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -39,6 +39,7 @@ pub fn generate(
     dart_api_impl_class_name: &str,
     dart_wire_class_name: &str,
     dart_output_file_root: &str,
+    block_cnt: usize,
```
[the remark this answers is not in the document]
dbsxdbsx: No, it is parsed with 1 and 2 in context of 2 blocks.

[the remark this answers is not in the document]
fzyzcjy: I see. So it is still an "index" not a "count" I guess

frb_codegen/src/utils.rs
dbsxdbsx: I have to say I didn't want to change it before 7ac242970f7992280b739532dc759c91c596bae2, but the CI lint just told me there are too many arguments after I introduce `exclude_symbols` (which I have to), I didn't find a better way, except for separating the `f:F`.

fzyzcjy: For https://github.com/fzyzcjy/flutter_rust_bridge/commit/7ac242970f7992280b739532dc759c91c596bae2#diff-92a01e35fcd86edac10d9dfa11ed17ec30ed1fdf565b37934105fb66e4f4f0a5L28, just create a struct as argument.

```
struct BindgenRustToDartArg{

    rust_crate_dir: &str,
    c_output_path: &str,
    dart_output_path: &str,
    dart_class_name: &str,
    c_struct_names: Vec<String>,
    exclude_symbols: Vec<String>,
    llvm_install_path: &[String],
    llvm_compiler_opts: &str,
}

pub fn bindgen_rust_to_dart(arg: BindgenRustToDartArg) -> anyhow::Result<()> {

File: .github/workflows/ci.yaml [hunk 12]
```diff
@@ -40,11 +40,15 @@ jobs:
       - name: Install ffigen
         run: dart pub global activate ffigen ^5.0.0 # TODO migrate to 6.x later
 
-      - name: Install dart dependencies
+      - name: Install dart dependencies (single block)
         working-directory: ./frb_example/pure_dart/dart
         run: dart pub get
 
-      - name: Run valgrind script
+      - name: Install dart dependencies (multi blocks)
```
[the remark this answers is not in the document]
dbsxdbsx: > So, in `valgrind_test`, the multi-blocks example has its dependency installed, but never executed? :(
> 
> In other words, we should have somewhere executing the multi-file package's tests.

Before, you said there is no need to do valgrind_test for the multi version. So there is only ` - name: Run valgrind script (single block)`. And the corresponding `run.sh` and `valgrind_util.py` are removed from `./frb_example/pure_dart_multi`.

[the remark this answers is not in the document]
fzyzcjy: Sure, no need to *valgrind* test the multi version; but need a simple `dart test` for the pure_dart_multi. Otherwise how can we ensure the code is right :)

frb_codegen/src/generator/rust/mod.rs
```diff
@@ -58,18 +64,16 @@ pub fn generate(
 
 struct Generator {
     extern_func_collector: ExternFuncCollector,
```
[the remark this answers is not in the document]
dbsxdbsx: fixed in c0d183c6054fde63bd5e0857232ebe417c9f8f63.

frb_codegen/src/commands.rs
```diff
@@ -96,16 +77,16 @@ fn execute_command(bin: &str, args: &[&str], current_dir: Option<&str>) -> Outpu
     result
 }
 
-fn cbindgen(
+pub fn cbindgen(
```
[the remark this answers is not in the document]
dbsxdbsx: fixed in f63fe9cadd9803c1dbda9a1420b37ae56bf25211. But after all, there is no need to make it pub 366ab46d614750fecca0c35f5e95220619f55355.

frb_codegen/src/utils.rs
dbsxdbsx: fixed

frb_codegen/src/generator/rust/mod.rs
```diff
@@ -162,7 +166,7 @@ impl Generator {
         lines.push(self.section_header_comment("executor"));
         lines.push(self.generate_executor(ir_file));
 
-        if self.options.should_generate_sync_execution_mode_utility {
+        if block_cnt == 1 {
```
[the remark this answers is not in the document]
dbsxdbsx: Fixed in 366ab46d614750fecca0c35f5e95220619f55355.

frb_codegen/src/generator/dart/mod.rs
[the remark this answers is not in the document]
dbsxdbsx: > I see. So it is still an "index" not a "count" I guess

Fixed in 366ab46d614750fecca0c35f5e95220619f55355.

frb_codegen/src/utils.rs
dbsxdbsx: > For [7ac2429#diff-92a01e35fcd86edac10d9dfa11ed17ec30ed1fdf565b37934105fb66e4f4f0a5L28](https://github.com/fzyzcjy/flutter_rust_bridge/commit/7ac242970f7992280b739532dc759c91c596bae2#diff-92a01e35fcd86edac10d9dfa11ed17ec30ed1fdf565b37934105fb66e4f4f0a5L28), just create a struct as argument.
> 
> ```
> struct BindgenRustToDartArg{
> 
>     rust_crate_dir: &str,
>     c_output_path: &str,
>     dart_output_path: &str,
>     dart_class_name: &str,
>     c_struct_names: Vec<String>,
>     exclude_symbols: Vec<String>,
>     llvm_install_path: &[String],
>     llvm_compiler_opts: &str,
> }
> 
> pub fn bindgen_rust_to_dart(arg: BindgenRustToDartArg) -> anyhow::Result<()> {
> ```

Fixed in d4575dc056dad1167a62ed7e81f06703130370df.

frb_codegen/src/utils.rs
dbsxdbsx: fixed and make it generic in a96cb32930e83bfddb9db35f896f058f8a7d2e36.

frb_example/pure_dart/rust/build.rs
dbsxdbsx: Fixed in d4575dc056dad1167a62ed7e81f06703130370df.

frb_codegen/src/utils.rs
```diff
@@ -24,3 +26,50 @@ pub fn with_changed_file<F: FnOnce() -> anyhow::Result<()>>(
...
+        // for avoiding redundant generation in dart
+        all_symbols.extend(raw_ir_file.get_all_symbols(config));
+    }
+    let duplicates = find_all_duplicates(&explicit_raw_symbols);
+    if !duplicates.is_empty() {
+        let mut duplicated_symbols = duplicates
+            .iter()
+            .map(|s| s.clone() + ",")
+            .collect::<String>();
+        duplicated_symbols.pop();
```
fzyzcjy, line 61: map + pop => join

dbsxdbsx: > map + pop => join

Fixed in f338868c3b285d4dad1e22d2a17a4c5d57e3f6e6.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
dbsxdbsx: > Sure, no need to _valgrind_ test the multi version; but need a simple `dart test` for the pure_dart_multi. Otherwise how can we ensure the code is right :)

I've added a simple test in run.sh within 0ef57e52004841b204416f54f100f42624f18955.

dbsxdbsx:
@fzyzcjy, if no other issue, I think it time to merge.

fzyzcjy:
Good job! I will review it a bit later (since this is a nontrivial PR) and if no problem surely it will be merged :)

.github/workflows/ci.yaml
```diff
@@ -40,20 +40,30 @@ jobs:
...
+
+      - name: Run bash and valgrind script (single block)
         working-directory: ./frb_example/pure_dart/dart
         run: chmod +x ./run.sh ./valgrind_util.py && ./run.sh
         env:
           CARGO_TARGET_DIR: /home/runner # NOTE here, must be an absolute path, since in my script uses it in multiple different directories
 
+      - name: Run bash script (multi blocks)
+        working-directory: ./frb_example/pure_dart_multi/dart
+        run: chmod +x ./run.sh
```
fzyzcjy, line 59: Hey this is wrong! You only `chmod`, i.e. add permission to file, but never execute it.

What about `dart test` or something like that.

.github/workflows/ci.yaml
```diff
@@ -40,20 +40,30 @@ jobs:
...
+      - name: Install dart dependencies (single block)
         working-directory: ./frb_example/pure_dart/dart
         run: dart pub get
 
-      - name: Run valgrind script
+      - name: Install dart dependencies (multi blocks)
+        working-directory: ./frb_example/pure_dart_multi/dart
+        run: dart pub get
+
+      - name: Run bash and valgrind script (single block)
```
fzyzcjy, line 51: No need to say `run bash` - seems you misunderstand what is `chmod`

frb_codegen/src/config.rs
```diff
@@ -78,9 +79,7 @@ pub struct Opts {
     pub manifest_path: String,
     pub dart_root: Option<String>,
     pub build_runner: bool,
-
-    // due to conflict of dart api of multiple blocks, keep it temporarily
-    pub exclude_sync_execution_mode_utility: bool,
+    pub block_index: usize, //NOTE: the index is counted from 1, not 0
```
fzyzcjy, line 82: What about creating a `BlockIndex` newtype as I have suggested.

fzyzcjy, line 82: In addition, maybe we should count starting from zero (not one), which is more intuitive to programmers?

frb_codegen/src/config.rs
```diff
@@ -222,7 +215,7 @@ pub fn parse(raw: RawOpts) -> Vec<Opts> {
                 manifest_path: manifest_paths[i].clone(),
                 dart_root: dart_roots[i].clone(),
                 build_runner, //same for all rust api blocks
-                exclude_sync_execution_mode_utility: exclude_sync_execution_mode_utilities[i],
+                block_index: i + 1,
```
fzyzcjy, line 218: If count from zero +1 is no longer needed

frb_codegen/src/generator/rust/mod.rs
```diff
@@ -162,7 +155,7 @@ impl Generator {
         lines.push(self.section_header_comment("executor"));
         lines.push(self.generate_executor(ir_file));
 
-        if self.options.should_generate_sync_execution_mode_utility {
+        if block_index == 1 {
```
fzyzcjy, line 158: Not changed yet?

.github/workflows/ci.yaml
dbsxdbsx: > Hey this is wrong! You only `chmod`, i.e. add permission to file, but never execute it.

Sorry for missing command, I've fixed in cc549d25ec2bffe128a3139952f1290c0be20cf7.

> What about `dart test` or something like that.
There is already some testing related to `pure_dart_multi `:
```yaml
      - name: Install dependencies for pure_dart_multi example
        working-directory: ./frb_example/pure_dart_multi/dart
        run: dart pub get
      - name: Verify formatting for pure_dart_multi example
        working-directory: ./frb_example/pure_dart_multi/dart
        run: dart format --output=none --set-exit-if-changed --line-length 120 .
      - name: Analyze project source for pure_dart_multi example
        working-directory: ./frb_example/pure_dart_multi/dart
        run: dart analyze --fatal-infos
```
You can see more tests by keyword `pure_dart_multi`, still not enough?

frb_codegen/src/generator/rust/mod.rs
dbsxdbsx: > Not changed yet?

What exactly you want here?  Making it `if block_index == 0` ?   Do you also mean that when outputting more than 1 same name Api, the suffix should started from 0? like `new_uint_8_list_0`?

fzyzcjy:
```
#[derive(Copy)]
pub struct BlockIndex(usize);

impl BlockIndex {
  const PRIMARY = BlockIndex(0);
}

// when want to pass block_index:
fn some_function(block_index: BlockIndex) {}

// when want to say it is the primary one
// replace "if block_index == 1 {" with:
if block_index == BlockIndex::PRIMARY {

File: .github/workflows/ci.yaml (hunk 17 above)
Comment (fzyzcjy) [reply .github/workflows/ci.yaml]: `format` and `analyze` are **NOT TESTS**.

File: frb_codegen/src/config.rs [hunk 22]
```diff
@@ -79,7 +80,7 @@ pub struct Opts {
     pub manifest_path: String,
     pub dart_root: Option<String>,
     pub build_runner: bool,
-    pub block_index: usize, //NOTE: the index is counted from 1, not 0
+    pub block_index: BlockIndex, //NOTE: the index is counted from 1, not 0
```
fzyzcjy, line 83: remove the outdated comments: "//NOTE: the index is counted from 1, not 0"

frb_codegen/src/utils.rs
```diff
@@ -69,3 +70,16 @@ pub fn get_symbols_if_no_duplicates(configs: &[crate::Opts]) -> Result<Vec<Strin
...
+#[derive(PartialEq, Eq, Debug, Clone, Copy)]
+pub struct BlockIndex(pub usize);
+
+impl BlockIndex {
+    pub const PRIMARY: BlockIndex = BlockIndex(0);
+}
+
+impl Display for BlockIndex {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0 + 1)
```
fzyzcjy, line 83: Well, shall we really "+1"? That will cause a lot of confusion IMHO

frb_codegen/src/config.rs
dbsxdbsx: > remove the outdated comments: "//NOTE: the index is counted from 1, not 0"

Fixed in https://github.com/fzyzcjy/flutter_rust_bridge/pull/527/commits/cc7cfbc4f1a0b944bff5d25383e8f87edc1a914f.

frb_codegen/src/generator/rust/mod.rs
dbsxdbsx: > ```
> #[derive(Copy)]
> pub struct BlockIndex(usize);
> 
> impl BlockIndex {
>   const PRIMARY = BlockIndex(0);
> }
> 
> // when want to pass block_index:
> fn some_function(block_index: BlockIndex) {}
> 
> // when want to say it is the primary one
> // replace "if block_index == 1 {" with:
> if block_index == BlockIndex::PRIMARY {
> ```

Fixed in [cc7cfbc](https://github.com/fzyzcjy/flutter_rust_bridge/pull/527/commits/cc7cfbc4f1a0b944bff5d25383e8f87edc1a914f).

frb_codegen/src/config.rs
dbsxdbsx: Fixed in [cc7cfbc](https://github.com/fzyzcjy/flutter_rust_bridge/pull/527/commits/cc7cfbc4f1a0b944bff5d25383e8f87edc1a914f).

frb_codegen/src/config.rs
dbsxdbsx: Now the final code is always counting from zero, even for the Api suffix shown for user. [119e29b](https://github.com/fzyzcjy/flutter_rust_bridge/pull/527/commits/119e29bd5203c2cf7fd92dfc103bf13184c66c26)

dbsxdbsx: > What about creating a `BlockIndex` newtype as I have suggested.

Fixed in [cc7cfbc](https://github.com/fzyzcjy/flutter_rust_bridge/pull/527/commits/cc7cfbc4f1a0b944bff5d25383e8f87edc1a914f).

.github/workflows/ci.yaml
dbsxdbsx: > `format` and `analyze` are **NOT TESTS**.

I've no idea what other tests should be done here. Currently, there is a `run.sh`:
```sh
#!/bin/bash

set -eux

if [[ -z "${CARGO_TARGET_DIR}" ]]; then
  echo 'Please set environment variable CARGO_TARGET_DIR'
  exit 1
fi

(cd ../rust && cargo build --verbose)

# dart pub get

# need to be AOT, since prod environment is AOT, and JIT+valgrind will have strange problems
dart compile exe lib/main.dart -o main.exe
```

Is there something else need to add?

fzyzcjy: That is nothing but *compile* things, but not *test* things.

Maybe firstly learn a bit about how people test things in dart/flutter, and then you will know it

dbsxdbsx: > That is nothing but _compile_ things, but not _test_ things.
> 
> Maybe firstly learn a bit about how people test things in dart/flutter, and then you will know it

There is already  some test code here `pure_dart_multi\dart\lib\main.dart`:
```dart
 test('dart call simpleAdder', () async {
    expect(await api1.simpleAdder1(a: 42, b: 100), 142);
  });

  final dylib2 = DynamicLibrary.open(dylibPath2);
  final api2 = ApiClass2Impl(dylib2);
  test('dart call simpleAdder2', () async {
    expect(await api2.simpleAdder2(a: 42, b: 100), 142);
  });
```
Do you mean there is an essential to explicitly add a test command, like `dart test ...`?

fzyzcjy: Yes. run command like `dart test` will execute it.

dbsxdbsx: > Yes. run command like `dart test` will execute it.

May I ask, why there is no such test command for the original example `pure_dart\dart`?

fzyzcjy: Because it compiled `.dart` into an executable binary, and use Valgrind to run that binary (with a small python wrapper script), because we want Valgrind to detect possible memory problems.

frb_example/pure_dart_multi/dart/run.sh
```diff
@@ -2,15 +2,15 @@
...
 set -eux
 
-if [[ -z "${CARGO_TARGET_DIR}" ]]; then
-  echo 'Please set environment variable CARGO_TARGET_DIR'
-  exit 1
-fi
+# if [[ -z "${CARGO_TARGET_DIR}" ]]; then
+#   echo 'Please set environment variable CARGO_TARGET_DIR'
+#   exit 1
+# fi
```
fzyzcjy, line 8: Temporary hack to check CI is ok, just remember not to submit to the final PR :)

dbsxdbsx: > Temporary hack to check CI is ok, just remember not to submit to the final PR :)

Actually, I am trying to figure out what make [this ](https://github.com/fzyzcjy/flutter_rust_bridge/runs/7174628203?check_suite_focus=true)issue;  

I still think there should be `api_1.rs` and `api_2.rs` in `pure_dart_multi\rust\src`, but not `api.rs` and `api_2.rs`.

fzyzcjy: > I still think there should be api_1.rs and api_2.rs in pure_dart_multi\rust\src, but not api.rs and api_2.rs.

Sure, you can put those two names.

Btw it will be even greater if it is named with meaning, e.g. `api_apple`, `api_orange`, etc. Otherwise people may wrongly think only numbers are supported.

fzyzcjy: >   thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', /home/runner/work/flutter_rust_bridge/flutter_rust_bridge/frb_codegen/src/config.rs:379:77

What about print out some error message during this unwrap. For example, replace `.unwrap()` with `.expect("the file is: {}", your_file_name)`. Then you know which file is it thinking.

For example, it maybe because, when assembling filename, you add an extra `/` etc

dbsxdbsx: > > I still think there should be api_1.rs and api_2.rs in pure_dart_multi\rust\src, but not api.rs and api_2.rs.
> 
> Sure, you can put those two names.
> 
> Btw it will be even greater if it is named with meaning, e.g. `api_apple`, `api_orange`, etc. Otherwise people may wrongly think only numbers are supported.

I would give detailed explanation in doc afterwards.

dbsxdbsx: > > thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', /home/runner/work/flutter_rust_bridge/flutter_rust_bridge/frb_codegen/src/config.rs:379:77
> 
> What about print out some error message during this unwrap. For example, replace `.unwrap()` with `.expect("the file is: {}", your_file_name)`. Then you know which file is it thinking.
> 
> For example, it maybe because, when assembling filename, you add an extra `/` etc

The [issue](https://github.com/fzyzcjy/flutter_rust_bridge/runs/7179679911?check_suite_focus=true) comes from that it still using `api.rs` instead of `api_1.rs` or `api_2.rs` as the path for `cargo:rerun-if-changed=...`. Is it related to something like cache, which makes it still using the old `cargo:rerun-if-changed=...` data when running `Run valgrind script (multi blocks)`?

fzyzcjy: Maybe search what is `cargo:rerun-if-changed=` and detailed documentation will be seen

dbsxdbsx: > Maybe search what is `cargo:rerun-if-changed=` and detailed documentation will be seen

[It](https://github.com/fzyzcjy/flutter_rust_bridge/runs/7181080218?check_suite_focus=true) still found the wrong path, I guess there is something wrong with the CI statement.
```yaml
  - name: Run valgrind script (multi blocks)
    working-directory: ./frb_example/pure_dart_multi/dart
    run: chmod +x ./run_multi.sh && ./run_multi.sh
    env:
      CARGO_TARGET_DIR: /home/runner # NOTE here, must be an absolute path, since in my script uses it in multiple different directories
```
How do you think?

fzyzcjy: What is the correct path?

dbsxdbsx: > What is the correct path?

Let's take a look at this [issue](https://github.com/fzyzcjy/flutter_rust_bridge/runs/7188985333?check_suite_focus=true),
```
  --- stderr
  thread 'main' panicked at 'panic with file: /home/runner/work/flutter_rust_bridge/flutter_rust_bridge/frb_example/pure_dart_multi/rust/src/api.rs', 
```
There is no `api.rs`, only `api_1.rs` and `api_2.rs`. And for testing, I even removed statement like `println!("cargo:rerun-if-changed={}", file_name);`
from `/pure_dart_multi/rust/build.rs`.

So I wonder what makes this issue occur.

fzyzcjy: Maybe dig into the rust code which generates such wrong path.

fzyzcjy: Btw you can also debug locally by running those commands on your computer

dbsxdbsx: > Btw you can also debug locally by running those commands on your computer

Now I am pretty sure it is caused by the shared `flutter_rust_bridge_example`. Look at [here](https://github.com/fzyzcjy/flutter_rust_bridge/runs/7189850597?check_suite_focus=true):  I switch the order of single and multi blocks, and as expected, the latter one is ruined, while the former one is ok.  And in this case, they both used the same `flutter_rust_bridge_example-3b9163c44a994e76`.

But... how to fix it?

fzyzcjy: Then, instead of one `valgrind script` in CI, lets create two:

* valgrind script (the old one)
* pure_dart_multi

Then, the two runs in parallel without any interference

dbsxdbsx: > Then, instead of one `valgrind script` in CI, lets create two "job"s:
> 
> * valgrind script (the old one)
> * pure_dart_multi
> 
> Then, the two runs in parallel without any interference
> 
> Btw we already have a ton of jobs. For example, valgrind test, run codegen, ...


I find a better way. The `[lib]` from `Cargo.toml` used by single and multi blocks examples are now separated by different names `flutter_rust_bridge_example` and `flutter_rust_bridge_example_multi`. Now there is [no valgrind conflict issue](https://github.com/fzyzcjy/flutter_rust_bridge/runs/7193628108?check_suite_focus=true).   

By the way, to make distinguishable,https://github.com/fzyzcjy/flutter_rust_bridge/pull/527/commits/c6ed19d4d2501d472e0891fce274b9662d5e38be, I also renamed the package names with `flutter_rust_bridge_example_single_block_test` and `flutter_rust_bridge_example_multi_blocks_test` respectively, is that ok?

fzyzcjy: Good idea! LGTM

fzyzcjy:
Great job!

PR #542, opened 2022-07-01
adds support for stream sink into any argument

lattice0:
StreamSink currently has to be the first argument. This PR makes it work in any place. 

Examples:

```rust
use flutter_rust_bridge::StreamSink;

pub fn tick(a: u32, b: u32, sink: StreamSink<i32>) -> Result<(), anyhow::Error> {
    todo!()
}

pub fn tick_2(a: u32, sink: StreamSink<i32>, b: u32) -> Result<(), anyhow::Error> {
    todo!()
}

pub fn tick_3(sink: StreamSink<i32>, a: u32, b: u32) -> Result<(), anyhow::Error> {
    todo!()
}

pub fn tick_4(sink: StreamSink<i32>) -> Result<(), anyhow::Error> {
    todo!()
}
```

PS: this is a draft. I'll erase the test files and the mess once the PR is approved

Closes #541

PR #543, opened 2022-07-01
Supporting methods

### LINKED ISSUE #67: Feature Request: Convert `impl`s to methods
So I'm not even sure if this can be implemented, but in the example, there is a struct:

```
#[derive(Debug, Clone)]
pub struct Point {
    pub x: f64,
    pub y: f64,
}
```

Say we had a corresponding `impl`

```
impl Point {
    pub fn norm(&self) -> f64 {
        (self.x * self.x + self.y * self.y).sqrt()
    }
}
```

In the binding, this `impl` could be converted to

```
class Point {
  final double x;
  final double y;

  Point({
    required this.x,
    required this.y,
  });

  Future<float> norm() {
    [[link to norm implementation]]
  }
}
```

I don't know how feasible this is, especially with using rust methods in ffi, but you can actually just turn them into regular functions like this:

```
pub fn point_norm(p: Point) -> f64 {
    p.norm()
}
```

**Additional context**
This is just a thought though, and I'm happy to be proven wrong :)

### LINKED ISSUE #539: Supporting methods
Just to know, isn't there support for methods?

```
pub struct MyStruct{
    something: String
}

impl MyStruct{
    fn do_something() {
        todo!()
    }
}

pub fn return_my_struct() -> MyStruct {
    todo!()
}
```

does not translate any methods.

Should I create my own pseudomethods by doing

```
pub fn my_struct_do_something(my_struct: &MyStruct) {
    //...
}
```

?

Also, I have to do

```
pub fn return_my_struct() -> MyStruct {
    todo!()
}
```

otherwise it won't create MyStruct on dart side. Is it supposed to be like this?

lattice0:
ULTRA draft, just a request for how to go in this. Also I don't promise it will be done, I just want to see how hard would it be. 

@fzyzcjy please look at [frb_codegen/src/parser/mod.rs](https://github.com/fzyzcjy/flutter_rust_bridge/compare/master...lattice0:flutter_rust_bridge:methods?expand=1#diff-69ebf7751e9ff8017c40b4e7257ba1289be95f8913c3552e5aaa3bf59bd7cfd1)

here I parse methods. What do you think of transforming a method to a function which receives a reference of the Self type?

For example

```rust
pub struct StructWithMethod{
    something: String
}

impl StructWithMethod {
    pub fn do_something(&self, _u: u32) {
        todo!()
    }
}
```

would turn into

```rust
pub fn struct_with_method_do_something(self_object: &StructWithMethod, _u: u32) {
//...
}
```

if done like this, then I'd do something like this

```rust
if let Item::Impl(ref item_impl) = item {
            for item in &item_impl.items {
                if let ImplItem::Method(item_method) = item {
                    if let Visibility::Public(_) = &item_method.vis {
                        let item_fn = modify_method_to_function(item_method);
                        src_fns.push(item_fn);
                    }
                }
            }
        }
```

where `modify_method_to_function` would just create a function out of the method.

Close #539 
Close #67


fzyzcjy:
> What do you think of transforming a method to a function which receives a reference of the Self type?

Sure. By doing that, we can reuse most of the existing code.

lattice0:
Is it okay to change

`fn extract_fns_from_file(file: &File) -> Vec<&ItemFn> {`

to

`fn extract_fns_from_file(file: &File) -> Vec<ItemFn> {`

Cause we're going to have to create some `ItemFn` inside this function

lattice0:
Please don't mind approving workflows for now. Sorry, gonna publish my sketches on another branch

lattice0:
The only way I found to transform a method implementation into a function in a way that we know later that it should be a method, is by marking the function name as ending with `is_a_method`: https://github.com/fzyzcjy/flutter_rust_bridge/blob/e7cbec7dd980ab638d337a71cdd58d6532f604bc/frb_codegen/src/parser/mod.rs#L236

What do you think of this approach? I prefer to ask first then proceed with the code. Currently it generated the rust part correctly. The other option would be to reengineer a lot of stuff, and the goal is to reuse the code, transforming a method into a function, and later transforming this to a method call.

This:
```rust
pub struct StructWithMethod{
    pub something: String
}

impl StructWithMethod {
    pub fn do_something(&self, _u: u32, _x: String) {
        todo!()
    }
}
```

turned into


```rust
#[no_mangle]
pub extern "C" fn wire_StructWithMethod_do_something_is_a_method(
    port_: i64,
    StructWithMethod_is_a_method: *mut wire_StructWithMethod,
    _u: u32,
    _x: *mut wire_uint_8_list,
) {
    FLUTTER_RUST_BRIDGE_HANDLER.wrap(
        WrapInfo {
            debug_name: "StructWithMethod_do_something_is_a_method",
            port: Some(port_),
            mode: FfiCallMode::Normal,
        },
        move || {
            let api_StructWithMethod_is_a_method = StructWithMethod_is_a_method.wire2api();
            let api__u = _u.wire2api();
            let api__x = _x.wire2api();
            move |task_callback| {
                Ok(StructWithMethod::do_something(
                    &api_StructWithMethod_is_a_method,
                    api__u,
                    api__x,
                ))
            }
        },
    )
}
```

of course there's the dart part as well and mutability or not, etc, and testing for different types, args, etc

fzyzcjy:
> Is it okay to change
fn extract_fns_from_file(file: &File) -> Vec<&ItemFn> {
to
fn extract_fns_from_file(file: &File) -> Vec<ItemFn> {

Sure. If you care about performance, maybe `Cow<ItemFn>` is better. But since codegen is not very slow currently you may just use ItemFn

fzyzcjy:
> 
Please don't mind approving workflows for now. Sorry, gonna publish my sketches on another branch

#542 is merged and now I do not need to approve :)

fzyzcjy:
> is by marking the function name as ending with is_a_method:

Generally looks ok to me.

Nit: What about ending with `__method`. Because (1) it is shorter (2) use double underscore to avoid conflict with normal functions, just like how Python and c++ uses double underscore

fzyzcjy:
> of course there's the dart part as well and mutability or not, etc, and testing for different types, args, etc

Yes, we should test for these details - great to see you have good testing in mind

fzyzcjy:
Btw, the main high-level idea is: Just transform a method into a function (and vise versa), and let the original flutter_rust_bridge to the heavy lifting.

So yes, should not have to rewrite too much code otherwise it is the wrong direction :)

lattice0:
The problem I had was that I transformed the `&self` into `struct_with_method: &StructWithMethod`, but then still there's no way to know that this is a method, so marking the name with `struct_with_method__method` solves this.

There's still the problem of not knowing if it's a mutable self or not. I could do `__method_mutable` versus `__method` for non mutable, or create a https://github.com/lattice0/flutter_rust_bridge/blob/e7cbec7dd980ab638d337a71cdd58d6532f604bc/frb_codegen/src/ir/ty.rs#L14 but `StructMut`. What do you think?

fzyzcjy:
Wait a minute - can it really be mutable? In other words, if the data is mutated, will we really get data mutated as well on the Dart side? I guess not currently.

lattice0:
You're right. I was thinking about opaque types but even for them it might be unsafe to do so from various isolates, so I should support only non mutable?

fzyzcjy:
> I should support only non mutable?

Currently, I guess yes.

lattice0:
I just remembered that https://github.com/mozilla/uniffi-rs only supports &self also, so yes this is the correct way. When you want to modify something you have to use interior mutability probably with a mutex

lattice0:
There's no way to generate methods for a struct without a `Vec<IrFunc>` inside `TypeDartGenerator` at https://github.com/lattice0/flutter_rust_bridge/blob/89e5ac0f91664c642cf935f3feb4c4968479f2d8/frb_codegen/src/generator/dart/mod.rs#L110

can I make an option to set this vec like this

```rust
pub trait TypeDartGeneratorTrait {
    fn api2wire_body(&self) -> Option<String>;

    fn api_fill_to_wire_body(&self) -> Option<String> {
        None
    }

    fn wire2api_body(&self) -> String {
        "".to_string()
    }

    fn structs(&self) -> String {
        "".to_string()
    }

    fn set_vec_ir_func(&self, ir_func: Vec<IrFunc>)  {
       todo!()
    }
}
```

or maybe


```rust
pub trait TypeDartGeneratorTrait {
    fn api2wire_body(&self) -> Option<String>;

    fn api_fill_to_wire_body(&self) -> Option<String> {
        None
    }

    fn wire2api_body(&self) -> String {
        "".to_string()
    }

    fn structs(&self, ir_funcs: Vec<IrFunc>) -> String { //possibly even ir_funcs: &[IrFunc]
        "".to_string()
    }
}
```

fzyzcjy:
The `set_vec_ir_func` sounds not like a good idea because it introduces mutability.

`fn structs(&self, ir_funcs: Vec<IrFunc>) -> String { //possibly even ir_funcs: &[IrFunc]` seems ok

But, maybe ir_funcs are *already* in ir_file? Thus, since https://github.com/lattice0/flutter_rust_bridge/blob/89e5ac0f91664c642cf935f3feb4c4968479f2d8/frb_codegen/src/generator/dart/mod.rs#L110 provides ir_file, we can already get ir_funcs from them

lattice0:
```rust
pub struct StructWithMethod {
    pub something: String,
}
pub fn return_struct() -> StructWithMethod {
    todo!()
}
```

turns roughly into


```dart
class StructWithMethod {
  final String something;

  StructWithMethod({
    required this.something,
  });



class SomethingImpl extends FlutterRustBridgeBase<SomethingWire>
    implements Something {
  factory SomethingImpl(ffi.DynamicLibrary dylib) =>
      SomethingImpl.raw(SomethingWire(dylib));

  SomethingImpl.raw(SomethingWire inner) : super(inner);

  Future<StructWithMethod> returnStruct({dynamic hint}) =>
      executeNormal(FlutterRustBridgeTask(
        callFfi: (port_) => inner.wire_return_struct(port_),
        parseSuccessData: _wire2api_struct_with_method,
        constMeta: kReturnStructConstMeta,
        argValues: [],
        hint: hint,
      ));
```

as you see, `class StructWithMethod` has no way of calling ffi stuff like `executeNormal`. What do you think of `class StructWithMethod` extending `FlutterRustBridgeBase<StructWithMethod>` when the struct has a method?

fzyzcjy:
> What do you think of class StructWithMethod extending FlutterRustBridgeBase<StructWithMethod> when the struct has a method?

What about this:

```dart
class StructWithMethod {
  final FlutterRustBridgeBase<SomethingWire> bridge;
  final String something;

  StructWithMethod({
    required this.bridge,
    required this.something,
  });

  Future<Something> someMethod(Something b) => bridge.someMethod(a: this, b: b);
}

**Comment by lattice0** [comment]
Ok. I guess `StructWithMethod` was never meant to be manually created in the dart side, right? It should be returned by `return_struct`

**Comment by fzyzcjy** [comment]
> I guess StructWithMethod was never meant to be manually created in the dart side, right? It should be returned by return_struct

Maybe we should allow it to be created directly in Dart side.

An extreme example is struct without fields, which acts like "namespace".

**Comment by lattice0** [comment]
Ok I got

```rust
pub struct StructWithMethod {
    pub something: String,
}

impl StructWithMethod {
    pub fn do_something(&self, _u: u32, _x: String) {
        todo!()
    }
}
```

to compile into

```dart

class StructWithMethod {
  final SomethingImpl bridge;
  final String something;

  StructWithMethod({
    required this.bridge,
    required this.something,
  });
  Future<void> doSomething({required int u, required String x, dynamic hint}) =>
      bridge.doSomethingMethod(
        structWithMethod: this,
        u: u,
        x: x,
      );
}


class SomethingImpl extends FlutterRustBridgeBase<SomethingWire>
    implements Something {
  factory SomethingImpl(ffi.DynamicLibrary dylib) =>
      SomethingImpl.raw(SomethingWire(dylib));

  SomethingImpl.raw(SomethingWire inner) : super(inner);

  Future<void> doSomethingMethod(
          {required StructWithMethod structWithMethod,
          required int u,
          required String x,
          dynamic hint}) =>
      executeNormal(FlutterRustBridgeTask(
        callFfi: (port_) => inner.wire_do_something__method(
            port_,
            _api2wire_box_autoadd_struct_with_method(structWithMethod),
            _api2wire_u32(u),
            _api2wire_String(x)),
        parseSuccessData: _wire2api_unit,
        constMeta: kDoSomethingMethodConstMeta,
        argValues: [structWithMethod, u, x],
        hint: hint,
      ));
      //...
```

it only fails here:
```dart
StructWithMethod _wire2api_struct_with_method(dynamic raw) {
  final arr = raw as List<dynamic>;
  if (arr.length != 1)
    throw Exception('unexpected arr length: expect 1 but see ${arr.length}');
  return StructWithMethod(
    /// because it's missing the bridge here
    something: _wire2api_String(arr[0]),
  );
}
```

now I don't know if I should pass the bridge. I'm confused now

fzyzcjy:
> now I don't know if I should pass the bridge. I'm confused now

pass *this* :)

lattice0:
but 

```dart
StructWithMethod _wire2api_struct_with_method(dynamic raw) {
  final arr = raw as List<dynamic>;
  if (arr.length != 1)
    throw Exception('unexpected arr length: expect 1 but see ${arr.length}');
  return StructWithMethod(
    something: _wire2api_String(arr[0]),
  );
}
```

is a function, there's no `this` for a bridge class. It's not inside anything.

fzyzcjy:
You are right. Then what about 

`_wire2api_struct_with_method(FlutterRustBridgeBase<SomethingWire> bridge, ...)`

and

modify `parseSuccessData` such that pass the bridge into it

fzyzcjy:
Btw I was wrong:

```
class StructWithMethod {
  final SomethingImpl bridge;
```

should be 

```
class StructWithMethod {
  final FlutterRustBridgeBase<SomethingWire> bridge;
```

b/c we should not hardcode implementation class there

lattice0:
I tried with

```dart
class StructWithMethod {
  final FlutterRustBridgeBase<SomethingWire> bridge;

```

but `FlutterRustBridgeBase` has no methods. 

The thing that actually has methods is the implementation/extension of `FlutterRustBridge`

```dart
class SomethingImpl extends FlutterRustBridgeBase<SomethingWire>
    implements Something {
  factory SomethingImpl(ffi.DynamicLibrary dylib) =>
      SomethingImpl.raw(SomethingWire(dylib));

  SomethingImpl.raw(SomethingWire inner) : super(inner);

  Future<void> doSomethingMethod(
          {required StructWithMethod structWithMethod,
          required int u,
          required String x,
          dynamic hint}) =>
      executeNormal(FlutterRustBridgeTask(
        callFfi: (port_) => inner.wire_do_something__method(
            port_,
            _api2wire_box_autoadd_struct_with_method(structWithMethod),
            _api2wire_u32(u),
            _api2wire_String(x)),
        parseSuccessData: _wire2api_unit,
        constMeta: kDoSomethingMethodConstMeta,
        argValues: [structWithMethod, u, x],
        hint: hint,
      ));
```

fzyzcjy:
You are right. So we should use `Something` not `SomethingImpl`, since the former is like an abstract interface

fzyzcjy:
Btw I am not sure why CI does not run (except for codacy). Maybe need to resolve conflict?

Anyway no worries about CI - only need to pass before merging

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -338,7 +338,7 @@ fn has_methods(struct_name: String, ir_file: &IrFile) -> bool {
 
 // Tests if the function in `f` is a method for struct with name `struct_name`
 fn is_method_for_struct(f: &&IrFunc, struct_name: String) -> bool {
-    f.name.contains("__method")
+    f.name.contains(METHOD_MARKER)
```
fzyzcjy, line 341: nit: maybe it is "endswith"?

lattice0: for `METHOD_MARKER=__method`, it can be, but for `STATIC_METHOD_MARKER=__static_method` don't because a static method is

`method_name__static_method___StructName`

I had to embed the struct for which the static method is. For common methods, it's always the first argument so no need

fzyzcjy: Agree. So `contains` LGTM

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -453,12 +453,20 @@ fn generate_api_func(func: &IrFunc, ir_file: &IrFile) -> GeneratedApiFunc {
             execute_func_name,
             func.wire_func_name(),
             wire_param_list.join(", "),
-            if func.name.contains(STATIC_METHOD_MARKER) {
+            if func.name.contains(STATIC_METHOD_MARKER)
+                && func.name.split("___").last().unwrap() == {
```
fzyzcjy, line 457: nit: `___` sounds like some magic constant

Shall we encapsulate all logic about function naming into one place? For example, this place will encode/decode a string (the final function name) from/to a struct (say, `struct DecodedFunctionName {real_name: String, is_static_method: bool, struct_name: str, ...}`)

lattice0: Ok, I think I can create like `is_static_method` and `split_static_method`

lattice0:
I removed all the trash I added to make it easier for you to review. Everything built on my side, both the rust and dart generated files, no errors. Now I still have to write the dart tests but they should pass

frb_codegen/src/generator/dart/mod.rs
fzyzcjy: Sure. The main idea is, a function name now indeed contains many different pieces of information, such as "real function name", "is it static", "struct name", etc. So it is indeed a classical serialization thing.

frb_codegen/src/utils.rs
```diff
@@ -83,3 +86,63 @@ impl Display for BlockIndex {
         write!(f, "{}", self.0)
     }
 }
+
+// Methods and static methods helper functions
+
+pub const STATIC_METHOD_MARKER: &str = "__static_method___";
```
fzyzcjy, line 92: nit: then it can become non-pub, since all methods are in this file

fzyzcjy:
> Do you think they should be static on the flutter side? If so, they have to receive bridge.

Not sure why I did not receive notification about this question...

Yes, I guess they should be static. Otherwise, suppose `ConcatenateWith` struct has 100 fields, then it means, in order to run the static method we have to provide a struct with 100 fields even if it is unneeded!

frb_codegen/src/parser/mod.rs
```diff
@@ -187,20 +187,163 @@ impl<'a> Parser<'a> {
...
+                    }
+                } else {
+                    None
+                }
+            };
+            Ident::new(
+                format!(
+                    "{}{}{}",
+                    item_method.sig.ident.clone(),
+                    STATIC_METHOD_MARKER,
```
lattice0, line 269: @fzyzcjy METHOD_MARKER and STATIC_METHOD_MARKER are still used here

fzyzcjy:
Oh I did not realized this is ready for review. I will review it soon.

Btw maybe need to resolve conflicts o/w ci will not run

frb_codegen/src/generator/rust/mod.rs
```diff
@@ -276,8 +286,31 @@ impl Generator {
...
+                "{}({})",
+                func.name,
+                inner_func_params.join(", ")
+            ))
+        } else {
+            let method_name = if is_method(&func.name) {
+                inner_func_params[0] = format!("&{}", inner_func_params[0]);
+                clear_method_marker(&func.name)
+            } else if is_static_method(&func.name) {
+                static_method_return_method_name(&func.name)
```
fzyzcjy, line 300: Can we have even stronger type system for naming as follows (just rough draft, surely you can change the struct name etc):

```rust
struct FunctionName {
  actual_name: String,
  method_info: enum {
    NotMethod,
    StaticMethod {struct_name: String},
    NonStaticMethod {struct_name: String},
  }
}

impl FunctionName{
  fn serialize(&self) -> String {} // assemble it into a string, e.g. `actual_name___thestructname___static` etc
}

  fn deserialze(s: String) -> Self {} // the reverse
```

frb_codegen/src/parser/mod.rs
```diff
@@ -187,20 +187,163 @@ impl<'a> Parser<'a> {
...
 }
 
+// Converts an item implementation (something like fn(&self, ...)) into a function where `&self` is a named parameter to `&Self`
+fn item_method_to_function(item_impl: &ItemImpl, item_method: &ImplItemMethod) -> Option<ItemFn> {
+    if let Type::Path(p) = item_impl.self_ty.as_ref() {
+        let struct_name = p.path.segments.first().unwrap().ident.to_string();
+        let span = item_method.sig.ident.span();
+        let is_static_method = {
+            let Signature {
+                constness: _,
```
fzyzcjy, line 222: maybe use `..`

frb_codegen/src/parser/mod.rs
```diff
@@ -187,20 +187,163 @@ impl<'a> Parser<'a> {
...
+                }
             }
         }
     }
 
     src_fns
 }
 
+// Converts an item implementation (something like fn(&self, ...)) into a function where `&self` is a named parameter to `&Self`
+fn item_method_to_function(item_impl: &ItemImpl, item_method: &ImplItemMethod) -> Option<ItemFn> {
```
fzyzcjy, line 216: Well, if I understatnd correctly, this approach is: convert a method into a (fake) function AST, then use `parse_function` to convert that fake function AST into our IR (IrFunc).

Then, what about mimic `parse_function` and create `parse_method` to directly convert ItemImpl to IrFunc? 

Surely, the possibly duplicated code between parse_function and parse_method should be extracted, if any

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -104,12 +110,20 @@ fn get_dart_api_spec_from_ir_file(ir_file: &IrFile, block_index: BlockIndex) ->
...
-        .map(generate_api_func)
+        .map(|f| generate_api_func(f, ir_file))
         .collect::<Vec<_>>();
     let dart_structs = distinct_types
         .iter()
-        .map(|ty| TypeDartGenerator::new(ty.clone(), ir_file).structs())
+        .map(|ty| {
+            {
+                let mut t = TypeDartGenerator::new(ty.clone(), ir_file);
+                t.set_dart_api_class_name(dart_api_class_name);
```
fzyzcjy, line 120: As discussed here https://github.com/fzyzcjy/flutter_rust_bridge/pull/543#issuecomment-1172990930, maybe we can avoid mutability?

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -394,14 +414,27 @@ fn generate_api_func(func: &IrFunc) -> GeneratedApiFunc {
...
+                    } else {
+                        "".to_string()
+                    }
+                })
+                || struct_has_methods(ir_file, func.inputs.get(0).as_ref().map(|x| &x.ty))
+            {
+                format!("(d) => _wire2api_{}(this, d)", func.output.safe_ident())
+            } else {
+                format!("_wire2api_{}", func.output.safe_ident())
+            },
```
fzyzcjy, line 437: nit: maybe extract to a function or at least not here inside an argument of `format!`

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -474,16 +507,26 @@ fn generate_api_fill_to_wire_func(ty: &IrType, ir_file: &IrFile) -> String {
     }
 }
 
-fn generate_wire2api_func(ty: &IrType, ir_file: &IrFile) -> String {
-    let body = TypeDartGenerator::new(ty.clone(), ir_file).wire2api_body();
+fn generate_wire2api_func(ty: &IrType, ir_file: &IrFile, dart_api_class_name: &str) -> String {
+    let bridge_requirement = if let StructRef(IrTypeStructRef { name, freezed: _ }) = ty {
```
fzyzcjy, line 511: uber nit: bridge_requirement -> extra_argument

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -474,16 +507,26 @@ fn generate_api_fill_to_wire_func(ty: &IrType, ir_file: &IrFile) -> String {
...
+        if has_methods(name, ir_file) {
+            format!("{} bridge,", dart_api_class_name)
+        } else {
+            "".to_string()
+        }
+    } else {
+        "".to_string()
+    };
```
fzyzcjy, line 519: nit: 

```
if matches!(ty, StructRef(IrTypeStructRef{name}) if has_methodds(name, irfile)) { format!(..) } else {""}

File: frb_codegen/src/generator/dart/mod.rs [hunk 12]
```diff
@@ -521,3 +564,47 @@ fn dart_metadata(metadata: &[IrDartAnnotation]) -> String {
...
+        s
+    } else {
+        return false;
+    };
+    if the_struct.safe_ident().contains("user_id") {
+        println!(
+            "struct_has_methods for file: {:?}, the_struct: {:?}",
+            file, the_struct
+        );
+    }
```
fzyzcjy: debug code?

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -521,3 +564,47 @@ fn dart_metadata(metadata: &[IrDartAnnotation]) -> String {
...
+    let struct_name = if let StructRef(IrTypeStructRef { name, freezed: _ }) = the_struct {
+        name
+    } else {
+        return false;
+    };
+
+    file.funcs.iter().any(|f| {
+        if let Some(IrField {
+            ty: _,
```
fzyzcjy, line 584: use the `..` operator

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -521,3 +564,47 @@ fn dart_metadata(metadata: &[IrDartAnnotation]) -> String {
...
+            name,
+            is_final: _,
+            comments: _,
+        }) = f.inputs.get(0)
+        {
+            if name.raw.to_case(Case::UpperCamel) == *struct_name {
+                println!(
```
fzyzcjy: btw maybe remove all debugging things

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -521,3 +564,47 @@ fn dart_metadata(metadata: &[IrDartAnnotation]) -> String {
...
+                    "returning true for struct_name: {} and name.raw: {}",
+                    struct_name,
+                    name.raw.to_case(Case::UpperCamel)
+                );
+                true
+            } else {
+                false
+            }
+        } else {
+            false
```
fzyzcjy, line 592: again, such `if-if-else-else`

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -521,3 +564,47 @@ fn dart_metadata(metadata: &[IrDartAnnotation]) -> String {
...
+
+    file.funcs.iter().any(|f| {
+        if let Some(IrField {
+            ty: _,
+            name,
+            is_final: _,
+            comments: _,
+        }) = f.inputs.get(0)
+        {
+            if name.raw.to_case(Case::UpperCamel) == *struct_name {
```
fzyzcjy: Well, do we really need to do such comparsion, especially with a re-casing? That sounds brittle...

frb_codegen/src/generator/dart/ty.rs
```diff
@@ -51,14 +54,54 @@ impl<'a> TypeDartGenerator<'a> {
...
         match ty {
-            Primitive(ir) => TypePrimitiveGenerator { ir, context }.into(),
-            Delegate(ir) => TypeDelegateGenerator { ir, context }.into(),
-            PrimitiveList(ir) => TypePrimitiveListGenerator { ir, context }.into(),
-            Optional(ir) => TypeOptionalGenerator { ir, context }.into(),
-            GeneralList(ir) => TypeGeneralListGenerator { ir, context }.into(),
-            StructRef(ir) => TypeStructRefGenerator { ir, context }.into(),
-            Boxed(ir) => TypeBoxedGenerator { ir, context }.into(),
-            EnumRef(ir) => TypeEnumRefGenerator { ir, context }.into(),
+            Primitive(ir) => TypePrimitiveGenerator {
```
fzyzcjy, line 57: about `dart_api_class_name` - as suggested before, try not to have such mutable fields, but maybe make it immutable

frb_codegen/src/generator/dart/ty_struct.rs
```diff
@@ -102,10 +155,13 @@ impl TypeDartGeneratorTrait for TypeStructRefGenerator<'_> {
                         f.name.dart_style()
                     )
                 })
-                .collect::<Vec<_>>()
-                .join("\n");
+                .collect::<Vec<_>>();
+            if has_methods {
```
fzyzcjy, line 159: Can we make it immutable

fzyzcjy (commented):
Great job! This is a nontrivial PR, so I made a bit more comments.

fzyzcjy:
Btw we need to add a bit of documentation later, probably be siblings of this chapter http://cjycode.com/flutter_rust_bridge/feature/zero_copy.html

frb_codegen/src/utils.rs
```diff
@@ -83,3 +86,60 @@ impl Display for BlockIndex {
...
+}
+
+pub fn has_methods(struct_name: &String, ir_file: &IrFile) -> bool {
+    ir_file.funcs.iter().any(|f| {
+        is_method_for_struct(&f, struct_name) || is_static_method_for_struct(&f, struct_name)
+    })
+}
+
+// Tests if the function in `f` is a method for struct with name `struct_name`
+pub fn is_method_for_struct(f: &&IrFunc, struct_name: &String) -> bool {
```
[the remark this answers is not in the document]
lattice0: because when used in an iterator, the iterator iterates over references so we get double reference

frb_codegen/src/generator/dart/mod.rs
lattice0: Does `matches!` allow for `if`? 
`
matches!(ty, StructRef(IrTypeStructRef{name, ..}) if has_methods(name, irfile) { format!("{} bridge,", dart_api_class_name) } else {"".to_string()});`

says

`no rules expected the token `{``

frb_example/pure_dart/dart/lib/bridge_generated.dart
```diff
@@ -199,6 +199,22 @@ abstract class FlutterRustBridgeExampleSingleBlockTest {
   Stream<Log> handleStreamSinkAt3({required int key, required int max, dynamic hint});
 
   FlutterRustBridgeTaskConstMeta get kHandleStreamSinkAt3ConstMeta;
+
+  Future<int> sumMethod({required SumWith sumWith, required int y, required int z, dynamic hint});
```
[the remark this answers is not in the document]
lattice0: It's hard to get the info of if `sumWith` is a 'self' or a simple argument, I think it would increase complexity only

frb_codegen/src/generator/dart/mod.rs
lattice0: We need to find the function for which the first argument is the struct. I don't see other way

frb_codegen/src/generator/dart/ty.rs
lattice0: I'll have to pass it in the constructor then. is it ok? Cause it will change all the enum variants above

frb_codegen/src/generator/dart/ty_struct.rs
lattice0: Do you have a better idea to conditionally do `inner.insert(0, "bridge: bridge,".to_string())` if `if has_methods`?

```
        let mut inner = s
            .fields
            .iter()
            .enumerate()
            .map(|(idx, field)| {
                format!(
                    "{}: _wire2api_{}(arr[{}]),",
                    field.name.dart_style(),
                    field.ty.safe_ident(),
                    idx
                )
            })
            .collect::<Vec<_>>();
        if has_methods {
            inner.insert(0, "bridge: bridge,".to_string());
        }
```

fzyzcjy:
Ping me when you have finished everything (e.g. code change, or comments), and I will look at all comments and reply all

frb_codegen/src/utils.rs
```diff
@@ -83,3 +86,60 @@ impl Display for BlockIndex {
...
+        }) = &f.inputs[0].ty
+        {
+            if let StructRef(IrTypeStructRef { name, freezed: _ }) = &**inner {
+                *name == *struct_name
+            } else {
+                false
+            }
+        } else {
+            false
+        }
```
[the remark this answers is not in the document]
lattice0: I was able to reduce one else but not the other cause the pattern is not irrefutable

frb_codegen/src/generator/rust/mod.rs
lattice0: I just tried this and ended up in the situation where sometimes I don't parse a string but I parse a IrFunc:

```rust
fn is_method_return_struct_name(func: &IrFunc) -> Option<String> {
        if Self::is_non_static_method(&func.name) {
            let input = func.inputs[0].clone();
            Some(input.name.to_string().to_case(Case::UpperCamel))
        } else if Self::is_static_method(&func.name) {
            Some(Self::static_method_return_struct_name(&func.name).to_case(Case::UpperCamel))
        } else {
            None
        }
    }
```

cause I need the input. Can we leave like this?

@fzyzcjy need your input before doing the rest, I'm almost in the end

frb_example/pure_dart/dart/lib/bridge_generated.dart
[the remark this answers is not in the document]
lattice0: ok found a way

lattice0:
@fzyzcjy done!

frb_codegen/src/method_utils.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    // Tests if the function is `f` is a static method for struct with name `struct_name`
+    pub fn is_static_method_for_struct(f: &&IrFunc, struct_name: &str) -> bool {
+        f.name.contains(STATIC_METHOD_MARKER)
+            && f.name.split(STATIC_METHOD_MARKER).last().unwrap() == struct_name
+    }
+
+    //Does `ir_file` has any methods directed for `struct_name`?
+    pub fn has_methods(struct_name: &String, ir_file: &IrFile) -> bool {
```
fzyzcjy, line 29: uber-nit: &String -> &str?

frb_codegen/src/method_utils.rs
```diff
@@ -0,0 +1,181 @@
+use crate::ir::IrType::{self, StructRef};
+use crate::ir::{IrField, IrTypeStructRef};
+use crate::ir::{IrFile, IrFunc};
+use convert_case::{Case, Casing};
+
+const STATIC_METHOD_MARKER: &str = "__static_method___";
```
fzyzcjy, line 6: nit: maybe a line of comment discussing why it is 2 underline before and 3 underline after it. readers may intuitively think it is both 2 underlines before and after.

or better, make them both 2 (or both 3?) underlines

frb_codegen/src/method_utils.rs
```diff
@@ -0,0 +1,181 @@
...
+    // Tests if the function is `f` is a static method for struct with name `struct_name`
+    pub fn is_static_method_for_struct(f: &&IrFunc, struct_name: &str) -> bool {
+        f.name.contains(STATIC_METHOD_MARKER)
+            && f.name.split(STATIC_METHOD_MARKER).last().unwrap() == struct_name
+    }
+
+    //Does `ir_file` has any methods directed for `struct_name`?
+    pub fn has_methods(struct_name: &String, ir_file: &IrFile) -> bool {
+        ir_file.funcs.iter().any(|f| {
+            Self::is_method_for_struct(&f, struct_name)
```
fzyzcjy, line 31: About the `&&IrFunc` problem - can we directly pass `f` not `&f` here? (And all similar places). Since I guess `&IrFunc` is `Copy`?

frb_codegen/src/generator/dart/mod.rs
fzyzcjy: `if matches!(ty, something if somethingelse) { sth } else {sth}`

https://doc.rust-lang.org/std/macro.matches.html

frb_codegen/src/generator/dart/mod.rs
fzyzcjy: All right. Then maybe we can add a comment about this

frb_codegen/src/generator/dart/ty.rs
fzyzcjy: Maybe we can do this:

```rust
let context = TypeGeneratorContext { ir_file };
```

becomes 

```rust
let context = TypeGeneratorContext { ir_file, dart_api_class_name };
```

then no need to modify that many things and is easier to read

frb_codegen/src/generator/dart/ty_struct.rs
fzyzcjy: I am thinking about `concat` but anyway that is also quite verbose in rust world...

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -104,12 +108,16 @@ fn get_dart_api_spec_from_ir_file(ir_file: &IrFile, block_index: BlockIndex) ->
...
         .funcs
         .iter()
-        .map(generate_api_func)
+        .map(|f| generate_api_func(f, ir_file))
         .collect::<Vec<_>>();
     let dart_structs = distinct_types
         .iter()
-        .map(|ty| TypeDartGenerator::new(ty.clone(), ir_file).structs())
+        .map(|ty| {
+            { TypeDartGenerator::new(ty.clone(), ir_file, Some(dart_api_class_name.to_string())) }
```
fzyzcjy, line 116: nit: why `{}`

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -631,3 +631,91 @@ pub fn handle_stream_sink_at_3(
...
+                    key,
+                    value: i.to_string(),
+                });
+            }
+            sink.close();
+        });
+        Ok(())
+    }
+
+    pub fn handle_some_static_stream_sink_at_1(sink: StreamSink<u32>) -> Result<(), anyhow::Error> {
```
fzyzcjy, line 712: nit: that "at_1" originally meant "streamsink argument is not at first argument" and was implemented in another pr. But it is the only arg here... Not sure what we should do.

frb_example/pure_dart/dart/lib/bridge_generated.dart
```diff
@@ -264,6 +298,47 @@ class Attribute {
...
 
+class ConcatenateWith {
+  final FlutterRustBridgeExampleSingleBlockTest bridge;
+  final String a;
+
+  ConcatenateWith({
+    required this.bridge,
+    required this.a,
+  });
+  static Future<ConcatenateWith> newConcatenateWith(
```
fzyzcjy, line 309: uber-nit: shall we add newlines between methods

frb_codegen/src/parser/mod.rs
```diff
@@ -187,20 +187,150 @@ impl<'a> Parser<'a> {
...
+                .serialize(),
+                span,
+            )
+        };
+
+        Some(ItemFn {
+            attrs: vec![],
+            vis: item_method.vis.clone(),
+            sig: Signature {
+                constness: None,
```
fzyzcjy, line 279: nit: Does `..Default::default()` trick help

frb_codegen/src/generator/dart/ty_struct.rs
```diff
@@ -56,13 +68,49 @@ impl TypeDartGeneratorTrait for TypeStructRefGenerator<'_> {
         )
     }
 
+    fn set_dart_api_class_name(&mut self, s: &str) {
```
fzyzcjy, line 71: nit: maybe no longer needed?

frb_codegen/src/method_utils.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    pub fn is_non_static_method(&self) -> bool {
+        matches!(
+            self,
+            FunctionName {
+                actual_name: _,
+                method_info: MethodInfo::NonStatic { .. }
+            }
+        )
+    }
```
fzyzcjy, line 180: uber-nit: `matches!(self.method_info, MethodInfo::NonStatic {..})`

frb_codegen/src/method_utils.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    pub fn is_static_method(&self) -> bool {
+        matches!(
+            self,
+            FunctionName {
+                actual_name: _,
+                method_info: MethodInfo::Static { .. }
+            }
+        )
```
fzyzcjy, line 169: same as the comment just now

frb_codegen/src/method_utils.rs
```diff
@@ -0,0 +1,181 @@
...
+            }
+            MethodInfo::NonStatic { struct_name } => {
+                MethodNamingUtil::mark_as_non_static_method(&self.actual_name, struct_name)
+            }
+        }
+    }
+
+    pub fn deserialize(s: &str) -> Self {
+        let actual_name: String;
+        let method_info = {
```
fzyzcjy, line 128: what about:

```
if ... {
  FunctionName {actualName: ..., method_info: ...}
} else {
  FunctionName {...}
}
```

In other words, no need to temporary variable actual_name and method_info

And, use early return seems good

```
if staticmethod { return ... }
if nonstatic { return ...}
return ... 
```

frb_codegen/src/method_utils.rs
```diff
@@ -0,0 +1,181 @@
...
+        s.contains(METHOD_MARKER)
+    }
+
+    // Is the function name for a static method?
+    pub fn is_static_method(s: &str) -> bool {
+        s.contains(STATIC_METHOD_MARKER)
+    }
+
+    // Tests if the function is `f` is a static method for struct with name `struct_name`
+    pub fn is_static_method_for_struct(f: &&IrFunc, struct_name: &str) -> bool {
```
fzyzcjy, line 23: same "non-pub" comments applies to most functions in this file.

e.g. whoever uses is_static_method_for_struct can change to deserialize + check struct_name equality

frb_codegen/src/method_utils.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    // Returns the name of the struct that this method is for
+    pub fn non_static_method_return_struct_name(s: &str) -> String {
+        s.split(METHOD_MARKER).last().unwrap().to_string()
+    }
+
+    // Returns the name of method itself
+    pub fn non_static_method_return_method_name(s: &str) -> String {
+        s.split(METHOD_MARKER).next().unwrap().to_string()
+    }
```
fzyzcjy, line 59: same applies here

frb_codegen/src/method_utils.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    // Tests if a given struct has methods, that is, if the `ir_file` contains
+    // a function that receives the struct as first argument
+    pub fn struct_has_methods(file: &IrFile, the_struct: Option<&IrType>) -> bool {
```
fzyzcjy, line 63: nit: shall we change type of `the_struct` into `&IrType`. Because a none definitely never has methods

Or even change to `&IrTypeStructRef`?

frb_codegen/src/method_utils.rs
```diff
@@ -0,0 +1,181 @@
...
+        let the_struct = if let Some(s) = the_struct {
+            s
+        } else {
+            return false;
+        };
+        let struct_name = if let StructRef(IrTypeStructRef { name, freezed: _ }) = the_struct {
```
fzyzcjy, line 69: uber-nit: `..`

fzyzcjy (commented):
Great job! The code is ready to be merged, with only some minor modifications.

Again, this is not a tiny PR, so there definitely has to exist more comments :)

fzyzcjy:
Again ping me when done! :)

frb_codegen/src/generator/dart/mod.rs
lattice0: I don't know, the formatter adds them automatically

frb_codegen/src/method_utils.rs
lattice0: I don't understand. Should I leave them as non pub? Then how to call from the `FunctionName` struct?

lattice0:
@fzyzcjy done. On some I couldn't comment, but now `MethodNamingUtil` is not used directly anymore, everything goes through `FunctionName`, except for struct_has_method.

frb_codegen/src/parser/mod.rs
fzyzcjy: @lattice0 seems this comment is folded so needs to be manually expanded to see it

fzyzcjy:
I am reviewing it.

Btw some comments are folded by github and need to click load more button. e.g. item_method_to_function-problem https://github.com/fzyzcjy/flutter_rust_bridge/pull/543#discussion_r919507081

frb_codegen/src/generator/dart/ty.rs
```diff
@@ -13,6 +13,8 @@ pub trait TypeDartGeneratorTrait {
         "".to_string()
     }
 
+    fn set_dart_api_class_name(&mut self, _s: &str) {}
```
fzyzcjy, line 16: Well this whole method may be completely *deleted*?

frb_codegen/src/generator/dart/ty_struct.rs
```diff
@@ -56,13 +68,49 @@ impl TypeDartGeneratorTrait for TypeStructRefGenerator<'_> {
...
         let src = self.ir.get(self.context.ir_file);
         let comments = dart_comments(&src.comments);
         let metadata = dart_metadata(&src.dart_metadata);
 
+        let ir_file = self.context.ir_file;
+        let methods = ir_file
+            .funcs
+            .iter()
+            .filter(|f| {
+                MethodNamingUtil::is_method_for_struct(f, &src.name)
```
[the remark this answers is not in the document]
fzyzcjy: Seems not resolved yet?

frb_codegen/src/method_utils.rs
fzyzcjy: @lattice0 Maybe FunctionName is within the same file, so non-pub functions can be happily called?

```rust
use std::mem;

struct A;

impl A {
    fn f() {}
}

struct B;

impl B {
    fn g() {
        A::f(); // NOTE it calls non-public
    }
}

fn main() {
}
```

this toy code compiles.

frb_codegen/src/method_utils.rs
```diff
@@ -124,33 +119,36 @@ impl FunctionName {
...
-            actual_name,
-            method_info,
         }
     }
 
     pub fn method_name(&self) -> String {
         self.actual_name.clone()
     }
 
+    pub fn static_method_name(&self) -> String {
```
fzyzcjy, line 148: nit: maybe this is `Option<String>`, and should `match &self.method_info { MethodInfo::Static=> return that thing; else => return none`

fzyzcjy:
Good job! Now there is almost no new review issues (only some old ones that may have been overlooked because of github's notification mechanism) and it is very close to merging.

fzyzcjy:
Ping me when things are done :)

frb_codegen/src/parser/mod.rs
lattice0: I don't understand what is the problem here. I convert to `ItemFn` because this is what the other function did. I just do it but for ItemImpl. Then later it is converted just like the other function.

fzyzcjy: Well, IMHO it is as follows:

Current approach:

```
syn::ImplItemMethod ---frb:: item_method_to_function---> syn::ItemFn
syn::ItemFn ----frb::parse_function---> frb::IrFunc
```

Proposed approach:

```
syn::ImplItemMethod ---frb::parse_method---> frb::IrFunc
syn::ItemFn ----frb::parse_function---> frb::IrFunc
```

The current approach creates some "artificial" temporary syn::ItemFn which does not seem to be elegant, since we indeed want IrFunc.

However, if it is very hard to convert syn::ImplItemMethod into frb::IrFunc directly (probably with some refactoring on parse_function), then the current approach seems ok.

lattice0: It would require a re-do of lots of things, right?

fzyzcjy: I guess maybe not. But anyway the current approach seems acceptable.

fzyzcjy:
Great job!

frb_codegen/src/generator/dart/ty.rs
```diff
@@ -48,17 +49,57 @@ pub enum TypeDartGenerator<'a> {
...
-            Primitive(ir) => TypePrimitiveGenerator { ir, context }.into(),
-            Delegate(ir) => TypeDelegateGenerator { ir, context }.into(),
-            PrimitiveList(ir) => TypePrimitiveListGenerator { ir, context }.into(),
-            Optional(ir) => TypeOptionalGenerator { ir, context }.into(),
-            GeneralList(ir) => TypeGeneralListGenerator { ir, context }.into(),
-            StructRef(ir) => TypeStructRefGenerator { ir, context }.into(),
-            Boxed(ir) => TypeBoxedGenerator { ir, context }.into(),
-            EnumRef(ir) => TypeEnumRefGenerator { ir, context }.into(),
+            Primitive(ir) => TypePrimitiveGenerator {
+                ir,
```
fzyzcjy, line 56: I realized a weird thing... This PR changes something like 

```rust
PrimitiveList(ir) => TypePrimitiveListGenerator { ir, context }.into(),
```

into

```rust
            PrimitiveList(ir) => TypePrimitiveListGenerator {
                ir,
                context,
                dart_api_class_name,
            }
            .into(),
```

I guess it is because some problem when merging from upstream? Because in master branch it is in the short form.

lattice0: Isn't it because it gets too long and it breaks into lines?

fzyzcjy: Well, it may look neater to put dart_api_class_name as a field of `TypeGeneratorContext` I guess

fzyzcjy: I see. So maybe not because merging from upstream problem, but because this PR really changes it. So maybe we can put dart_api_class_name as a field of TypeGeneratorContext

Issue #560, opened 2022-07-09
No function named `new_with_null_ptr`


TENX-S:
frb version: 1.37.1
rustc: 1.62.0 stable
macOS 12.4

<img width="1080" alt="Screen Shot 2022-07-09 at 08 57 15" src="https://user-images.githubusercontent.com/40336192/178085834-9e855dd0-12d5-4a7d-b764-f05cc959762f.png">

fzyzcjy:
could you please provide a minimal reproducible sample, especially what is the api.rs?

TENX-S:
api.rs:
```rust
#[derive(Debug, Serialize, Deserialize)]
pub enum Scope {
    Project,
    Setting,
    Trash,
    Request,
}

#[derive(Debug, Serialize, Deserialize)]
pub struct Command {
    pub scope: Box<Scope>,
    pub body: String,
}

pub fn execute(command: Command) -> ZeroCopyBuffer<Vec<u8>> {
    println!("{:#?}", command);
    ZeroCopyBuffer(vec![1,2,3])
}
```

fzyzcjy:
Interesting. Could you please make a PR and add failing tests into the pure_dart example? Notice there are already examples there about such enums, so maybe just modify the existing ones as little as possible.

TENX-S:
I'll give it a try

TENX-S:
when running codegen:
```
flutter_rust_bridge_codegen --rust-input rust/src/api.rs --dart-output dart/lib/bridge_generated.dart --dart-format-line-length 120
```

I got this error:
```shell
thread 'main' panicked at 'parse_type failed for ty=N', /Users/tenx/.cargo/registry/src/rsproxy.cn-8f6827c7555bfaf8/flutter_rust_bridge_codegen-1.37.1/src/parser/ty.rs:142:32
stack backtrace:
   0:        0x10a497384 - std::backtrace_rs::backtrace::libunwind::trace::hee82a97921985ffc
                               at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:        0x10a497384 - std::backtrace_rs::backtrace::trace_unsynchronized::hd2f3f197f68d4196
                               at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
[...]

PR #561, opened 2022-07-09
reproduce #560

TENX-S:


fzyzcjy:
Good job!

fzyzcjy:
Seems like an enum problem. 

/cc @Desdaemon who implemented this great feature. What do you think?

frb_example/pure_dart/rust/src/bridge_generated.rs
```diff
@@ -1085,6 +1112,11 @@ pub extern "C" fn new_box_u8_0(value: u8) -> *mut u8 {
     support::new_leak_box_ptr(value)
 }
 
+#[no_mangle]
+pub extern "C" fn new_box_weekdays_0() -> *mut i32 {
+    support::new_leak_box_ptr(i32::new_with_null_ptr())
```
Desdaemon, line 1117: This should be `*mut i32`

Desdaemon (commented):
I probably missed a pointer check in the gen-code for Box, I'll take a look at this soon.

PR #562, opened 2022-07-09
Pool and cancellable

lattice0:
merge new stuff + `pub struct CancelTokenHandle(pub i64, pub i64)` compiles.

I'm unsure of what `nativeVisionUtils` should be. Perhaps you forgot to put the code for it? Also there's no `CancelTokenHandle`


lattice0:
dammm I messed up the commits

PR #565, opened 2022-07-09
Precompiled binary CI workflow

### LINKED ISSUE #436: Publish precompiled binary.
I suggest , you to publish precompiled binary in [release section](https://github.com/fzyzcjy/flutter_rust_bridge/releases/).

I am using windows (x64)

Desdaemon:
Closes #436. Modeled after the CI process of Starship.rs.

- [x] Setup CI to compile binaries on each release
  - [ ] ~Figure out notarizing MacOS binaries (using [gon](https://github.com/mitchellh/gon)?)~ Todo
- [x] Setup basic binary distribution
  - [x] Linux/Universal: `cargo-binstall`
  - [x] Windows: `scoop`
  - [x] MacOS: Homebrew

Action required:
- [ ] @fzyzcjy Create a new secret GH_PAT for a `public-repo`-only token


fzyzcjy:
Btw you may be interested in adding a `workflow_dispatch` (https://github.com/fzyzcjy/flutter_rust_bridge/blob/d1097eae26f1bed4231945062cdad142b28f645e/.github/workflows/post_release.yaml#L8), and then manually trigger it

contrib/flutter_rust_bridge_codegen.json
```diff
@@ -0,0 +1,34 @@
+{
+    "version": "1.37.1",
```
fzyzcjy, line 2: Btw is this auto-generated by script or manually written? If manually, we may need justfile's `release` recipe to bump all these versions

Desdaemon: I made it so that `just release` will automatically update this file, and Scoop users can use the auto-update definitions so they don't have to re-download this manifest.

Desdaemon: In the future, this file should live in the main Scoop bucket repository and managed automatically.

fzyzcjy: LTGM. Btw nit: shall we put it in `contrib` folder, or `frb_codegen/contrib`? Since if in the future we have another binary to distribute it will be a problem

Desdaemon: I don't really mind where they stay, so your call.

fzyzcjy: Then maybe we can put it into `frb_codegen/contrib`

frb_codegen/contrib/flutter_rust_bridge_codegen.json
```diff
@@ -0,0 +1,36 @@
...
+    "homepage": "https://fzyzcjy.github.io/flutter_rust_bridge/",
+    "license": "MIT",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/fzyzcjy/flutter_rust_bridge/releases/download/v1.37.2/flutter_rust_bridge_codegen-x86_64-pc-windows-msvc-v1.37.2.zip",
+            "hash": "Not Found"
+        },
+        "32bit": {
+            "url": "https://github.com/fzyzcjy/flutter_rust_bridge/releases/download/v1.37.2/flutter_rust_bridge_codegen-i686-pc-windows-msvc-v1.37.2.zip",
+            "hash": "Not Found"
```
Desdaemon, line 13: To be updated after merging.

fzyzcjy:
Good job! So what should I do?

fzyzcjy:
Well, I am considering the naming things. Currently the users will need to type `daedaemon/...` when wanting to install this package (https://github.com/fzyzcjy/flutter_rust_bridge/pull/565/files#diff-0fa4b7c861c1d83db3bde952e8f31666e6e8c3f31d50cb14592511e40fdb1524), while they see the package is under `fzyzcjy/...`. This seems uncommon according to my experiences when using packages, so it may cause some confusion and make downloaders wonder whether they are downloading the wrong package, or that `fzyzcjy` may discontinue in maintaining the package.

Surely your contribution is already mentioned in README (along with all other contributions) :)

Issue #585, opened 2022-07-20
Issue returning struct with implementations


Zaitam:
While using the new feature added in the PR https://github.com/fzyzcjy/flutter_rust_bridge/pull/543, I found an issue while returning an struct containg an implementation.
Ex code:
```rs 
pub fn get_sum_struct() -> SumWith {
    SumWith {x: 21}
}

pub struct SumWith { pub x: u32 }

impl SumWith {
    pub fn sum(&self, y: u32) -> u32 { self.x + y }
}
```
> Highlight on the first function, `get_sum_struct`

If we now run the flutter_rust_bridge, no errors will be found. Anyhow, there is an issue in the `bridge_generated.dart` file.

Dart bridge generated code:
```dart
Future<SumWith> getSumStruct({dynamic hint}) => executeNormal(FlutterRustBridgeTask(
        callFfi: (port_) => inner.wire_get_sum(port_),
        parseSuccessData: _wire2api_sum_with,
        constMeta: kGetSumConstMeta,
        argValues: [],
        hint: hint,
      ));
```

The problem is with the `_wire2api_sum_with` var. It's type is of `SumWith Function(Native, dynamic)` instead of the expected `SumWith Function(dynamic)`.

I think the issue is quite simple to fix, and can be done by adding a nullable parameter, even thought I don't think they exist in rust. I haven't got the needed exprience in Rust to take hands in the matter and make a PR.

fzyzcjy:
Hmm sounds reasonable.

/cc @lattice0 who implemented this wonderful feature.

lattice0:
Fix: https://github.com/fzyzcjy/flutter_rust_bridge/pull/587

Zaitam:
So... 
To start with, it now works for singular items. Thanks @lattice0 for that.

However it doesn't work for arrays nor vectors.
```rs
pub fn get_sum_struct() -> [SumWith; 1] {
    [SumWith {x: 21}]
}
```
```rs
pub fn get_sum_struct() -> Vec<SumWith> {
    vec![SumWith {x: 21}]
}
```

```dart
List<SumWith> _wire2api_list_get_sum_struct(dynamic raw) {
  return (raw as List<dynamic>).map(_wire2api_get_sum_struct).toList();
}
```
With similar issue where `_wire2api_get_sum_struct` is of type `SumWith Function(Native, dynamic)` instead of `dynamic Function(dynamic)`

[...]

PR #587, opened 2022-07-20
Fix method return

lattice0:
fixes https://github.com/fzyzcjy/flutter_rust_bridge/issues/585, that is, when we return a struct with a method in a non method function, it should work

## Checklist

- [x] New tests are added to ensure new features are working (probably by modifying the `frb_example/pure_dart` example).
- [x] All existing and new tests are passing.
- [x] If this PR adds/changes features, documentations (in `book` folder) are modified as well.


fzyzcjy:
Good job! I guess need to run code formatter to make CI happy

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -403,7 +408,11 @@ fn generate_api_func(func: &IrFunc, ir_file: &IrFile) -> GeneratedApiFunc {
                 "".to_string()
             }
         })
+        // If struct has a method with first element `input0`
         || (input_0.is_some() && MethodNamingUtil::struct_has_methods(ir_file, input_0.unwrap()))
+        //If output is a struct with methods
+        || (func_output_struct_name.is_some()
+            && MethodNamingUtil::has_methods(&func_output_struct_name.unwrap(), ir_file))
```
fzyzcjy: nit: Shall we merge `has_methods` and `struct_has_methods` functions? They look a bit similar. (method_utils.rs)

lattice0: They use different techniques. One does a look on the name and also searches using the method name, the other searches for the first argument receiving a self to the struct type

fzyzcjy: I see. That is partially why I think it may be a bit confusing...

lattice0: removed!

fzyzcjy:
Now wait for CI and merge when it goes green :)

lattice0:
Can you rerun? this error is strange

fzyzcjy:
sure

fzyzcjy:
No worries about that, that did occur before (though very weird)

Issue #406, opened 2022-04-20
[Bug]  Variant Enum in stream


LambdaMan2K:
### Describe the bug

Hi :) 
now I try to adding a  rust variant enum like in the example in the guide and I get this while using the generator 

when doing 
```
pub enum KitchenSink {
    Empty,
    Primitives {
        /// Dart field comment
        int32: i32,
        float64: f64,
        boolean: bool,
    },

    Buffer(ZeroCopyBuffer<Vec<u8>>),
}
pub struct Msg {
    pub id: String,
    pub test: KitchenSink,
}

pub fn message_handler(sink: StreamSink<Msg>) -> anyhow::Result<()>{Ok(())}
```

if this is expected please close this :) 

### Codegen logs with `RUST_LOG=debug` env variable

```shell
[2022-04-20T15:47:54Z WARN  flutter_rust_bridge_codegen::commands] command="powershell" "-noprofile" "-c" "cd \"C:\\Users\\me\\projects\\rust\\project\"; dart run build_runner build --delete-conflicting-outputs" stdout= stderr=Could not find a file named "pubspec.yaml" in "C:\Users\me\AppData\Local\Pub\Cache\hosted\pub.dartlang.org\flutter_rust_bridge-1.25.0".
    #0      new Pubspec.load (package:pub/src/pubspec.dart:288:7)
    #1      new Package.load (package:pub/src/package.dart:149:27)
    #2      SystemCache.load (package:pub/src/system_cache.dart:113:20)
    #3      Entrypoint._createPackageGraph (package:pub/src/entrypoint.dart:112:63)
    #4      Entrypoint.packageGraph (package:pub/src/entrypoint.dart:107:54)
    #5      getExecutableForCommand (package:pub/src/executable.dart:338:19)
    #6      RunCommand.run (package:dartdev/src/commands/run.dart:232:32)
    #7      CommandRunner.runCommand (package:args/command_runner.dart:209:27)
    #8      DartdevRunner.runCommand (package:dartdev/dartdev.dart:232:30)
    #9      CommandRunner.run.<anonymous closure> (package:args/command_runner.dart:119:25)
    #10     new Future.sync (dart:async/future.dart:296:31)
    #11     CommandRunner.run (package:args/command_runner.dart:119:14)
    #12     runDartdev (package:dartdev/dartdev.dart:67:29)
[...]

PR #599, opened 2022-07-28
Run build_runner either for Dart or Flutter

Roms1383:
This PR follows the issue #406.
I'm gonna need your recommendation @fzyzcjy on how to correctly differentiate a Dart from a Flutter project, hence the draft.

## Checklist

- [x] New tests are added to ensure new features are working (probably by modifying the `frb_example/pure_dart` example).
- [x] All existing and new tests are passing.
- [ ] If this PR adds/changes features, documentations (in `book` folder) are modified as well.

### DIFF
--- a/frb_codegen/src/commands.rs
+++ b/frb_codegen/src/commands.rs
@@ -258,18 +258,20 @@ pub fn format_dart(path: &str, line_length: i32) -> Result {
 pub fn build_runner(dart_root: &str) -> Result {
     info!("Running build_runner at {}", dart_root);
     let context = guess_context(&dart_root)?;
-    let cmd = if context == Context::Dart { "dart run build_runner build" } else { "flutter pub run build_runner build" };
+    let cmd = if context == Context::Dart {
+        "dart run build_runner build"
+    } else {
+        "flutter pub run build_runner build"
+    };
     let out = if cfg!(windows) {
         call_shell(&format!(
             "cd \"{}\"; {} --delete-conflicting-outputs",
-            dart_root,
-            cmd
+            dart_root, cmd
         ))
     } else {
         call_shell(&format!(
             "cd \"{}\" && {} --delete-conflicting-outputs",
-            dart_root,
-            cmd
+            dart_root, cmd
         ))
     };
     if !out.status.success() {
@@ -284,13 +286,18 @@ pub fn build_runner(dart_root: &str) -> Result {
 
 pub fn guess_context(dart_root: &str) -> std::result::Result<Context, Error> {
     info!("Guessing context the runner is run into");
-    let pubspec = PathBuf::from(dart_root)
-        .join("pubspec.lock");
-    if !pubspec.exists() { return Err(Error::StringError(format!("missing pubspec.lock in {}", dart_root))); }
+    let pubspec = PathBuf::from(dart_root).join("pubspec.lock");
+    if !pubspec.exists() {
+        return Err(Error::StringError(format!(
+            "missing pubspec.lock in {}",
+            dart_root
+        )));
+    }
     let pubspec = std::fs::read_to_string(pubspec)
         .map_err(|_| Error::StringError(format!("unable to read pubspec.lock in {}", dart_root)))?;
-    let pubspec: Pubspec = serde_yaml::from_str(&pubspec)
-        .map_err(|_| Error::StringError(format!("unable to parse pubspec.lock in {}", dart_root)))?;
+    let pubspec: Pubspec = serde_yaml::from_str(&pubspec).map_err(|_| {
+        Error::StringError(format!("unable to parse pubspec.lock in {}", dart_root))
+    })?;
     if pubspec.packages.contains_key("flutter") {
         return Ok(Context::Flutter);
     }
@@ -300,18 +307,18 @@ pub fn guess_context(dart_root: &str) -> std::result::Result<Context, Error> {
 #[derive(Debug, PartialEq)]
 pub enum Context {
     Dart,
-    Flutter
+    Flutter,
 }
 
 #[derive(Debug, Deserialize)]
 struct Pubspec {
-  pub packages: HashMap<String, Info>,
+    pub packages: HashMap<String, Info>,
 }
 
 #[derive(Debug, Deserialize)]
 #[allow(dead_code)]
 struct Info {
-  pub version: String,
+    pub version: String,
 }
 
 #[cfg(test)]
@@ -324,15 +331,14 @@ mod tests {
     fn guess_dart_context() {
         let root = env!("CARGO_MANIFEST_DIR");
         let at = PathBuf::from(root)
-        .join("..")
-        .join("frb_example")
-        .join("pure_dart")
-        .join("dart")
-        .into_os_string()
-        .into_string()
-        .unwrap();
-        let context = guess_context(&at)
-        .expect("can get context from frb_example/pure_dart/dart");
+            .join("..")
+            .join("frb_example")
+            .join("pure_dart")
+            .join("dart")
+            .into_os_string()
+            .into_string()
+            .unwrap();
+        let context = guess_context(&at).expect("can get context from frb_example/pure_dart/dart");
         assert_eq!(context, Context::Dart);
     }
 
@@ -340,14 +346,13 @@ mod tests {
     fn guess_flutter_context() {
         let root = env!("CARGO_MANIFEST_DIR");
         let at = PathBuf::from(root)
-        .join("..")
-        .join("frb_example")
-        .join("with_flutter")
-        .into_os_string()
-        .into_string()
-        .unwrap();
-        let context = guess_context(&at)
-        .expect("can get context from frb_example/with_flutter");
+            .join("..")
+            .join("frb_example")
+            .join("with_flutter")
+            .into_os_string()
+            .into_string()
+            .unwrap();
+        let context = guess_context(&at).expect("can get context from frb_example/with_flutter");
         assert_eq!(context, Context::Flutter);
     }
 }
--- a/frb_codegen/src/commands.rs
+++ b/frb_codegen/src/commands.rs
@@ -313,3 +313,41 @@ struct Pubspec {
 struct Info {
   pub version: String,
 }
+
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use crate::commands::{guess_context, Context};
+
+    #[test]
+    fn guess_dart_context() {
+        let root = env!("CARGO_MANIFEST_DIR");
+        let at = PathBuf::from(root)
+        .join("..")
+        .join("frb_example")
+        .join("pure_dart")
+        .join("dart")
+        .into_os_string()
+        .into_string()
+        .unwrap();
+        let context = guess_context(&at)
+        .expect("can get context from frb_example/pure_dart/dart");
+        assert_eq!(context, Context::Dart);
+    }
+
+    #[test]
+    fn guess_flutter_context() {
+        let root = env!("CARGO_MANIFEST_DIR");
+        let at = PathBuf::from(root)
+        .join("..")
+        .join("frb_example")
+        .join("with_flutter")
+        .into_os_string()
+        .into_string()
+        .unwrap();
+        let context = guess_context(&at)
+        .expect("can get context from frb_example/with_flutter");
+        assert_eq!(context, Context::Flutter);
+    }
+}
--- a/frb_codegen/src/commands.rs
+++ b/frb_codegen/src/commands.rs
@@ -285,7 +285,7 @@ pub fn build_runner(dart_root: &str) -> Result {
 }
 
 fn guess_context(dart_root: &str) -> anyhow::Result<Context> {
-    info!("Guessing context the runner is run into");
+    debug!("Guessing context the runner is run into");
     let pubspec = PathBuf::from(dart_root).join("pubspec.lock");
     if !pubspec.exists() {
         return Err(anyhow::Error::msg(format!(
--- a/frb_codegen/src/commands.rs
+++ b/frb_codegen/src/commands.rs
@@ -257,7 +257,7 @@ pub fn format_dart(path: &str, line_length: i32) -> Result {
 
 pub fn build_runner(dart_root: &str) -> Result {
     info!("Running build_runner at {}", dart_root);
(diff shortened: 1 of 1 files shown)


Roms1383:
A hacky way would be to check if there's a `flutter` property in the root `pubspec.yaml`, but there might be a better way that I don't know about ?
Or you would prefer to let the user define it as an env var ?

SecondFlight:
I appreciate this PR! I think this addresses a significant pain point with the codegen script.

The easiest way I can think of to tell if this is a Flutter project is as follows:

1. Check if the `flutter` command exists. If not, then it doesn't matter whether this is a Flutter project because `flutter run build_runner build` won't work in the first place.
2. Run both `dart pub deps -s compact` and `flutter pub deps -s compact`. One of these commands may fail for the same reason that `dart run build_runner build` sometimes fails, but that's okay.
3. Collect the output from step 2 and check for a line in either output that starts with `- flutter` (regex: `^- flutter`). If you find this line in either output, then it's a Flutter project. If not, then it's not a Flutter project.

SecondFlight:
According to [the docs for pub deps](https://dart.dev/tools/pub/cmd/pub-deps) it looks like there's a `--json` flag which spits out JSON instead of a list with arbitrary formatting. This might be easier to parse, or at least more resilient to future changes in the tool.

fzyzcjy:
> I'm gonna need your recommendation @fzyzcjy on how to correctly differentiate a Dart from a Flutter project, hence the draft.

My personal suggestion: Check `pubspec.lock` (not `pubspec.yaml`). If there is a dependency name called "flutter" (note you will need to parse the YAML and see all those dependency names, instead of manipulating it as a string), it is a flutter lib.

I use this approach personally in all my own scripts internally.

frb_codegen/src/commands.rs
```diff
@@ -254,15 +254,21 @@ pub fn format_dart(path: &str, line_length: i32) -> Result {
 
 pub fn build_runner(dart_root: &str) -> Result {
     info!("Running build_runner at {}", dart_root);
+    // what is the correct way to differentiate a Dart project from a Flutter project ?
+    // or is it better to let the user specify it, let's say with an env var ?
+    let is_dart = true; // FIXME: determine whether it's a Dart or Flutter project
+    let cmd = if is_dart { "dart run build_runner build" } else { "flutter pub run build_runner build" };
```
fzyzcjy, line 260: uber-nit: maybe we can extract the common "run build_runner build"?

fzyzcjy (commented):
Good job

frb_codegen/src/commands.rs
Roms1383: Yes, but there's a slight difference:
- in dart : `dart run build_runner build` (no `pub`)
- in flutter : `flutter pub run build_runner build` (notice the mandatory `pub`)

Roms1383:
Ok I added the method based on @fzyzcjy recommendation since it doesn't involve running a shell command as in @SecondFlight recommendation, if that's ok with both of you guys. I have tried it manually and would like some recommendation for unit-test : my idea is to load the `pubspec.lock` files from the examples (`frb_example/pure_dart` and `frb_example/with_flutter` at least) and assert the results. Does this sound good to you guys ?

Roms1383:
Oh please also let me know if the `pubspec.lock` parsing is correct @fzyzcjy : for example if I should check more properties (like `transitive` or other ones).

frb_codegen/src/commands.rs
Roms1383: I tested both on my external project :
:white_check_mark: _with pub_

```sh
➜  cerclo-app git:(main) ✗ flutter pub run build_runner build
[INFO] Generating build script...
[INFO] Generating build script completed, took 171ms

[INFO] Initializing inputs
[INFO] Reading cached asset graph...
[INFO] Reading cached asset graph completed, took 25ms

[INFO] Checking for updates since last build...
[INFO] Checking for updates since last build completed, took 279ms

[INFO] Running build...
[INFO] Running build completed, took 5ms

[INFO] Caching finalized dependency graph...
[INFO] Caching finalized dependency graph completed, took 16ms

[INFO] Succeeded after 25ms with 0 outputs (0 actions)
```

:x: _without pub_
```sh
➜  cerclo-app git:(main) ✗ flutter run build_runner build 
Target file "build_runner" not found.
```

SecondFlight:
Anything is fine with me as long as it works :) In fact, I think the code you have here is cleaner than what I was suggesting.

frb_codegen/src/commands.rs
fzyzcjy: Sure. So maybe "flutter pub" and "dart", i.e. `build_runner build` are indeed the same

fzyzcjy:
> would like some recommendation for unit-test

Good code habit! I guess your idea is sufficient.

frb_codegen/src/commands.rs
```diff
@@ -274,3 +283,88 @@ pub fn build_runner(dart_root: &str) -> Result {
     }
     Ok(())
 }
+
+pub fn guess_context(dart_root: &str) -> std::result::Result<Context, Error> {
```
fzyzcjy, line 287: uber-nit:

* maybe no need for `pub`？
* maybe use `anyhow::Result`?

frb_codegen/src/commands.rs
```diff
@@ -274,3 +283,88 @@ pub fn build_runner(dart_root: &str) -> Result {
...
+    info!("Guessing context the runner is run into");
```
fzyzcjy, line 288: uber-nit: Well maybe `debug!` is enough

frb_codegen/src/commands.rs
```diff
@@ -274,3 +283,88 @@ pub fn build_runner(dart_root: &str) -> Result {
...
+        Error::StringError(format!("unable to parse pubspec.lock in {}", dart_root))
+    })?;
+    if pubspec.packages.contains_key("flutter") {
+        return Ok(Context::Flutter);
+    }
+    Ok(Context::Dart)
+}
+
+#[derive(Debug, PartialEq)]
+pub enum Context {
```
fzyzcjy, line 308: same as above (and below) - the `pub` thing

frb_codegen/src/commands.rs
```diff
@@ -274,3 +283,88 @@ pub fn build_runner(dart_root: &str) -> Result {
...
+    Flutter,
+}
+
+#[derive(Debug, Deserialize)]
+struct Pubspec {
+    pub packages: HashMap<String, Info>,
+}
+
+#[derive(Debug, Deserialize)]
+#[allow(dead_code)]
```
fzyzcjy, line 319: uber-nit: can this be even an empty struct? s.t. we do not need to add `allow(dead_code)`

frb_codegen/src/commands.rs
fzyzcjy, line 308: Well I am considering the naming... "Context" does not seem to be very intuitive I guess...

fzyzcjy (commented):
Good job :) Except for minor changes

frb_codegen/src/commands.rs
Roms1383: fixed in d93a792ef32befbcf28331b7fe6dca3be5630e9c and fa4ecba03052ef00fc754fc3636c7519735c54cf

frb_codegen/src/commands.rs
Roms1383: oh yes indeed, in 1aa61fa9336101263bfe2ec47f8055986dcb4429

frb_codegen/src/commands.rs
Roms1383: ok in 2eb2d45fcbeac9d6c47c78fcebd4f84dd47d555c

frb_codegen/src/commands.rs
Roms1383: oh well spotted indeed 👌 probably like 8ab66e0bfcecb08475e2ee903e665663407937aa

frb_codegen/src/commands.rs
Roms1383: ok what about 59e00c119ef3bba900efef2fd2e5940b40da19bc ?

Roms1383: sorry this one too 86dba0970f1daf016a37f95a583ed94b8bda8671

frb_codegen/src/commands.rs
Roms1383: Toolchain ?

frb_codegen/src/commands.rs
Roms1383: do you want to simplify even some more while we're at it ?
Screen Shot 2565-07-29 at 20 16 32

fzyzcjy: LGTM :)

frb_codegen/src/commands.rs
```diff
@@ -274,3 +280,91 @@ pub fn build_runner(dart_root: &str) -> Result {
...
+    fn as_run_command(&self) -> &'static str {
+        match self {
+            Context::Dart => "dart",
+            Context::Flutter => "flutter pub",
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+struct Pubspec {
```
fzyzcjy, line 320: uber-nit: `PubspecLock` (since o/w people may think it is pubspec yaml not pubspec lock)

frb_codegen/src/commands.rs
```diff
@@ -274,3 +280,91 @@ pub fn build_runner(dart_root: &str) -> Result {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use crate::commands::{guess_context, Context};
+
+    #[test]
+    fn guess_dart_context() {
```
fzyzcjy, line 331: For these tests: shall we extract a common helper like `fn guess_context_base(path: Path, expect_context: Context)`, such that no duplicate of code for these.

frb_codegen/src/commands.rs
```diff
@@ -274,3 +280,91 @@ pub fn build_runner(dart_root: &str) -> Result {
     }
     Ok(())
 }
+
+fn guess_context(dart_root: &str) -> anyhow::Result<Context> {
```
fzyzcjy, line 284: uber-nit: Btw what do you think about extracting all these into a new file

fzyzcjy (commented):
Only uber-nit now :)

frb_codegen/src/commands.rs
fzyzcjy: maybe DartToolchain? (since toolchain can mean rust toolchain etc)

Roms1383: if you're not annoyed by `DartToolchain::Dart` repetition, then I'm fine I'll make the changes.

fzyzcjy: That sounds good to me :)

frb_codegen/src/commands.rs
Roms1383: Ok if no further modifications, let's resolve this one.

frb_codegen/src/commands.rs
Roms1383: I personally think it's unnecessary right now since it's very small (especially with method refactoring as you suggested above).

frb_codegen/src/commands.rs
Roms1383: Ah yes this makes sense, I'm on it.

frb_codegen/src/commands.rs
Roms1383: Ok done in 4c7f1d46b74c858ef167bebae26f13e852240a1e

frb_codegen/src/commands.rs
Roms1383: Done in e241da7321605672cf0d70ab178dfbd49bdbfb51

frb_codegen/src/commands.rs
Roms1383: Done in a0f97a19d839719db0b3e1e3e0d029c0d4fe6d20 except `path` parameter being `PathBuf` for simplicity.

Roms1383: You might actually like [test-case](https://crates.io/crates/test-case) crate if your testing base grows with similar units  :)

Roms1383: Also renamed every var 'context' into 'toolchain' for clarity with de2f7ea5d97c34c301ad3e9ee20c742d951f1422

Roms1383:
I'm gonna mark it as _ready for review_. I suggest you squash the commits to avoid polluting your commits history.

frb_codegen/src/commands.rs
```diff
@@ -274,3 +280,88 @@ pub fn build_runner(dart_root: &str) -> Result {
...
+}
+
+#[derive(Debug, PartialEq)]
+enum DartToolchain {
+    Dart,
+    Flutter,
+}
+
+impl DartToolchain {
+    fn as_run_command(&self) -> &'static str {
```
fzyzcjy, line 311: neat!

fzyzcjy:
Now waiting for CI to pass and I am ready to merge it :)

frb_codegen/src/commands.rs
Roms1383: if `PathBuf` annoys you in the method signature, I can still change it to `&Path` and convert back and forth with `.as_path()` and `.to_path_buf()`. Just let me know it's a one minute change.

Roms1383:
I unresolved a former discussion just to make sure you're ok with the current state of the code, and suggested if you want me to [refactor a bit more](https://github.com/fzyzcjy/flutter_rust_bridge/pull/599#discussion_r933723372).

frb_codegen/src/commands.rs
Roms1383: this is what I have in mind:
test-changes

fzyzcjy: No worries - both LGTM :)

fzyzcjy:
No worries, both LGTM, it depends on you which you like more :)

frb_codegen/src/commands.rs
Roms1383: ok concise way in 1bed76d2630edb5f629ecf020c41ef73bea8cac7

fzyzcjy:
🎉

fzyzcjy:
@all-contributors please add @Roms1383 for code

PR #605, opened 2022-08-01
Avoid global ffigen

Roms1383:
Fixes #602.

## Checklist

- [x] An issue to be fixed by this PR is listed above.
- [ ] New tests are added to ensure new features are working. End-to-end tests are usually in the `./frb_example/pure_dart` example, more specifically, `rust/src/api.rs` and `dart/lib/main.dart`.
- [x] The code generator has been run, and code changes are commited. (Otherwise, CI will fail.)
- [x] If this PR adds/changes features, documentations (in the `./book` folder) are updated.
- [ ] CI is passing.

PR #613, opened 2022-08-03
Chore/improve version checking

Roms1383:
This PR aims at improving version checking for frb_codegen.
It follows up this [discussion](https://github.com/fzyzcjy/flutter_rust_bridge/pull/605#discussion_r935180160) on former PR.

## Checklist

- [ ] The code generator has been run, and code changes are commited. (Otherwise, CI will fail.)
- [ ] CI is passing.

Remark: If the PR is submitted but I do not reply for a few days, maybe I just did not see it, so please ping me.


Roms1383:
It's far from being ready but I have a question @fzyzcjy : some methods return a `crate::error::Error`, others return `anyhow::Error`, which leads sometimes to being required to downcast (if an inner function return `anyhow` based error, but top-level function is `crate` based error for example), or lose informations from the source of the initial error.
Could we harmonize it to make our lives easier ?

fzyzcjy:
> It's far from being ready but I have a question @fzyzcjy : some methods return a `crate::error::Error`, others return `anyhow::Error`, which leads sometimes to being required to downcast (if an inner function return `anyhow` based error, but top-level function is `crate` based error for example), or lose informations from the source of the initial error. Could we harmonize it to make our lives easier ?

Totally agree! The original (very old) version was a command line app so it used anyhow, and later we added support such that it can be used as a lib, so the thiserror is used as well. Mixing the two does not sound wise.

I guess most *libraries* (not *applications*) use thiserror? So we should stick to it as well.

But seems that thiserror does not have good support about backtraces...

Roms1383:
Yeah this is not an easy pick especially that I would like to avoid refactoring too much bits.
The latest commit ba247763502121331fe1a776e209a34aaeb788ce can provide a temporary workaround in the form of :
```rust
impl From<anyhow::Error> for Error {
    fn from(e: anyhow::Error) -> Self {
        if let Some(e) = e.downcast_ref::<Error>() {
            return e.clone();
        }
        Error::StringError(e.to_string())
    }
}
```

Roms1383:
So far this PR will ensure that a dependency e.g. `ffi`, `ffigen` :
- has been correctly specified in `pubspec.yaml` (not performing any range requirement check)
- if it does, ensure that it has been correctly pinned in `pubspec.lock` (performing a range requirement check on the version found there)

I have the feeling that [semver_constraints](https://crates.io/crates/semver_constraints) crate could allow to improve the logic by checking directly in `pubspec.yaml`, no matter if an exact version or a range of versions is specified. But to be honest I haven't looked any further yet.

Should we go step by step and review / maybe merge this one first ? @fzyzcjy

fzyzcjy:
The logic in this PR already looks quite good to me, and I am ready to briefly review and then merge it. Good job!

In my humble opinion, I guess it may also be a great choice to have a look at [some other issues (feature requests / bugs)](https://github.com/fzyzcjy/flutter_rust_bridge/issues), compared with further improving this version checking in future PRs.

Roms1383:
> In my humble opinion, I guess it may also be a great choice to have a look at [some other issues (feature requests / bugs)](https://github.com/fzyzcjy/flutter_rust_bridge/issues), compared with further improving this version checking in future PRs.

utterly agree ^^

frb_codegen/src/commands.rs
```diff
@@ -18,31 +21,32 @@ fn call_shell(cmd: &str) -> Output {
...
-    let ffi = ensure_dependencies(dart_root, "ffi", PackageManager::Dependencies).unwrap();
-    if ffi.is_none() {
-        return Err(Error::MissingDep {
-            name: "ffi".into(),
-            context: PackageManager::Dependencies,
-            version: "^2.0.1".into(),
-        });
+    if let Err(e) = repo.has_specified("ffi", PackageManager::Dependencies, &FFI_REQUIREMENT) {
+        return Err(e.into());
+    }
```
fzyzcjy, line 32: Seems Rust has an operator `?` for this usage: `some_function_that_return_result()?` means `if error return error`

p.s. the same applies for 4 places around this code

frb_codegen/src/tools.rs
```diff
@@ -0,0 +1,442 @@
...
+                "unable to parse {} in {}",
+                DartToolchain::manifest_filename(),
+                at
+            ))
+        })?;
+        let deps = if manager == PackageManager::DevDependencies {
+            manifest_file.dev_dependencies.unwrap_or_default()
+        } else {
+            manifest_file.dependencies.unwrap_or_default()
+        };
```
fzyzcjy, line 234: uber-nit: use `match` maybe

frb_codegen/src/tools.rs
```diff
@@ -0,0 +1,442 @@
...
+                "unable to parse {} in {}",
+                DartToolchain::lock_filename(),
+                at
+            ))
+        })?;
+        let dependency = lock_file.packages.get(package);
+        let version = match dependency {
+            Some(dependency) => {
+                let pm = dependency.installed_in();
+                if pm.is_none() || (pm.unwrap() != manager) {
```
fzyzcjy, line 264: uber-nit: `pm != Some(manager)`

fzyzcjy (commented):
Neat code :) Only some nits and I am ready to merge it

frb_codegen/src/commands.rs
Roms1383: done in d7f913f9665145617c107e8a4d529226be3b665e

frb_codegen/src/tools.rs
Roms1383: damn I didn't even know it was possible to check a condition like this lol
done in 47037b7d40a6afe68d16be5fff38c3253e9fa7cb

fzyzcjy:
🎉

PR #619, opened 2022-08-06
enhance mirror

Cupnfish:

## Checklist

- [x] The code generator has been run, and code changes are commited. (Otherwise, CI will fail.)
- [x] CI is passing.

Remark for PR creator: If the PR is submitted but the owner (fzyzcjy) does not reply for a few days, maybe he just did not see it, so please ping him.

### DIFF
--- a/frb_codegen/src/markers.rs
+++ b/frb_codegen/src/markers.rs
@@ -11,15 +11,19 @@ pub fn extract_mirror_marker(attrs: &[Attribute]) -> Vec<Path> {
                     path,
                     nested: mirror,
                     ..
-                })) if path.is_ident("mirror") && !mirror.is_empty() => {
-                    let mut res = Vec::with_capacity(mirror.len());
-                    for mirror_lable in mirror.into_iter() {
-                        if let NestedMeta::Meta(Meta::Path(path)) = mirror_lable {
-                            res.push(path.clone());
-                        }
-                    }
-                    Some(res)
-                }
+                })) if path.is_ident("mirror") && !mirror.is_empty() => Some(
+                    mirror
+                        .into_iter()
+                        .filter_map(|lable| {
+                            if let NestedMeta::Meta(Meta::Path(path)) = lable {
+                                Some(path)
+                            } else {
+                                None
+                            }
+                        })
+                        .cloned()
+                        .collect::<Vec<_>>(),
+                ),
                 _ => None,
             }),
             _ => None,
--- a/frb_codegen/src/markers.rs
+++ b/frb_codegen/src/markers.rs
@@ -1,7 +1,7 @@
 use syn::*;
 
 /// Extract a path from marker `#[frb(mirror(path), ..)]`
-pub fn extract_mirror_marker(attrs: &[Attribute]) -> Option<Path> {
+pub fn extract_mirror_marker(attrs: &[Attribute]) -> Vec<Path> {
     attrs
         .iter()
         .filter(|attr| attr.path.is_ident("frb"))
@@ -11,16 +11,20 @@ pub fn extract_mirror_marker(attrs: &[Attribute]) -> Option<Path> {
                     path,
                     nested: mirror,
                     ..
-                })) if path.is_ident("mirror") && mirror.len() == 1 => {
-                    match mirror.first().unwrap() {
-                        NestedMeta::Meta(Meta::Path(path)) => Some(path.clone()),
-                        _ => None,
+                })) if path.is_ident("mirror") && !mirror.is_empty() => {
+                    let mut res = Vec::with_capacity(mirror.len());
+                    for mirror_lable in mirror.into_iter() {
+                        if let NestedMeta::Meta(Meta::Path(path)) = mirror_lable {
+                            res.push(path.clone());
+                        }
                     }
+                    Some(res)
                 }
                 _ => None,
             }),
             _ => None,
         })
+        .unwrap_or_default()
 }
 
 /// Checks if the `#[frb(non_final)]` attribute is present.
--- a/frb_codegen/src/source_graph.rs
+++ b/frb_codegen/src/source_graph.rs
@@ -11,7 +11,7 @@ use std::{collections::HashMap, fmt::Debug, fs, path::PathBuf};
 
 use cargo_metadata::MetadataCommand;
 use log::{debug, warn};
-use syn::{Attribute, Ident, ItemEnum, ItemStruct, UseTree};
+use syn::{Attribute, Ident, ItemEnum, ItemStruct, PathArguments, UseTree};
 
 use crate::markers;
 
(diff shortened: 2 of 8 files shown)


Cupnfish:
before: 
```rust
#[frb(mirror(MessageId))]
pub struct MId(pub [u8; 32]);

#[frb(mirror(BlobId))]
pub struct BId(pub [u8; 32]);

#[frb(mirror(FeedId))]
pub struct FId(pub [u8; 32]);
```
now:
```rust
#[frb(mirror(MessageId, BlobId, FeedId))]
pub struct Id(pub [u8; 32]);
```

frb_codegen/src/markers.rs
```diff
@@ -11,16 +11,20 @@ pub fn extract_mirror_marker(attrs: &[Attribute]) -> Option<Path> {
...
-                        NestedMeta::Meta(Meta::Path(path)) => Some(path.clone()),
-                        _ => None,
+                })) if path.is_ident("mirror") && !mirror.is_empty() => {
+                    let mut res = Vec::with_capacity(mirror.len());
+                    for mirror_lable in mirror.into_iter() {
+                        if let NestedMeta::Meta(Meta::Path(path)) = mirror_lable {
+                            res.push(path.clone());
+                        }
                     }
+                    Some(res)
```
fzyzcjy, line 21: nit: Maybe use iterator (map, filter, etc) so we can avoid `mut`

frb_codegen/src/source_graph.rs
```diff
@@ -183,10 +183,11 @@ impl Debug for Module {
 }
 
 /// Get a struct or enum ident, possibly remapped by a mirror marker
-fn get_ident(ident: &Ident, attrs: &[Attribute]) -> (Ident, bool) {
+fn get_ident(attrs: &[Attribute]) -> Vec<Ident> {
     markers::extract_mirror_marker(attrs)
-        .and_then(|path| path.get_ident().map(|ident| (ident.clone(), true)))
-        .unwrap_or_else(|| (ident.clone(), false))
+        .into_iter()
+        .filter_map(|path| path.get_ident().map(|ident| ident.clone()))
```
fzyzcjy, line 189: uber-nit: Do we need that clone?

frb_codegen/src/source_graph.rs
```diff
@@ -209,34 +210,70 @@ impl Module {
...
+                                visibility: syn_vis_to_visibility(&item_struct.vis),
+                                path: {
+                                    let mut path = self.module_path.clone();
+                                    path.push(ident_str);
+                                    path
+                                },
+                                mirror: true,
+                            });
+                        }
+                    }
```
fzyzcjy, line 243: nit: Is it possible to extract a function, or, e.g. something like `idents = get_ident(&item_struct.attrs).or_else(vec![item_struct.ident.clone()])`? Seems we are duplicating some code currently here

frb_codegen/src/source_graph.rs
```diff
@@ -209,34 +210,70 @@ impl Module {
...
+                                visibility: syn_vis_to_visibility(&item_enum.vis),
+                                path: {
+                                    let mut path = self.module_path.clone();
+                                    path.push(ident_str);
+                                    path
+                                },
+                                mirror: true,
+                            });
+                        }
+                    }
```
fzyzcjy, line 276: same comments as above. Btw would be even greater if can extract common code between enum and struct handling

frb_codegen/src/markers.rs
```diff
@@ -1,7 +1,7 @@
 use syn::*;
 
 /// Extract a path from marker `#[frb(mirror(path), ..)]`
-pub fn extract_mirror_marker(attrs: &[Attribute]) -> Option<Path> {
+pub fn extract_mirror_marker(attrs: &[Attribute]) -> Vec<Path> {
```
fzyzcjy, line 4: Please add a bit of test in `pure_dart` api.rs and main.dart, so that we can ensure it works now and works in the future :)

(unrelated to this line of code; just b/c no place to put this comment)

fzyzcjy (commented):
Good job! Just a few nits

fzyzcjy:
Btw I am curious - what is the use case of this?

frb_codegen/src/source_graph.rs
Cupnfish: Yes, `clone` is needed here, the `ident` returned must be owned, as it was in previous versions.

fzyzcjy: Oh yes, I did not check types when viewing it in github

Cupnfish:
> Btw I am curious - what is the use case of this?

A simple test has been added, but it is still not enough to show that the feature exists. In fact, as in the case [I provided](https://github.com/fzyzcjy/flutter_rust_bridge/pull/619#issuecomment-1207223773), in some cases there will be structures with different names that have the same internal structure, such structures will be used in different scenarios and they will greatly improve the readability of the code, but we need multiple different mirrors when we need a mirror, which makes the code bloated.

Cupnfish:
@fzyzcjy I think it's ready for a new round of reviews, and I've made the appropriate changes to the ones you mentioned before.

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -531,6 +532,17 @@ pub fn get_message() -> ApplicationMessage {
...
+pub struct _Numbers(pub Vec<i32>);
+
+pub fn repeat_number(num: i32, times: usize) -> Numbers {
+    external_lib::repeat_number(num, times)
+}
+
+pub fn repeat_sequence(seq: i32, times: usize) -> Sequences {
+    external_lib::repeat_sequences(seq, times)
+}
+
```
fzyzcjy, line 545: Would be great if there are a little bit more tests. For example: Functions that *accept* Numbers/Sequences as arguments. This is because rust functions that *accept arguments of type T* is implemented very differently from *return type T*.

frb_example/pure_dart/dart/lib/bridge_generated.dart
```diff
@@ -28,7 +28,11 @@ abstract class FlutterRustBridgeExampleSingleBlockTest {
   /// Newlines are preserved.
   ///
   Future<int> primitiveTypes(
-      {required int myI32, required int myI64, required double myF64, required bool myBool, dynamic hint});
+      {required int myI32,
```
fzyzcjy, line 31: Btw you need to format the code :) I will review again after code is formatted, since currently there are so many diff (and most are from code formats)

frb_codegen/src/source_graph.rs
```diff
@@ -209,34 +226,40 @@ impl Module {
...
-                            path.push(ident_str);
-                            path
-                        },
-                        mirror,
-                    });
+                    let (idents, mirror) = get_ident(&item_struct.ident, &item_struct.attrs);
+
+                    for ident in idents {
+                        let ident_str = ident.to_string();
+                        scope_structs.push(Struct {
```
fzyzcjy, line 233: nit: Is it possible we use something like a `.map()`, `.extend()` etc?

fzyzcjy (commented):
Good job, except for a few minor problems :)

frb_example/pure_dart/dart/lib/bridge_generated.dart
Cupnfish: I use the dart plugin for my vscode and it does the formatting after saving. I don't know what formatting tool you are using and do you have documentation on the formatting?

fzyzcjy: https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/justfile IIRC the `refresh_all` works

Cupnfish:
I've got the code formatted so you can review it again. @fzyzcjy

fzyzcjy:
Sure. I clicked the "run CI" button

Cupnfish:
@fzyzcjy I fixed the error prompted in CI where I was treating 1 as first, when in fact 0 was the first in the array.One more CI and it should be fine.

frb_example/pure_dart/dart/lib/bridge_generated.dart
```diff
@@ -540,6 +564,14 @@ class Point {
   });
 }
 
+class Sequences {
+  final Int32List field0;
+
+  Sequences({
+    required this.field0,
+  });
+}
```
fzyzcjy, line 573: So is this expected behavior? Numbers and Sequences are still two classes, not one

fzyzcjy (commented):
Good job, I am going to merge it :)

frb_example/pure_dart/dart/lib/bridge_generated.dart
Cupnfish: @fzyzcjy Yes, this is the desired result, less template code and more mirrors generated.

fzyzcjy:
🎉

fzyzcjy:
@all-contributors please add @Cupnfish for code

PR #620, opened 2022-08-06
Support array

Cupnfish:
Fixes #572

## Checklist

- [ ] An issue to be fixed by this PR is listed above.
- [ ] New tests are added to ensure new features are working. End-to-end tests are usually in the `./frb_example/pure_dart` example, more specifically, `rust/src/api.rs` and `dart/lib/main.dart`.
- [ ] The code generator has been run, and code changes are commited. (Otherwise, CI will fail.)
- [ ] If this PR adds/changes features, documentations (in the `./book` folder) are updated.
- [ ] CI is passing.

Remark for PR creator: If the PR is submitted but the owner (fzyzcjy) does not reply for a few days, maybe he just did not see it, so please ping him.

PR #623, opened 2022-08-08
init suport array

### LINKED ISSUE #572: Error converting 32 byte array in new release
### Describe the bug

After upgrade to 1.37.2, it breaks arrays like
```rust
struct Foo {
    bar: [u8; 32]
}
```
compiling that produces the error from the `bridge_generated.rs` file:
```
the trait `Wire2Api<[u8; 32]>` is not implemented for `*mut wire_uint_8_list`
```

```shell
error[E0277]: the trait bound `*mut wire_uint_8_list: Wire2Api<[u8; 32]>` is not satisfied
   --> NativeUI/src/bridge_generated.rs:159:39
    |
159 |             wallet_id: self.wallet_id.wire2api(),
    |                                       ^^^^^^^^ the trait `Wire2Api<[u8; 32]>` is not implemented for `*mut wire_uint_8_list`
    |
    = help: the following other types implement trait `Wire2Api<T>`:
              <*mut wire_uint_8_list as Wire2Api<String>>
              <*mut wire_uint_8_list as Wire2Api<Vec<u8>>>

For more information about this error, try `rustc --explain E0277`.
error: could not compile `my_crate` due to previous error
```

### To Reproduce

Create a struct with [u8; 32] as a field and call it from a function

### Expected behavior

_No response_

### Generated binding code

_No response_

### OS

Ubuntu 20.05

1.37.2

### Flutter info

```shell
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.0.4, on KDE neon User - 5.25 5.4.0-121-generic, locale en_GB.UTF-8)
Checking Android licenses is taking an unexpectedly long time...[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
[✓] Chrome - develop for the web
[✓] Linux toolchain - develop for Linux desktop
[!] Android Studio (not installed)
[✓] VS Code (version 1.69.0)
[✓] Connected device (2 available)
[✓] HTTP Host Availability

! Doctor found issues in 1 category.
```

### Version of `clang++`

10.0.0-4ubuntu1

### Version of `ffigen`

 6.0.1

### Additional context

_No response_

Cupnfish:
Close #572

## Checklist

- [x] CI is passing.

Remark for PR creator: If the PR is submitted but the owner (fzyzcjy) does not reply for a few days, maybe he just did not see it, so please ping him.

### DIFF
--- a/frb_codegen/src/generator/dart/ty_delegate.rs
+++ b/frb_codegen/src/generator/dart/ty_delegate.rs
@@ -11,8 +11,9 @@ impl TypeDartGeneratorTrait for TypeDelegateGenerator<'_> {
         match self.ir {
             IrTypeDelegate::GeneralArray { .. } | IrTypeDelegate::PrimitiveArray { .. } => {
                 let body = format!(
-                    "return api2wire_{}(raw.inner);",
-                    self.ir.get_delegate().safe_ident()
+                    "return api2wire_{}({});",
+                    self.ir.get_delegate().safe_ident(),
+                    self.ir.array_cast_string()
                 );
                 Acc::distribute(Some(body))
             }
@@ -133,19 +134,19 @@ impl TypeDartGeneratorTrait for TypeDelegateGenerator<'_> {
             | IrTypeDelegate::PrimitiveArray { length, .. } => {
                 format!(
                     "
-                class {0} extends NonGrowableListView<{2}> {{
-                    {0}({1} inner)
+                class {0} extends NonGrowableListView<{1}> {{
+                    {0}({2} inner)
                         : assert(inner.length == {length}),
                           super(inner);
-                    {0}.unchecked({1} inner)
+                    {0}.unchecked({2} inner)
                         : super(inner);
-                    {1} get inner => {3};
+                    {3}
                   }}
                 ",
                     self.ir.dart_api_type(),
-                    self.ir.get_delegate().dart_api_type(),
                     self.ir.inner_dart_api_type(),
-                    self.ir.array_cast_string()
+                    self.ir.get_delegate().dart_api_type(),
+                    self.ir.dart_init_method(),
                 )
             }
             _ => "".into(),
--- a/frb_codegen/src/generator/dart/ty_delegate.rs
+++ b/frb_codegen/src/generator/dart/ty_delegate.rs
@@ -9,6 +9,13 @@ type_dart_generator_struct!(TypeDelegateGenerator, IrTypeDelegate);
 impl TypeDartGeneratorTrait for TypeDelegateGenerator<'_> {
     fn api2wire_body(&self) -> Acc<Option<String>> {
         match self.ir {
+            IrTypeDelegate::GeneralArray { .. } | IrTypeDelegate::PrimitiveArray { .. } => {
+                let body = format!(
+                    "return api2wire_{}(raw.inner);",
+                    self.ir.get_delegate().safe_ident()
+                );
+                Acc::distribute(Some(body))
+            }
             IrTypeDelegate::String => Acc {
                 io: Some("return api2wire_uint_8_list(utf8.encoder.convert(raw));".into()),
                 wasm: Some("return raw;".into()),
@@ -64,6 +71,20 @@ impl TypeDartGeneratorTrait for TypeDelegateGenerator<'_> {
 
     fn wire2api_body(&self) -> String {
         match &self.ir {
+            IrTypeDelegate::GeneralArray { length, general } => format!(
+                r"final inner = (raw as List<dynamic>).map(_wire2api_{}).toList();
+                if (inner.length != {length}) throw Exception('unexpected array length: expect {length} but see ${{inner.length}}');
+                return {}.Unchecked(inner);",
+                general.safe_ident(),
+                self.ir.dart_api_type()
+            ),
+            IrTypeDelegate::PrimitiveArray { length, .. } => format!(
+                r"final inner = _wire2api_{}(raw);
+                if (inner.length != {length}) throw Exception('unexpected array length: expect {length} but see ${{inner.length}}');
+                return {}.Unchecked(inner);",
+                self.ir.get_delegate().safe_ident(),
+                self.ir.dart_api_type()
+            ),
             IrTypeDelegate::ZeroCopyBufferVecPrimitive(
                 IrTypePrimitive::I64 | IrTypePrimitive::U64,
             ) => {
@@ -102,14 +123,36 @@ impl TypeDartGeneratorTrait for TypeDelegateGenerator<'_> {
     }
 
     fn structs(&self) -> String {
-        if let IrTypeDelegate::PrimitiveEnum { ir, .. } = &self.ir {
-            super::TypeEnumRefGenerator {
+        match &self.ir {
+            IrTypeDelegate::PrimitiveEnum { ir, .. } => super::TypeEnumRefGenerator {
                 ir: ir.clone(),
                 context: self.context.clone(),
             }
-            .structs()
-        } else {
-            "".into()
+            .structs(),
+            IrTypeDelegate::GeneralArray { length, .. }
+            | IrTypeDelegate::PrimitiveArray { length, .. } => {
+                format!(
+                    "class {0} {{
+                        final {1} inner;
+
+                        {0}(this.inner): assert(inner.length == {length});
+
+                        {0}.Unchecked(this.inner);
+
+                        {2} operator [](int index) {{
+                            return inner[index];
+                        }}
+                        
+                        void operator []=(int index, {2} value) {{
+                            inner[index] = value;
+                        }}
+                }}",
+                    self.ir.dart_api_type(),
+                    self.ir.get_delegate().dart_api_type(),
+                    self.ir.inner_dart_api_type()
+                )
+            }
+            _ => "".into(),
         }
     }
 }
--- a/frb_codegen/src/generator/dart/ty_delegate.rs
+++ b/frb_codegen/src/generator/dart/ty_delegate.rs
@@ -9,24 +9,27 @@ type_dart_generator_struct!(TypeDelegateGenerator, IrTypeDelegate);
 impl TypeDartGeneratorTrait for TypeDelegateGenerator<'_> {
     fn api2wire_body(&self) -> Acc<Option<String>> {
         match self.ir {
-            IrTypeDelegate::GeneralArray { .. } => Acc::distribute(Some(format!(
-                "return api2wire_{}(raw);",
-                self.ir.get_delegate().safe_ident(),
-            ))),
-            IrTypeDelegate::PrimitiveArray { length, .. } => Acc {
-                io: Some(format!(
-                    "final ans = inner.new_{}_{}({length});
-                            ans.ref.ptr.asTypedList({length}).setAll(0, raw);
-                            return ans;",
-                    self.ir.get_delegate().safe_ident(),
-                    self.context.config.block_index,
-                )),
-                wasm: Some(format!(
-                    "return {}.fromList(raw);",
-                    self.ir.get_delegate().dart_api_type()
-                )),
-                ..Default::default()
+            IrTypeDelegate::Array(ref array) => match array {
+                IrTypeDelegateArray::GeneralArray { .. } => Acc::distribute(Some(format!(
+                    "return api2wire_{}(raw);",
+                    array.get_delegate().safe_ident(),
+                ))),
+                IrTypeDelegateArray::PrimitiveArray { length, .. } => Acc {
+                    io: Some(format!(
+                        "final ans = inner.new_{}_{}({length});
+                                    ans.ref.ptr.asTypedList({length}).setAll(0, raw);
+                                    return ans;",
+                        array.get_delegate().safe_ident(),
+                        self.context.config.block_index,
+                    )),
+                    wasm: Some(format!(
+                        "return {}.fromList(raw);",
+                        array.get_delegate().dart_api_type()
+                    )),
+                    ..Default::default()
+                },
             },
+
             IrTypeDelegate::String => Acc {
                 io: Some("return api2wire_uint_8_list(utf8.encoder.convert(raw));".into()),
                 wasm: Some("return raw;".into()),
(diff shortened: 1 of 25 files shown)


fzyzcjy:
P.S. As a side remark

> After upgrade to 1.37.2, it breaks arrays like... https://github.com/fzyzcjy/flutter_rust_bridge/issues/572

We already have some support on arrays: http://cjycode.com/flutter_rust_bridge/feature/lang_vec.html

fzyzcjy:
👀

Cupnfish:
@fzyzcjy I see that the CI error is related to a memory leak, and I was wondering how the memory check part works locally?

fzyzcjy:
> how the memory check part works locally?

Look at CI script, and mimic it to run locally. https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/.github/workflows/ci.yaml

fzyzcjy:
Btw I am a bit curious: We already have http://cjycode.com/flutter_rust_bridge/feature/lang_vec.html#t-n, i.e. array support. So could you please add a bit of doc showing what this PR does in addition to the old code?

Cupnfish:
> Btw I am a bit curious: We already have http://cjycode.com/flutter_rust_bridge/feature/lang_vec.html#t-n, i.e. array support. So could you please add a bit of doc showing what this PR does in addition to the old code?

The previous implementation only supported arrays of the results returned from the rust side , but not parameters , this time it mainly supported parameters as well , which I thought was supported by reading the documentation before , but it didn't support them , so the documentation didn't actually need to be changed this time because what the documentation said misled people into thinking that arrays were originally supported , but in fact only the values returned from the rust side were supported .

fzyzcjy:
I see. Looking forward to merging this PR!

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -668,6 +669,26 @@ pub fn get_sum_struct() -> SumWith {
...
+    id.0.iter().all(|n| n.eq(&repeat_value))
+}
+
+pub struct Logs(pub [Log; 2]);
+
+pub fn logs(log_a: Log, log_b: Log) -> Logs {
+    Logs([log_a, log_b])
+}
+
+pub fn first_log(logs: Logs) -> Log {
```
fzyzcjy, line 688: Btw, do we support things like: 

* `f(arg: [Sth; 2])`
* `f(arg: Box<[Sth; 2]>)` 

etc? If we support more kinds of things in this PR, would be great to add them to tests, such that we can verify it works correctly, and in the future nobody will accidentially break it.

frb_codegen/src/generator/rust/ty_general_list.rs
```diff
@@ -17,22 +17,36 @@ impl TypeGeneralListGenerator<'_> {
 
 impl TypeRustGeneratorTrait for TypeGeneralListGenerator<'_> {
     fn wire2api_body(&self) -> Option<String> {
-        Some(TypeGeneralListGenerator::WIRE2API_BODY.to_string())
+        Some(match &self.ir {
+            IrTypeGeneralList::GeneralList { .. } => TypeGeneralListGenerator::WIRE2API_BODY.into(),
+            IrTypeGeneralList::GeneralArray { len, .. } => format!(
```
fzyzcjy, line 22: Looks like the generated code of `GeneralArray` works very similarly as `GeneralList`. Is it possible we do not duplicate this code string here, but instead merge it with `TypeGeneralListGenerator::WIRE2API_BODY`?

frb_codegen/src/generator/rust/ty_primitive_list.rs
```diff
@@ -8,18 +8,29 @@ type_rust_generator_struct!(TypePrimitiveListGenerator, IrTypePrimitiveList);
...
-            support::vec_from_leak_ptr(wrap.ptr, wrap.len)
-        }"
+        Some(match &self.ir {
+            IrTypePrimitiveList::PrimitiveList { .. } => "unsafe {
+                    let wrap = support::box_from_leak_ptr(self);
+                    support::vec_from_leak_ptr(wrap.ptr, wrap.len)
+                }"
             .into(),
-        )
+            IrTypePrimitiveList::PrimitiveArray { len, .. } => format!(
```
fzyzcjy, line 17: Is it possible we *delegate* a PrimitiveArray using a PritmitiveList, such that we even do not need to deal with any unsafe operations!

When talking about "delegate" - for example, when implementing a `String`, we delegate it to a `Vec<u8>`. See ty_delegate.rs for details.

frb_codegen/src/generator/rust/ty_general_list.rs
fzyzcjy, line 22: Similarly, is it possible to delegate GeneralArray to a GeneralList to avoid unsafe operations

frb_codegen/src/generator/rust/ty_primitive_list.rs
```diff
@@ -8,18 +8,29 @@ type_rust_generator_struct!(TypePrimitiveListGenerator, IrTypePrimitiveList);
...
+                "unsafe {{
+                    let wrap = support::box_from_leak_ptr(self);
+                    debug_assert_eq!(wrap.len, {len});
+                    let slice = core::slice::from_raw_parts_mut(wrap.ptr, wrap.len as usize);
+                    let ptr = slice.as_mut_ptr() as *mut {};
+                    let boxed = support::box_from_leak_ptr(ptr);
+                    *boxed
```
fzyzcjy, line 24: TODO(fzyzcjy): If we use the delegate approach then it is ok; otherwise @fzyzcjy should carefully review these several unsafe code segments again

frb_codegen/src/ir/ty_primitive_list.rs
```diff
@@ -2,21 +2,41 @@ use crate::ir::*;
...
 impl IrTypeTrait for IrTypePrimitiveList {
     fn visit_children_types<F: FnMut(&IrType) -> bool>(&self, f: &mut F, _ir_file: &IrFile) {
-        f(&IrType::Primitive(self.primitive.clone()));
+        f(&IrType::Primitive(self.primitive().clone()));
     }
 
     fn safe_ident(&self) -> String {
-        self.dart_api_type().to_case(Case::Snake)
+        if let Self::PrimitiveArray { .. } = self {
+            self.dart_api_type().replace("List", "Array")
```
fzyzcjy, line 31: nit: Well, maybe we can make it less hacky here... This code depends on the current fact that dart_api_type has a "List" substring in it, but that looks a bit fragile

fzyzcjy (commented):
Good job! Only some discussions here

frb_codegen/src/generator/rust/ty_primitive_list.rs
Cupnfish: I used `delegate` when I first implemented it but got a response about whether it could be a separate module from list, the new implementation is with list and has some differences

frb_codegen/src/generator/rust/ty_general_list.rs
Cupnfish: This part now has a new change, I'm not sure if it fits in more minimally, you can check it out.

frb_codegen/src/generator/rust/ty_primitive_list.rs
Cupnfish: Even with the use of `delegate`, the use of `unsafe` cannot be avoided and I really need someone to check that this part of `unsafe` is as expected

frb_codegen/src/generator/rust/ty_primitive_list.rs
fzyzcjy: > but got a response about whether it could be a separate module from list

Did I do that? Could you please point out my comments (I am so forgetful haha since I am so busy recently)

I will recheck why I said this should not use a delegate.

frb_codegen/src/generator/rust/ty_primitive_list.rs
fzyzcjy: May I know why unsafe cannot be avoided? Since if we delegate PrimitiveArray via a PrimitiveList, we have the following code:

* convert PrimitiveArray to a PrimitiveList - safe (possibly using some 3rd party crates, but definitely can be safe)
* use existing code of transfering PrimitiveList data between Rust and Dart - unsafe, but all unsafe are old code that exist for a long time, no new code at all now.

frb_codegen/src/generator/rust/ty_primitive_list.rs
fzyzcjy: > In addition, I am not sure whether adding IrTypeDelegate::ArrayGeneral is the best approach. We already have existing support for arrays indeed. https://github.com/fzyzcjy/flutter_rust_bridge/pull/620#issuecomment-1207290477

Do you mean this one?

I see. Then yes, we should make array have only one IR, not two. But that does not conflict with delegating: What about this - we put array IR the same as current PR, but use delegating when *generating code*?

fzyzcjy: This may need a bit of refactor. Namely:

* Maybe we need to enable *every* type to be a delegate, instead of "only types in IRDelegate can be delegate".
* After that, we can simply rename IRDelegate to something else, say, IRMisc, since they are no longer the only delegates.

Cupnfish:
@fzyzcjy Learned from previous experiences and mistakes, the new refactoring has been committed for review.

frb_example/pure_dart/dart/lib/bridge_definitions.dart
```diff
@@ -752,6 +892,78 @@ class SumWith {
...
+    required this.field0,
+  });
+}
+
+class U8Array1600 {
+  final Uint8List inner;
+
+  U8Array1600(this.inner) : assert(inner.length == 1600);
+
+  U8Array1600.Unchecked(this.inner);
```
fzyzcjy, line 908: uber-nit: constructor name ("Unchecked") should be lowercase ("unchecked")

fzyzcjy (commented):
Do not have much brain power currently (busy with https://github.com/fzyzcjy/flutter_smooth), generally good job :) only some nits

frb_codegen/src/generator/dart/ty_delegate.rs
```diff
@@ -102,14 +123,36 @@ impl TypeDartGeneratorTrait for TypeDelegateGenerator<'_> {
...
                 context: self.context.clone(),
             }
-            .structs()
-        } else {
-            "".into()
+            .structs(),
+            IrTypeDelegate::GeneralArray { length, .. }
+            | IrTypeDelegate::PrimitiveArray { length, .. } => {
+                format!(
+                    "class {0} {{
```
Desdaemon, line 135: Replacing `List<T>` and typed lists with a class is an API-breaking change; perhaps have the new class implement [DelegatingList](https://pub.dev/documentation/collection/latest/collection/DelegatingList-class.html) so list methods are visible on it?

frb_codegen/src/generator/dart/ty_delegate.rs
```diff
@@ -102,14 +123,36 @@ impl TypeDartGeneratorTrait for TypeDelegateGenerator<'_> {
...
+                        {0}.Unchecked(this.inner);
+
+                        {2} operator [](int index) {{
+                            return inner[index];
+                        }}
+                        
+                        void operator []=(int index, {2} value) {{
+                            inner[index] = value;
+                        }}
+                }}",
```
Desdaemon, line 149: A lot of common methods here can be factored out, perhaps something like this:
```dart
class TArrayN extends NonGrowableListView<T> {
  TArrayN(List<T> inner) : assert(inner.length == N), super(inner);
}
```

Desdaemon (commented):
LGTM with some nits

frb_codegen/src/generator/dart/ty_delegate.rs
Cupnfish: Cool！

frb_example/pure_dart/dart/lib/bridge_definitions.dart
```diff
@@ -682,20 +651,12 @@ class FeedId {
...
-
-  void operator []=(int index, int value) {
-    inner[index] = value;
-  }
+class I32Array2 extends NonGrowableListView<int> {
+  I32Array2(Int32List inner)
+      : assert(inner.length == 2),
+        super(inner);
+  I32Array2.unchecked(Int32List inner) : super(inner);
+  Int32List get inner => Int32List.fromList(super.cast());
```
fzyzcjy, line 659: Hmm have you tried to use inner like `.inner[0] = 42` and see whether the the original I32Array2 changes? I am not sure about it, because `fromList` is said to copy elements, and `cast` here seems also unnecessary.

Indeed, when calling `.inner`, someone may not expect they get a brand new copy (with O(N) time)

fzyzcjy (commented):
Good job, only a nit!

frb_example/pure_dart/dart/lib/bridge_definitions.dart
Cupnfish: I also think the use of `inner` here is a bit out of place, my intention was that the previous implementation had an `inner` field which was used in the `wire2api` section, so I tried to restore that effect in the new implementation as well, but it worked poorly, the new patch has been committed and I'm waiting for the test results to see if it works as expected.

Cupnfish: @fzyzcjy `CI / Dart (Web) unit tests (pull_request) ` timeout, I don't know if this is accidental or common, and I don't know how to restart this failed CI on github, if you know how to restart this CI to retest this problem, please let me know.

fzyzcjy: Just submit arbitrary code (e.g. add a new line etc), and ci will rerun.

Yes it does sometimes timeout according to @Desdaemon

fzyzcjy (commented):
Only did a brief glance, LGTM :)

fzyzcjy:
Good job! Busy on https://github.com/fzyzcjy/flutter_smooth, will probably merge today if no further problems :)

frb_example/pure_dart/dart/lib/bridge_generated.dart
```diff
@@ -1299,6 +1426,12 @@ int _wire2api_SyncReturn_u8(Uint8List raw) {
   return dataView.getUint8(0);
 }
 
+TestIdArray2 _wire2api_TestId_array_2(dynamic raw) {
+  final inner = (raw as List<dynamic>).map(_wire2api_test_id).toList();
+  if (inner.length != 2) throw Exception('unexpected array length: expect 2 but see ${inner.length}');
+  return TestIdArray2.unchecked(inner);
```
fzyzcjy, line 1432: nit: seems these 2 lines can be eliminated to 1 line: `TestIdArray2(inner)`?

frb_example/pure_dart/dart/lib/bridge_definitions.dart
```diff
@@ -752,6 +853,46 @@ class SumWith {
...
+  });
+}
+
+class U8Array1600 extends NonGrowableListView<int> {
+  U8Array1600(Uint8List inner)
+      : assert(inner.length == 1600),
+        super(inner);
+  U8Array1600.unchecked(Uint8List inner) : super(inner);
+  U8Array1600.init() : super(Uint8List(1600));
+}
```
fzyzcjy, line 870: uber-nit: what about extracting 1600 to a const

```suggestion
class U8Array1600 extends NonGrowableListView<int> {
  static const arraySize = 1600;
  U8Array1600(Uint8List inner)
      : assert(inner.length == arraySize),
        super(inner);
  U8Array1600.unchecked(Uint8List inner) : super(inner);
  U8Array1600.init() : super(Uint8List(arraySize));
}
```

frb_codegen/src/ir/ty_delegate.rs
```diff
@@ -66,6 +86,55 @@ impl IrTypeDelegate {
...
+                self.inner_dart_api_type()
+            ),
+            IrTypeDelegate::PrimitiveArray { length, .. } => format!(
+                "{0}.init(): super({1}({length}));",
+                self.dart_api_type(),
+                self.get_delegate().dart_api_type()
+            ),
+            _ => panic!("Unexpected type"),
+        }
+    }
```
fzyzcjy, line 137: Hmm not very sure - should we put several methods to IrTypeDelegate, but indeed it is only reasonable for GeneralArray and PrimitiveArray, while always throw for others

What about this:

```rust
struct IrTypeDelegate {
  Array(IrTypeDelegateArray),
  ... all old things...
}

struct IrTypeDelegateArray {
    GeneralArray {
        length: usize,
        general: Box<IrType>,
    },
    PrimitiveArray {
        length: usize,
        primitive: IrTypePrimitive,
    },
}

impl IrTypeDelegateArray { // NOTE
  pub fn inner_dart_api_type() ... the several methods ...
}
```

fzyzcjy (commented):
Only a few nits and I am ready to merge it :) Well done!

PR #651, opened 2022-08-27
fix "Skipping unresolvable module" by align latest

alanlzhang:
There are warnings "Skipping unresolvable module XXX" when we use the newer rust module system, eg:
```
some_module.rs
some_module/sub_module.rs
```
The cause is that source_graph assume that there must be a `mod.rs` in the module folder, which is not true sometimes.
This PR fix that.

### DIFF
--- a/frb_codegen/src/source_graph.rs
+++ b/frb_codegen/src/source_graph.rs
@@ -282,53 +282,81 @@ impl Module {
                             child_module
                         }
                         None => {
-                            let folder_path =
-                                self.file_path.parent().unwrap().join(ident.to_string());
-                            let folder_exists = folder_path.exists();
+                            fn get_module_file_path(
+                                module_name: String,
+                                parent_module_file_path: &PathBuf,
+                            ) -> Result<PathBuf, Vec<PathBuf>> {
+                                let folder_path = parent_module_file_path.parent().unwrap();
+                                let file_path = folder_path.join(&module_name).with_extension("rs");
+                                let mut tried = Vec::new();
 
-                            let file_path = if folder_exists {
-                                folder_path.join("mod.rs")
-                            } else {
-                                self.file_path
-                                    .parent()
-                                    .unwrap()
-                                    .join(ident.to_string() + ".rs")
-                            };
+                                if file_path.exists() {
+                                    return Ok(file_path);
+                                }
 
-                            let file_exists = file_path.exists();
+                                tried.push(file_path);
+                                let file_path = folder_path.join(&module_name).join("mod.rs");
+                                if file_path.exists() {
+                                    return Ok(file_path);
+                                }
 
-                            if !file_exists {
-                                warn!(
-                                    "Skipping unresolvable module {} (tried {})",
-                                    &ident,
-                                    file_path.to_string_lossy()
-                                );
-                                continue;
+                                tried.push(file_path);
+                                let folder_path = parent_module_file_path.with_extension("");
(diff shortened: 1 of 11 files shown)


frb_codegen/src/source_graph.rs
```diff
@@ -282,53 +287,81 @@ impl Module {
...
-                                folder_path.join("mod.rs")
-                            } else {
-                                self.file_path
-                                    .parent()
-                                    .unwrap()
-                                    .join(ident.to_string() + ".rs")
-                            };
-
-                            let file_exists = file_path.exists();
+                            fn get_module_file_path(
```
fzyzcjy, line 290: uber-nit: shall we extract it to a first-level function

frb_codegen/src/source_graph.rs
```diff
@@ -282,53 +287,81 @@ impl Module {
...
+                                }
 
-                            let mut child_module = Module {
-                                visibility: syn_vis_to_visibility(&item_mod.vis),
-                                file_path,
-                                module_path,
-                                source,
-                                scope: None,
-                            };
+                                tried.push(file_path);
```
fzyzcjy, line 315: Well, there are some duplicates here. Is it possible we do sth like:

```
let tries = vec![
 || construct-first-try(),
 || construct-second-try(),
 ...
];

// then, use the `tries` vec

File: frb_codegen/src/source_graph.rs [hunk 3]
```diff
@@ -282,53 +287,81 @@ impl Module {
...
+                                            .map(|it| it.to_string_lossy().to_string())
+                                            .fold(String::new(), |mut a, b| {
+                                                a.push_str(&b);
+                                                a.push(',');
+                                                a
+                                            })
+                                    );
+                                    continue;
+                                }
+                            }
```
fzyzcjy, line 364: There seems to be not tests yet, so we cannot verify the behavior. What about we add files like `some_module.rs`, `some_module/sub_module.rs`, and a function that uses the structs in those modules in `api.rs` and a verify test in `.dart`. Then we are pretty sure this PR works :)

fzyzcjy (commented):
Good job, just a few nits :)

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -779,3 +781,10 @@ pub fn multiply_by_ten(measure: Measure) -> Option<Measure> {
         },
     }
 }
+
+pub fn call_both_module_system() -> String {
+    let n = use_new_module_system(1);
```
fzyzcjy, line 786: Seems that we are not using the structs in the old/new module systems? If I understand correctly, this PR tries to allow frb to *find* the structs.

Thus, what about something like `call_both_module_system(a: StructInNewModuleSystem, b: StructInOldModuleSystem)`

fzyzcjy (commented):
Good job! One minor step before merging

frb_example/pure_dart/rust/src/api.rs
alanlzhang: Fix that. But I found there's one restriction: `sub_module` must be `pub` currently, because generated code does not consider `pub use`. But that's another issue. ^_^

fzyzcjy (approved):
LGTM, wait for CI to pass

alanlzhang:
Failed on iPad, but I think it's a CI environment problem.

fzyzcjy:
Sure, I think so

fzyzcjy:
Good job!

PR #666, opened 2022-09-05
Chore/add crate version to generated code header

### LINKED ISSUE #662: Add version information into generated file header
For example, currently it is:

```
// AUTO GENERATED FILE, DO NOT EDIT.
// Generated by `flutter_rust_bridge`.
```

We can auto generate something like:

```
// AUTO GENERATED FILE, DO NOT EDIT.
// Generated by `flutter_rust_bridge` @ 1.40.0
```

Thus, it is easier to know what version has been used to generate the file.

Roms1383:
meme-low

Fixes #662 ^^

## Checklist

- [x] CI is passing.


frb_codegen/src/generator/dart/mod.rs
```diff
@@ -297,11 +297,10 @@ fn generate_dart_implementation_code(
 
 fn generate_file_prelude() -> DartBasicCode {
     DartBasicCode {
-        import: format!("{}
+        import: format!("{CODE_HEADER} @ {CARGO_PKG_VERSION}
```
fzyzcjy, line 300: uber-nit: What about we put pkg version *into* the CODE_HEADER variable, otherwise we are having something like `Generated by `flutter_rust_bridge`. @ 1.42.0` which has a period there

frb_codegen/src/others.rs
```diff
@@ -35,6 +35,8 @@ lazy_static! {
         vec!["store_dart_post_cobject".to_string()];
 }
 
+pub const CARGO_PKG_VERSION: &str = env!("CARGO_PKG_VERSION");
+
 pub const CODE_HEADER: &str = "// AUTO GENERATED FILE, DO NOT EDIT.
```
fzyzcjy, line 40: e.g. 

```
pub const CODE_HEADER: &str = format!("// AUTO GENERATED FILE, DO NOT EDIT.
// Generated by `flutter_rust_bridge` @ {}.", env!("CARGO_PKG_VERSION"));

**Review by fzyzcjy — COMMENTED** [review]
No worries, small PR are also helpful PRs :)

File: frb_codegen/src/others.rs (hunk 2 above)
Comment (Roms1383) [reply frb_codegen/src/others.rs:40]: Thing is `format!` is not a `const` function so it raises a compile-time error. I'm gonna search for a way that doesn't require to add a dependency like e.g. [const_format](https://docs.rs/const_format/latest/const_format/).

Comment (fzyzcjy) [reply frb_codegen/src/others.rs:40]: Then what about

```
pub fn code_header() -> String { ... }

Roms1383: done in 9da58a7be169b7d669486b892d4a4d472324264a

frb_codegen/src/generator/dart/mod.rs
Roms1383: done in 9da58a7be169b7d669486b892d4a4d472324264a

fzyzcjy (approved):

frb_codegen/src/others.rs
Roms1383: I switched CODE_HEADER visibility to `private` so that people are more inclined to use `fn code_header()` instead.

frb_codegen/src/others.rs
```diff
@@ -35,8 +35,12 @@ lazy_static! {
         vec!["store_dart_post_cobject".to_string()];
 }
 
-pub const CODE_HEADER: &str = "// AUTO GENERATED FILE, DO NOT EDIT.
-// Generated by `flutter_rust_bridge`.";
+const CODE_HEADER: &str = "// AUTO GENERATED FILE, DO NOT EDIT.
```
fzyzcjy, line 38: Sure. Indeed I prefer something like:

```
const CODE_HEADER_FORMAT: &str = "// AUTO GENERATED FILE, DO NOT EDIT.
// Generated by `flutter_rust_bridge`@ {}.";

pub fn code_header() -> String {
    format!(CODE_HEADER_FORMAT, env!("CARGO_PKG_VERSION"))
}
```

So that we can freely place format anywhere. But anyway this is just nits

fzyzcjy:
Maybe that's not a problem? https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_example/pure_dart/rust/Cargo.toml has its version 1.0.0 and use build.rs https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_example/pure_dart/rust/build.rs but your code succeeds

frb_codegen/src/others.rs
Roms1383: Screen Shot 2565-09-05 at 15 11 15
looks like it needs to be explicit, what do you think ?

Roms1383:
Ah yes maybe I overthink it, great then !

frb_codegen/src/others.rs
Roms1383: I wondered, so I also tried with raw string literals, to no avail:

Screen Shot 2565-09-05 at 15 15 48

Screen Shot 2565-09-05 at 15 15 29

fzyzcjy: Ah you are right. This is because `format!` is a macro not a function. Then what about directly inline it?

```
pub fn code_header() -> String {
    format!("// AUTO GENERATED FILE, DO NOT EDIT.
// Generated by `flutter_rust_bridge`@ {}.", 
      env!("CARGO_PKG_VERSION")
  )
}

Roms1383: indeed, done in 11240c12b1c53138931112a69813e477f1c0881f 👌

Roms1383:
@fzyzcjy looks like we're good to go!

PR #672, opened 2022-09-06
Feat/chrono

Roms1383:
Relate to the discussion #667.

## Checklist

- [ ] CI is passing.


Roms1383:
Ok, I'm bad I couldn't resist giving it a try but I'll need some guidance here.
Let me write in a separate message the obstacles where I'm still confused.

fzyzcjy:
Sure, take your time and feel free to discuss :)

frb_codegen/src/ir/ty_datetime.rs
```diff
@@ -0,0 +1,43 @@
+use crate::ir::*;
+
+#[derive(Debug, Clone)]
+pub struct IrTypeDateTime(i64);
```
fzyzcjy, line 4: No, no need for ty_datetime.rs.

Just look at ty_delegate.rs, and mimic how we deal with `String`

frb_example/pure_dart/dart/lib/bridge_generated.dart
```diff
@@ -504,6 +508,14 @@ class ExoticOptionals {
   });
 }
 
+class FeatureChrono {
+  final int date;
+
+  FeatureChrono({
+    required this.date,
+  });
+}
```
Roms1383, line 517: Where should I make any change to have this instead ?

```dart
class FeatureChrono {
  final DateTime date;

  FeatureChrono({
  required this.date,
  });
}
```

Roms1383 (commented):
Here's the first step where I could use some guidance.

frb_codegen/src/ir/ty_datetime.rs
Roms1383: The thing is, with the usual delegate approach, one will e.g. convert to a primitive in each of FFI step (dart wire type / dart class type) as far as I understand. What I'd like to do is to have dart wire type be a `int` / `Int64` while the final dart class type has a proper dart `DateTime`.

Roms1383: My idea is to have something like:
- final Rust type be a `DateTime<Utc>`
- Rust intermediate wire type be a `i64`
- Dart intermediate wire type be a `int` / `Int64`
- final Dart type be a `DateTime`

fzyzcjy: > have dart wire type be a int / Int64 while the final dart class type has a proper dart DateTime.

I guess `String` -> `Vec<u8>` is just the same.

We have Dart String and Rust String for end users, while internally it utilizes the `Vec<u8>` do to the heavy job.

So, mimic the scenario:

We have Dart *DateTime* and Rust *chrono datetime* for end users, while internally it utilizes the *int* do to the heavy job.

fzyzcjy: https://github.com/fzyzcjy/flutter_rust_bridge/blob/50064a9563e63d05e4b22d2c31b35ff943f7da87/frb_codegen/src/ir/ty_delegate.rs#L6

Maybe the first step is, just add a sibling to the "String" enum variant here, say, "DateTime". And mimic how "IrTypeDelegate::String" deal with the thing.

Roms1383: So the flow could be like, e.g.:
### from Rust to Dart
1. Rust has a `DateTime<Utc>`
2. it turns it into a wire type `i64` by doing e.g. `self.timestamp_micros()`
3. send it over the wire
4. Dart receives a wire type `int` / `Int64`
5. Dart finally turns it into a `DateTime` doing e.g. `DateTime.fromMicrosecondsSinceEpoch(value)`

### from Dart to Rust
1. Dart has a `DateTime`
2. it turns it into a wire type `int` / `Int64` by doing e.g. `self. microsecondsSinceEpoch`
3. send it over the wire
4. Rust receives a wire type `i64`
5. Rust finally turns it into a `DateTime<Utc>` by doing e.g. `DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(value / 1_000_000 as i64, 0u32), Utc)`

fzyzcjy: Roughly speaking yes, have a look at http://cjycode.com/flutter_rust_bridge/contributing/design.html for details

Roms1383: So, thing is, I don't understand for example why the latest commit 3d482f8113310a770757d0ebc2d63a6bdf9badd8 didn't trigger any change in the `bridge_*` files. I'm missing something because the code supposed to be generated doesn't appear anywhere.

Roms1383: Yes, to be honest I'm completely confused even following [design](http://cjycode.com/flutter_rust_bridge/contributing/design.html) architecture.
It seems that some parts get conditionally generated from different sources based on parsing, wrapper, ir, etc.
Which makes it difficult to find where one piece of code originates from.

fzyzcjy: > So, thing is, I don't understand for example why the latest commit https://github.com/fzyzcjy/flutter_rust_bridge/commit/3d482f8113310a770757d0ebc2d63a6bdf9badd8 didn't trigger any change in the bridge_* files. I'm missing something because the code supposed to be generated doesn't appear anywhere.

Seems you forget to modify the parser. Thus, no field is of type IrDelegate::DateTime

Roms1383: My understanding was that `SyncReturn` is a completely different type solely for synchronous functions, is it?

fzyzcjy: Yes. Ignore it since you are dealing with async things

Roms1383: But then I already added a condition at the bottom for DateTime<Utc> in frb_codegen/src/parser/ty.rs line 239, so?

fzyzcjy: What about some classical debugging skills: For example, use `println!` to see what is happening. I suspect that string may not `DateTime` but `DateTime<Utc>` or something like that (anyway it is just one of the infinitely many possibility of bugs) - so just debug it :)

Roms1383: @fzyzcjy you're utterly right, sorry I thought I'd manage to get away without the tedious debugging part lol

Anyway if you don't mind I'd like to start this whole PR from scratch again since working directly on `1.45` is way easier than trying to resolve merge conflicts here and potentially getting any unwanted regression.

So, if that's ok with you, I'm gonna close this one, create a new branch and open a brand new PR over there, while mentioning this one as a reminder.

fzyzcjy: Sure!

PR #677, opened 2022-09-07
Cache Rust files in CI

Desdaemon:
In CI, `cargo install` is called by the iOS and Android workflows which is a major bottleneck for CI. This PR caches the installed binaries along with files in select `target` directories.


fzyzcjy (approved):
LGTM once ci passes

fzyzcjy:
No worries about the run codegen failure. I am fixing it in #678

fzyzcjy:
Btw, I am considering sending a post in r/Rust and r/FlutterDev, and maybe hackernews? (never did the last before) What do you think?

Possible titles (still VERY rough draft):

> GUI.platforms += 1: Web is supported now, besides phones & desktops [for r/Rust]
> Using Rust (WASM) on Web now, besides phones & desktops, enpowering reliablility and performance [for r/FlutterDev]

Desdaemon:
> Btw, I am considering sending a post in r/Rust and r/FlutterDev, and maybe hackernews?

If it's HN, maybe they will appreciate the "Show HN" post format, but otherwise great idea!

fzyzcjy:
> If it's HN, maybe they will appreciate the "Show HN" post format, but otherwise great idea!

Good idea, I will do that :)

fzyzcjy:
Will release new version around 18hrs later, when posts will be sent roughly then :)

PR #720, opened 2022-09-18
Fix: WireSyncReturnStruct should be freed after buffer being consumed

hsfzxjy:
In `frb_dart/lib/src/basic.dart`, the method `free_WireSyncReturnStruct` should be invoked **after** the buffer being consumed (either parsed as success data or as error string), in case of accessing deallocated memory.

### DIFF
--- a/frb_dart/lib/src/basic.dart
+++ b/frb_dart/lib/src/basic.dart
@@ -58,13 +58,14 @@ abstract class FlutterRustBridgeBase<T extends FlutterRustBridgeWireBase> {
     } catch (err, st) {
       throw FfiException('EXECUTE_SYNC_ABORT', '$err', st);
     }
-    final bytes = raw.buffer;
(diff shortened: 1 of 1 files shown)


fzyzcjy:
I think you are right! Really not sure why this has been there for a long time without problems...

By the way, how did you find out this bug?

I will fix master branch CI first

hsfzxjy:
> By the way, how did you find out this bug?

I was working on a similar project but for Dart and Golang, and was seeking for some design inspiration from frb. So I took a review on the code where I suddenly spotted the bug. 😂

fzyzcjy:
Aha, nice catch!

fzyzcjy:
Maybe need to merge from master s.t. CI can pass

hsfzxjy:
Sure. I've rebase the commit onto the front of latest master.

fzyzcjy:
@all-contributors please add @hsfzxjy for code

PR #722, opened 2022-09-19
Add i18n support

trdthg:
## Add gettext and i18n support
This is still a draft now, I want to discuss some questions here.
I have tried to use code from https://github.com/rust-lang/mdBook/pull/1864 and https://github.com/rust-lang/mdBook/pull/1306, and found a bug there. It was [fixed](https://github.com/rust-lang/mdBook/pull/1864#issuecomment-1250385916) quickly. This translation process is feasible, but both of them are not merged yet, so  configuring the environment is a little complicated.


fzyzcjy:
> This is still a draft now

Done

> This translation process is feasible, but both of them are not merged yet, so configuring the environment is a little complicated.

Hmm good question. Is it possible that, you firstly configure everything under your fork?

fzyzcjy:
@all-contributors please add @trdthg for doc

trdthg:
> Is it possible that, you firstly configure everything under your fork?

Yes, It's not really that complicated, except that you have to clone two other copies of the mdBook.
And the config them [here](https://github.com/fzyzcjy/flutter_rust_bridge/pull/722/commits/3077c95d8e355c9074a58408dc5e3162a7aac3dc).
```sh
# [Add commands for Gettext-based translations i18n #1864](https://github.com/rust-lang/mdBook/pull/1864)
# extract and generate pot/po file
MDBOOK1="path to #1864 mdbook"

# [Add option to localize books in multiple languages #1306](https://github.com/rust-lang/mdBook/pull/1306)
# add multi language support
MDBOOK2="path to #1306 mdbook"
```

fzyzcjy:
Great!

Btw is the translatetool written by you? Maybe we can merge it into `justfile`, which has excellent support for commands, so no need for checkArgNumber etc

trdthg:
> Btw is the translatetool written by you?

Yes.
> Maybe we can merge it into justfile,

Good idea. You can do some more detailed check.

trdthg:
Here is the current progress.

I moved scripts to `justfile` and update CI workflow. After the rust 1.64 update, pr 1306 is broken, so I folked and fixed that problem. Now `just` and CI works.

But I don't really like this solution, and I found https://github.com/rust-lang/mdBook/issues/5#issuecomment-1258922986 today, which seems to be better. Should I switch?

fzyzcjy:
> But I don't really like this solution, and I found https://github.com/rust-lang/mdBook/issues/5#issuecomment-1258922986 today, which seems to be better. Should I switch?

Depend on you. When doing https://github.com/fzyzcjy/flutter_smooth, I tried http://cjycode.com/flutter_smooth/, which uses https://docusaurus.io/. That one seems to have excellent multi-lang support.

So you can even switch the whole book to docusaurus as well :)

trdthg:
Looks good, I'll give it a try.

fzyzcjy:
Btw @Desdaemon what do you think about the doc site? (Since you have written several chapters)

Desdaemon:
> Btw @Desdaemon what do you think about the doc site? (Since you have written several chapters)

It looks great! I might contribute some translations myself if it works out well.

fzyzcjy:
👀

trdthg:
> eyes

Busy with other things, but maybe next week will be free to continue 😵

fzyzcjy:
take your time!

mgeisler:
Hi @trdthg, I found this PR linked from https://github.com/rust-lang/mdBook/issues/5 :smile:

I see you've been using the code I wrote in https://github.com/rust-lang/mdBook/pull/1864 — I just wanted to let you know that I've updated the code to be much easier to use. You no longer need to modify mdbook itself, you just use a preprocessor (to translate) and a renderer (to extract strings). Please see https://github.com/google/comprehensive-rust/pull/130.

Feel free to drop me a comment on that PR if you try it out!

fzyzcjy:
👀

mgeisler:
The code in question is now released as a stand-alone crate: https://crates.io/crates/mdbook-i18n-helpers. This should let you simplify the PR here a great deal :smile:

trdthg:
@mgeisler, Thanks for the reminder, hope my reply is not too late

`Comprehensive-rust` already made a good example, my solution (with modified `mdbook-i18n-helpers` + `mdbook-i18n`) seems to add extra complexity.

@fzyzcjy I think we can use the same way as it, what do you think？

fzyzcjy:
@trdthg You are feel to use any method that is simple and easy to maintain :)

.github/workflows/setup-rust-cache/action.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+description: Configure the rust-cache workflow.
+
+runs:
+  using: composite
+  steps:
+    - name: Setup Rust cache
+      uses: Swatinem/rust-cache@v2
+      with:
+         prefix-key: v1  # Remove after merging #1322.
```
mgeisler, line 11: You should remove the key prefix here since you start with no caches.

book/po/messages.pot
mgeisler, line 1: This file is an intermediate file and doesn't need to be committed (it can be reproduced on demand).

trdthg: Removed

trdthg:
@mgeisler Both deleted

fzyzcjy:
image

So, I briefly checked it: https://trdthg.github.io/flutter_rust_bridge/zh/quickstart.html The "zh" mode still only contains English. Maybe I am checking the wrong location?

trdthg:
> The "zh" mode still only contains English.

@fzyzcjy Yes, I only translate "Introduce" in sidebar to "介绍" for now. More translations will be added gradually in other PRs.

fzyzcjy:
I see. So this PR adds the necessary *code* instead of translation content if I understand correctly.

fzyzcjy:
Seems the CI fails. Could you please check it a bit?

image

book/theme/index.hbs
```diff
@@ -0,0 +1,384 @@
...
+                            <i id="git-repository-button" class="fa {{git_repository_icon}}"></i>
+                        </a>
+                        {{/if}}
+                        {{#if git_repository_edit_url}}
+                        {{#if (eq language "en")}}
+                        <a href="{{git_repository_edit_url}}" title="Suggest an edit" aria-label="Suggest an edit">
+                            <i id="git-edit-button" class="fa fa-edit"></i>
+                        </a>
+                        {{else}}
+                        <a href="https://github.com/google/comprehensive-rust/edit/main/po/{{language}}.po"
```
fzyzcjy, line 220: nit: looks like this link needs to be changed

fzyzcjy:
Good job! I will check soon

fzyzcjy:
Firstly, seems that this is not deployed successfully?

image

.github/workflows/gh-pages.yaml
```diff
@@ -11,24 +11,40 @@ on:
     paths:
       - 'book/**'
 
+env:
+  CARGO_TERM_COLOR: always
+  # Update the language picker in index.hbs to link new languages.
+  LANGUAGES: zh
```
fzyzcjy, line 17: not very sure, but seems this env is set to the *whole* gh-pages. So will this cause the whole book to be zh instead of en?

.github/workflows/setup-rust-cache/action.yaml
```diff
@@ -0,0 +1,9 @@
+name: Setup Rust cache
```
fzyzcjy, line 1: nit: if this sub-workflow is as simple as just one step of two lines, maybe we should inline it? What do you think about it

.github/workflows/install-mdbook/action.yaml
```diff
@@ -0,0 +1,18 @@
+name: Install mdbook and dependencies
+
+description: Install mdbook with the dependencies we need.
+
+runs:
```
fzyzcjy, line 5: (same)

book/theme/index.hbs
```diff
@@ -0,0 +1,384 @@
+<!DOCTYPE HTML>
```
fzyzcjy, line 1: nit: Is it copied (and modified) from some other repository? If so, it seems good to add a link to the original place to honor the original author and to make future readers clear

book/theme/index.hbs
```diff
@@ -0,0 +1,384 @@
...
+        <title>{{ title }}</title>
+        {{#if is_print }}
+        <meta name="robots" content="noindex" />
+        {{/if}}
+        {{#if base_url}}
+        <base href="{{ base_url }}">
+        {{/if}}
+
+        <script async src="https://www.gstatic.com/brandstudio/kato/cookie_choice_component/cookie_consent_bar.v3.js"
+                data-autoload-cookie-consent-bar="true"></script>
```
fzyzcjy, line 15: hmm seems we do not need to use cookies - we never track anything!

book/theme/index.hbs
```diff
@@ -0,0 +1,384 @@
...
+
+        <script async src="https://www.googletagmanager.com/gtag/js?id=G-ZN78TEJMRW"></script>
+        <script>
+          window.dataLayer = window.dataLayer || [];
+          function gtag(){dataLayer.push(arguments);}
+          gtag('js', new Date());
+          gtag('config', 'G-ZN78TEJMRW');
+        </script>
```
fzyzcjy, line 23: Hey we do NOT track users! Seems that you copied the file but forgot to delete? ;)

book/theme/index.hbs
```diff
@@ -0,0 +1,384 @@
...
+
+        <!-- Custom HTML head -->
+        {{> head}}
+
+        <meta name="description" content="{{ description }}">
```
fzyzcjy, line 28: TODO(fzyzcjy): Go to deployed site and check network requests, e.g. check whether there is some weird network requests or 404 requests etc

book/theme/index.hbs
```diff
@@ -0,0 +1,384 @@
...
+                (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+                (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+                m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+                })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+                ga('create', '{{google_analytics}}', 'auto');
+                ga('send', 'pageview');
+            }
+        </script>
+        {{/if}}
```
fzyzcjy, line 328: (same)

fzyzcjy (commented):
Good jobs! Mainly nits

book/theme/index.hbs
mgeisler: That bit is from the upstream `mdbook` project: they have support for setting up Google Analytics (using an old-school way): https://github.com/rust-lang/mdBook/blob/master/src/theme/index.hbs#L244-L260. It's of course off by default.

book/theme/index.hbs
mgeisler: Yeah, please don't use the Google Analytics from my book — that'll be confusing :smile:

I think it would be easier to start with https://github.com/rust-lang/mdBook/blob/master/src/theme/index.hbs and then take the parts of https://github.com/google/comprehensive-rust/commits/main/theme/index.hbs that you need/want.

book/theme/index.hbs
mgeisler: Yeah, don't bother with that — I had to add it to my site because we use Google Analytics there.

.github/workflows/gh-pages.yaml
mgeisler: It use it in [Comprehensive Rust](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/publish.yml#L22) to maintain a list of languages for which we should publish translations. The variable name is not significant for `mdbook` or anything else — I just added it so that I can decouple having translations in the repository from actually publishing them on our site.

.github/workflows/setup-rust-cache/action.yaml
trdthg: agree, inline is clear enough.

.github/workflows/install-mdbook/action.yaml
trdthg: inlined

book/theme/index.hbs
trdthg: as mgeisler mentioned, this file comes from the mdbook itself, how about the following:
```html
{{!-- Modified from https://github.com/rust-lang/mdBook/blob/master/src/theme/index.hbs --}}
 ```

File: book/theme/index.hbs (hunk 7 above)
Comment (trdthg) [reply book/theme/index.hbs]: deleted (didn't check carefully before, thank you for your patience)

File: book/theme/index.hbs (hunk 8 above)
Comment (trdthg) [reply book/theme/index.hbs]: deleted

File: .github/workflows/gh-pages.yaml [hunk 11]
```diff
@@ -11,24 +11,49 @@ on:
...
+
 jobs:
   deploy:
-    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
fzyzcjy: Hmm... I see:

image

Does it mean, after merging, the master branch will fail to deploy?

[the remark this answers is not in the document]
trdthg: I didn't get it. where did you see it? Is it something I can't access?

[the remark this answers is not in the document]
fzyzcjy: I see - probably because I am using Refined GitHub (an open source plugin)

image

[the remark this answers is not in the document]
fzyzcjy: > My understanding is that the previous deployment process was broken, and after I modified gh-pages.yaml, the deployment process should have worked, but since I didn't modify the contents of the book folder, it didn't trigger the automatic deployment.

Then maybe trigger the deployment s.t. we are 100% sure ;)
e.g. commit to change one newline; then commit to change it back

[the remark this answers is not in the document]
trdthg: > I see - probably because I am using Refined GitHub (an open source plugin)

doesn't work for me, but I think it refers to this page https://github.com/fzyzcjy/flutter_rust_bridge/deployments

fzyzcjy:
Btw, it seems the zh homepage is somehow broken:

image

book/theme/index.hbs
fzyzcjy: looks ok

fzyzcjy (approved):
Good job! I am going to merge it, maybe after the fix of broken homepage if it is easy to fix

trdthg:
> Btw, it seems the zh homepage is somehow broken:

@fzyzcjy seems the reason comes from mdbook-i18n-helper, after I install it from source, then build and serve locally. it's fixed
image

@mgeisler I saw that compensive-rust also has this little problem, maybe you need to release a new version?

.github/workflows/gh-pages.yaml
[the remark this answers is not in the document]
fzyzcjy: Continuing this thread: @trdthg - what about adding this information at the top of each `zh` page, saying something like:

```
This page is translated by [@trdthg](https://github.com/trdthg)
```

.github/workflows/gh-pages.yaml
```diff
@@ -11,24 +11,52 @@ on:
...
+      - name: Build all translations
+        working-directory: ./book
+        run: |
+          for po_lang in ${{ env.LANGUAGES }}; do
+              echo "::group::Building $po_lang translation"
+              MDBOOK_BOOK__LANGUAGE=$po_lang \
+              MDBOOK_OUTPUT__HTML__SITE_URL=/flutter-rust-bridge/$po_lang/ \
+              mdbook build -d book/$po_lang
+              echo "::endgroup::"
+          done
```
fzyzcjy, line 59: nit: it seems good to move them to `justfile`. Then, everyone who wants to play with it locally (including you and me ;) ) can execute a one-line command like `just whatever-command-name` instead of copy-and-paste these long code

fzyzcjy (approved):

.github/workflows/gh-pages.yaml
trdthg: > what about adding this information at the top of each zh page

Not bad, hat about the people who contribute to the same page afterward?
But in any case, with the current translation process, it seems not so easy to add additional information to the translated page

trdthg: > it seems good to move them to justfile

agree, one thing to note is that if you want to serve all languages, you will have to first build and then use a separate http server to serve them, such as https://github.com/TheWaWaR/simple-http-server

fzyzcjy: > Not bad, what about the people who contribute to the same page afterward?

Good question. Maybe just list all and add a link pointing to a separate page showing everyone's detailed contribution?

> But in any case, with the current translation process, it seems not so easy to add additional information to the translated page

Just a wild guess, Is it possible to modify the HTML?

> agree, one thing to note is that if you want to serve all languages, you will have to first build and then use a separate http server to serve them, such as [TheWaWaR/simple-http-server](https://github.com/TheWaWaR/simple-http-server)

I see. Then it would be great to have a bit of doc, telling users how to build and serve the doc locally

mgeisler:
> > Btw, it seems the zh homepage is somehow broken:
> 
> @fzyzcjy seems the reason comes from mdbook-i18n-helper, after I install it from source, then build and serve locally. it's fixed image
> 
> @mgeisler I saw that compensive-rust also has this little problem, maybe you need to release a new version?

Hey @fzyzcjy! Yes, I need to release a new version to get this fixed. Thanks for the ping!

I didn't do this yet since I have been working on a way to improve the text extraction. The current `main` branch has this new system: it will turn `# Some heading` in the Markdown file into just `"Some heading"` in the PO file. Similarly, tables are now extracted on a cell-by-cell basis. This removes most of the Markdown syntax and makes the job easier for translators.

For this to work smoothly with existing translations, I've written a conversion tool: https://github.com/google/mdbook-i18n-helpers/pull/46. When that is merged, I will make a new release, upgrade our own PO files, and also update the publishing pipeline to use the new version.

mgeisler:
Hi @fzyzcjy, I've finally released a version 0.2 of mdbook-i18n-helpers — which improves the way text is extracted and which fixed the rendering of the badges on the translated pages: https://google.github.io/comprehensive-rust/pt-BR.

I know this PR is closed, but if you ever end up using the plugins, please add a link to your project in the mdbook-i18n-helpers README.

fzyzcjy:
@mgeisler Sure! I am happy to see i18n support, this PR is closed by stale bot.

PR #726, opened 2022-09-21
Make threadpool size dynamic

### LINKED ISSUE #725: Make the `WORKER_COUNT` configurable
Currently, the `thread::WORKER_COUNT` is 4. If we already have 4 ffi calls that are not finished at once, we can't make another ffi call immediately.

TENX-S:
Fixes #725 

## Checklist

- [ ] CI is passing.

### DIFF
--- a/frb_rust/src/macros.rs
+++ b/frb_rust/src/macros.rs
@@ -9,7 +9,11 @@
 macro_rules! transfer {
     (|| $block:block) => {{
         #[cfg(not(target_family = "wasm"))]
-        { move || $block }
+        {
+            move || {
+                $block
+            }
+        }
 
         #[cfg(target_family = "wasm")]
         {
@@ -83,7 +87,16 @@ macro_rules! spawn {
         let worker = $crate::transfer!($($tt)*);
         #[cfg(not(target_family = "wasm"))]
         {
-            $crate::thread::THREAD_POOL.lock().execute(worker)
+            let current_max = $crate::thread::THREAD_POOL.lock().max_count();
+            if $crate::thread::THREAD_POOL.lock().active_count() + 1 >= current_max {
+                $crate::thread::THREAD_POOL.lock().set_num_threads(current_max + 4);
+            }
+            $crate::thread::THREAD_POOL.lock().execute(worker);
+            let current_active = $crate::thread::THREAD_POOL.lock().active_count();
+            let current_max = $crate::thread::THREAD_POOL.lock().max_count();
+            if current_max > 4 && current_max > current_active * 2 {
+                $crate::thread::THREAD_POOL.lock().set_num_threads(current_active + 4);
+            }
         }
 
         #[cfg(target_family = "wasm")]


TENX-S:
I want to write an example for this. Can I make it into `frb_exmaple/with_flutter` ?

frb_rust/src/macros.rs
```diff
@@ -83,7 +87,16 @@ macro_rules! spawn {
...
+            let current_max = $crate::thread::THREAD_POOL.lock().max_count();
+            if $crate::thread::THREAD_POOL.lock().active_count() + 1 >= current_max {
+                $crate::thread::THREAD_POOL.lock().set_num_threads(current_max + 4);
+            }
+            $crate::thread::THREAD_POOL.lock().execute(worker);
+            let current_active = $crate::thread::THREAD_POOL.lock().active_count();
+            let current_max = $crate::thread::THREAD_POOL.lock().max_count();
+            if current_max > 4 && current_max > current_active * 2 {
+                $crate::thread::THREAD_POOL.lock().set_num_threads(current_active + 4);
+            }
```
fzyzcjy, line 99: Well, I am hesitate whether this is a good idea. seems that we are going to expand the thread pool (and no shrinking). That sounds very expensive...

fzyzcjy:
Indeed I suggest implement the async support, then you can call a lot.

frb_rust/src/macros.rs
TENX-S: The shrinking is below the `$crate::thread::THREAD_POOL.lock().execute(worker);`

fzyzcjy: Oh yes. Then it is even more expensive - we need to pay for expanding and shrinking :/

TENX-S: Yeah, kind of. But it doesn't happen a lot, I think

fzyzcjy: Hope so :) Indeed it reminds me of the dynamic sized thread pool in java (which causes a lot of trouble though..)

TENX-S: https://github.com/rust-threadpool/rust-threadpool/blob/a9dd15c278a8ab07d694efb169348138823bb89d/src/lib.rs#L576-L581

I think shrinking is not going to spawn any threads. And if it needs to expand the pool, it will spawn 4 threads at one time. I will investigate more about the resizing strategy if you are ok with this way.

fzyzcjy: Sure. I mean, IIRC thread creation and deletion are expensive by nature, especially if you are going to call frequently

I do not have much brain power now, so hopefully my comments are correct

fzyzcjy: Indeed I suggest implement the async support, then you can call a lot.

fzyzcjy: What is the nature of your rust code?

* If they are all heavy computation, then 4 threads (or #thread == #cores) is good - having more threads will not make it run faster
* If they are network or disk r/w, i.e. they are I/O, then using async is the right approach. Use a large thread pool is expensive for that.

TENX-S: For async context, we can send the async ffi call to the async runtime (tokio or async-std), which manage their threadpool internally IIRC. But we still need to handle any number of our sync ffi calls. I believe that will need the threadpool to have a dynamic size.

TENX-S: > * If they are all heavy computation, then 4 threads (or #thread == #cores) is good - having more threads will not make it run faster

> * If they are network or disk r/w, i.e. they are I/O, then using async is the right approach. Use a large thread pool is expensive for that.

For my own usage, async api is the best choice, it stops me wrestling with transform async to sync. But it's seem a little scary to implement😂

fzyzcjy: Not scary! I am always willing to help - you see I am quite active on github. And there are already so many contributors to this lib, it is simple :)

TENX-S: Alright, I will try to investigate async support next month as I have 7 days off 🤣

fzyzcjy: Take your time and looking forward to your PR!

PR #728, opened 2022-09-22
Feat/UUID

Roms1383:
Oops I did it again. This PR aims at supporting commonly used [uuid crate](https://docs.rs/uuid/1.1.2/uuid/index.html) type.

## Checklist

- [x] CI is passing.

### Remark for maintainer

So basically there's one thing that I haven't settled down yet :

I ran some experimentations with [uuid::Uuid](https://docs.rs/uuid/1.1.2/uuid/struct.Uuid.html) (which boils down to a `[u8;16]`, see [uuid::Bytes](https://docs.rs/uuid/1.1.2/uuid/type.Bytes.html)).

When converted `IntoDart`, I already turn `Uuid` into its :crab: `Vec<u8>` / :arrows_counterclockwise: `DartCObject` / :dart: `Uint8List` IR, so in the case of a `Vec<Uuid>` by default one get a :crab: `Vec<Vec<u8>>` / :arrows_counterclockwise: `Vec<DartCObject>` / :dart: some list of `Uint8List`.

Looking at documentation on types used on both sides (especially `Vec<u8>` and `Uint8List`), there's actually convenient methods to reconstruct it from bytes. So I benchmarked concatenating running `IntoDart` on different sized vecs of `Uuid` in a single `Vec<u8>` vs using default Vec<T> implementation, out of curiosity.

I was actually dubious whether it's overoptimizing or unnecessary, but the results of the benchmark were actually... surprising. See this [workflow run](https://github.com/sunshine-protocol/allo-isolate/actions/runs/3100560866/jobs/5020977112) for example :

|  size | duration (default : nested vecs) | duration (packed : flattened in single vec) |
|------------|-------------|----------------------|
|  100             |  8.7135 µs   |  397.21 ns  |
|  100_000    |  16.226 ms  |  390.64 µs |
|  1000_000  |  175.02 ms  |  5.9609 ms |

If my benchmark is correct, it's actually almost 30 times faster, so it might be worth the effort, what do you think ?
I gave it a try already, but still haven't found how to tweak codegen to have a specific `Vec<T>`.


fzyzcjy:
Looks interesting! So I guess maybe we can firstly have a PR about uuid, and then another PR to do the promising optimization :)

Roms1383:
> Looks interesting! So I guess maybe we can firstly have a PR about uuid, and then another PR to do the promising optimization :)

I agree, thing is it's dependent on the [impl IntoDart in allo-isolate](https://github.com/sunshine-protocol/allo-isolate/pull/30/files#diff-e35e7d29dbacdd107ebd11a53561037168d1a552fba14493d8b449a023da506cR45-R50) so if it gets merged and I open another PR to change it later on, conceptually it's a breaking change (even if realistically speaking, probably nobody would have implemented it in another crate by that time, but still). Hence why I wonder if I should implement it in one-go.

Roms1383:
@fzyzcjy I still have this weird error locally that `pubspec.lock` files are not in sync, giving stuff like this:
Screen Shot 2565-09-22 at 15 07 07

That's not a big deal but it's a bit annoying since I need to pay attention each time I commit lol.

fzyzcjy:
You may need to upgrade to flutter 3.3.1?

Anyway, ignore this problem. Only need to worry when I am going to merge the Pr

Roms1383:
> You may need to upgrade to flutter 3.3.1?

Indeed! My bad 😅

Roms1383:
I'm benchmarking on my machine if converting in native endianness would make any difference, performance-wise, but it seems that it's only a delta of ~2%. Do you know of any machines available in Github CI that would be `big` / `little` endians to check ? @fzyzcjy

fzyzcjy:
github ci may not be suitable for benchmark I guess, they are virtual machine

Roms1383:
Ok then at first sight it might not worth the hassle for such a small delta. Maybe we can just call it a day (well, if CI passes of course) ^^

frb_codegen/src/generator/dart/ty_delegate.rs
```diff
@@ -81,6 +99,18 @@ impl TypeDartGeneratorTrait for TypeDelegateGenerator<'_> {
...
+            #[cfg(feature = "uuid")]
+            IrTypeDelegate::Uuid => "return UuidValue.fromByteList(raw);".to_owned(),
+            #[cfg(feature = "uuid")]
+            IrTypeDelegate::Uuids => "
+            final count = raw.lengthInBytes / 16;
+            var buffer = List<UuidValue>.empty(growable: true);
+            for (var i = 0; i < count; i++) {
+              buffer.add(UuidValue.fromByteList(Uint8List.view(raw.buffer, i * 16, 16)));
+            }
+            return buffer;
```
Roms1383, line 111: I'm such a big noob with Dart (or with every language? lol) but anyway, you guys @fzyzcjy @Desdaemon experts at Dart, I have a feeling that this is a code smell ... any recommendation on how to make it faster / more idiomatic Dart ?

Roms1383 (commented):
Maybe this point can be further optimized ?

frb_codegen/src/generator/dart/ty_delegate.rs
Roms1383: especially the part where I declare `List<UuidValue>.empty` usually that means that the list is not pre-allocated.

Roms1383: apparently they mention about `List.generate`, I'm gonna try that.

Roms1383: done in a9ada6d536ea96a37a0595fea41a2c83b556a06d

fzyzcjy:
> but then I'm missing something in valgrind runs

maybe intro that dep in frb example packages?

Roms1383:
> maybe intro that dep in frb example packages?

No, turns out I'm dumb 😅, there was an error shown way above in the Valgrind logs. Mostly casting issues, I'm solving them.

fzyzcjy:
That's OK

Roms1383:
If you have time, let's review and fix the nits while waiting for the upstream PR to get merged :)

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -203,10 +203,18 @@ fn generate_import_header(imports: HashSet<&IrDartImport>) -> DartBasicCode {
 
 fn generate_common_header() -> DartBasicCode {
     DartBasicCode {
-        import: "import 'dart:convert';
+        import: if cfg!(feature = "uuid") {
+            "import 'dart:convert';
+            import 'dart:async';
+            import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
+            import 'package:uuid/uuid.dart';"
```
fzyzcjy, line 210: nit: would be great if `if cfg { "import uuid" } else { "" }` so less repeat

fzyzcjy (commented):
(not finish review, just start looking at it)

frb_codegen/src/generator/dart/mod.rs
Roms1383: resolved in 32b0f287c1c6502ba8de3db0873a40f2dbeb3209

frb_rust/Cargo.toml
```diff
@@ -16,9 +16,14 @@ threadpool = "1.8.1"
...
 flutter_rust_bridge_macros = { path = "../frb_macros", version = "1" }
+uuid = { version = "1.1.2", optional = true }
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
-allo-isolate = { version = "0.1.14-beta.2", features = ["anyhow", "chrono"] }
+allo-isolate = { git = "https://github.com/Roms1383/allo-isolate.git", branch = "feat/uuid", features = [
+  "anyhow",
+  "chrono",
+  "uuid",
+] }
```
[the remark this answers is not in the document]
Roms1383: resolved in 89270f14194258e43a4d38452caccecd0ec40d7d

fzyzcjy:
I am very busy with https://github.com/flutter/flutter/issues/101227, but will review as soon as possible

.github/workflows/ci.yaml
```diff
@@ -681,7 +681,7 @@ jobs:
       - name: Run codegen to pure_dart example (non-windows)
         if: ${{ matrix.os.family != 'windows' }}
         working-directory: ./frb_codegen
-        run: cargo run --package flutter_rust_bridge_codegen --bin flutter_rust_bridge_codegen --
+        run: cargo run --package flutter_rust_bridge_codegen --bin flutter_rust_bridge_codegen --features "chrono,uuid" --
           --rust-input ../frb_example/pure_dart/rust/src/api.rs
```
fzyzcjy, line 685: btw dart web tests seem to fail (unrelated to this line of code)

frb_codegen/src/generator/dart/ty_delegate.rs
```diff
@@ -81,6 +89,18 @@ impl TypeDartGeneratorTrait for TypeDelegateGenerator<'_> {
...
+            #[cfg(feature = "uuid")]
+            IrTypeDelegate::Uuid => {
+                "return UuidValue.fromByteList(_wire2api_uint_8_list(raw));".to_owned()
+            }
+            #[cfg(feature = "uuid")]
+            IrTypeDelegate::Uuids => "
+            final bytes = _wire2api_uint_8_list(raw);
+            return List<UuidValue>.generate(
+              bytes.lengthInBytes ~/ 16,
+              (int i) => UuidValue.fromByteList(Uint8List.view(bytes.buffer, i * 16, 16)),
```
fzyzcjy, line 101: nit: I wonder will uuid always be 16 bytes

if so, can we define `const uuidSize = 16`

frb_rust/src/ffi/web.rs
```diff
@@ -54,6 +54,28 @@ impl IntoDart for chrono::Duration {
...
+        self.as_bytes().to_vec().into_dart()
+    }
+}
+
+#[cfg(feature = "uuid")]
+impl IntoDart for Vec<uuid::Uuid> {
+    #[inline]
+    fn into_dart(self) -> DartAbi {
+        use std::io::Write;
+        let mut buffer = Vec::<u8>::with_capacity(self.len() * 16);
```
fzyzcjy, line 71: nit: Again the "16" magic number

frb_rust/src/ffi/web.rs
```diff
@@ -54,6 +54,28 @@ impl IntoDart for chrono::Duration {
...
+        for id in self {
+            let _ = buffer.write(id.as_bytes());
```
fzyzcjy, line 73: nit: Not sure can we use `.map` and `.join`

fzyzcjy (commented):
Do not have much brain power now, but the PR looks good to me except for some nits. Good job!

fzyzcjy:
@Desdaemon What do you think? 👀

.github/workflows/ci.yaml
Roms1383: it says timeout so not sure : github ci has an option to rerun only failed jobs.

fzyzcjy: Ok, will not be problem if future commit it is green :)

frb_dart/lib/src/helpers.dart
```diff
@@ -148,3 +148,15 @@ Duration wire2apiDuration(int ts) {
...
   return Duration(microseconds: ts);
 }
+
+Uint8List wire2apiConcatenateBytes(List raw) {
+  var builder = BytesBuilder();
+  if (raw.isEmpty) {
+    return builder.toBytes();
+  }
+  final count = raw.length;
+  for (var i = 0; i < count; i++) {
```
Desdaemon, line 158: nit: can be replaced with for-each loop and ignore the empty check

frb_rust/Cargo.toml
```diff
@@ -16,9 +16,14 @@ threadpool = "1.8.1"
 lazy_static = "1.4.0"
 parking_lot = "0.12.1"
 flutter_rust_bridge_macros = { path = "../frb_macros", version = "1" }
+uuid = { version = "1.1.2", optional = true }
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
-allo-isolate = { version = "0.1.14-beta.2", features = ["anyhow", "chrono"] }
+allo-isolate = { version = "0.1.14-beta.3", features = [
```
Desdaemon, line 22: nit: replace with top-level [dependency features](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) to allow users to opt out

frb_rust/src/ffi/web.rs
Desdaemon: IMO looks fine, perhaps unwrap and fail fast instead of silently ignoring?

frb_dart/lib/src/helpers.dart
```diff
@@ -148,3 +148,15 @@ Duration wire2apiDuration(int ts) {
   }
   return Duration(microseconds: ts);
 }
+
+Uint8List wire2apiConcatenateBytes(List raw) {
```
Desdaemon, line 152: nit: `List<UuidValue>`, we can widen this later if needed

Desdaemon (commented):
Some nits, otherwise great job!

frb_rust/src/ffi/web.rs
Roms1383: this one [was already raised](https://github.com/sunshine-protocol/allo-isolate/pull/30#discussion_r977856082) by @shekohex upstream and matches with my implementation of `IntoDart` there : I settled down on not panicking as it raised a concern for him, and instead mentioned in the docs that it could return an incomplete buffer (guaranteed though, that any contained uuid will be valid) as per `std::io::Write::write`.

frb_codegen/src/generator/dart/ty_delegate.rs
Roms1383: resolved in 86edba18948fa2e64a68c0cad89bd96ad168f558

frb_rust/src/ffi/web.rs
Roms1383: resolved in 86edba18948fa2e64a68c0cad89bd96ad168f558

frb_dart/lib/src/helpers.dart
Roms1383: resolved in 8be4048bd8476aa23480a143e586ed694c75b28d

frb_rust/Cargo.toml
Roms1383: not sure, like 52c14600fddce107b36658b21bea024eebdd0d71 ?

frb_dart/lib/src/helpers.dart
Roms1383: ah so the reason I did it this way was to avoid forcing [package uuid](https://pub.dev/packages/uuid) as dependency on `frb dart`, are you sure ?

Desdaemon: Dart dependencies are relatively cheap to add and can be tree-shaken, so I think it should be fine.

Roms1383: ok ! 👌

Roms1383: ok done in de8eaac806974e1d1fc70e6a1f690c5fc9b0f01e, sorry I forgot the co-authoring thing I'm getting tired lol 😴

fzyzcjy (approved):
LGTM

Waiting for @Desdaemon whether OK

Desdaemon:
LGTM

Roms1383:
Thanks guys, I know I solicited you a lot lately ! 🚀 🚀 🚀

rob-mur:
Hi

Unsure if this is the right place, but I ran into an issue with this. When trying to use flutter_rust_bridge and build on the web I currently get an error when using the command (which is the one used by flutter_rust_bridge_serve or equivalently dart run flutter_rust_bridge:serve)
```
cargo build --lib --target wasm32-unknown-unknown -Z build-std=std,panic_abort
```

With the actual error being:

```
error[E0432]: unresolved import `crate::ffi::UUID_SIZE_IN_BYTES`
 --> src\ffi\web.rs:4:5
  |
4 | use crate::ffi::UUID_SIZE_IN_BYTES;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `UUID_SIZE_IN_BYTES` in `ffi`
```

I've checked that this now doesn't occur on the latest master branch, so presumably it's been fixed but not released?

Anything I could do do work around this issue in the meantime?

fzyzcjy:
@rob-mur hi could you please create an issue in the issues section? may need more info e.g versions

fzyzcjy:
@rob-mur btw ensure your flutter_rust_bridge dependency, in your Cargo.toml (and Cargo.lock), is *also the latest*

rob-mur:
So you are correct that the template repository is still pointing to version 1 and so my Cargo.toml had that too. However, pointing to 1.48.0 leads to the observed crash. Thankfully though this gives me a workaround, if I use the following entry in Cargo.toml then it works as the current master branch works.

```
flutter_rust_bridge = {git = "https://github.com/fzyzcjy/flutter_rust_bridge.git", branch = "master"}
```

I can make an issue on why it doesn't work on 1.48.0 but at least I can keep going for now!

PR #663, opened 2022-09-03
Refactor and enhance `SyncReturn` to support more types

SoLongAndThanksForAllThePizza:
If there are any problems with my code or design, please let me know😊😊

## Adjustments

* Added a new IR type: IrTypeSyncReturn.
* SyncReturn now supports String, Vec\<u8\> and most of the primitive types(`bool`, `u8`, `u16`, `u32`, `u64`, `i8`, `i16`, `i32`, `i64`, `f32`, `f64`).
* Removed SyncReturnVecU8 from IrTypeDelegate since SyncReturn now has an individual IR.

## Design

* Added `parseSuccessData` for `FlutterRustBridgeSyncTask`, it does the type conversion for Dart.
* Implemented `From` trait for `WireSyncReturnStruct`, it does the type conversion for Rust.
* Added generic type constraint for SyncReturn, now if users are trying to use an unsupported type for SyncReturn, they will immediately get a compiler error.

## Checklist

- [x] New tests are added to ensure new features are working. End-to-end tests are usually in the `./frb_example/pure_dart` example, more specifically, `rust/src/api.rs` and `dart/lib/main.dart`.
- [x] The code generator is run and the code is formatted (e.g. via `just refresh_all`).
- [x] If this PR adds/changes features, documentations (in the `./book` folder) are updated.
- [x] CI is passing.

PR #730, opened 2022-09-22
Allow streaming functions to omit return type

Desdaemon:
Fixes a long-standing bug with the codegen emitting broken code for streaming functions.

## Checklist

- [x] CI is passing.

## Remark for PR creator


fzyzcjy:
LGTM, except that maybe we should remain some tests to return streams, instead of letting all return unit?

Desdaemon:
I kept the registration test since it is supposed to test the error return path, anything else would you like to keep @fzyzcjy?

fzyzcjy:
Then LGTM :) Waiting for CI to merge

Issue #731, opened 2022-09-23
[Bug] Broken links in the docs


Roms1383:
### Describe the bug

I was reading the docs and realized some intra links are broken. Apparently this is related to well-known issue with `mdbook` and turns out that links from subfolders are not generated when not referenced in `SUMMARY.md`.

People also pointed out to [mdbook-linkcheck](https://github.com/Michael-F-Bryan/mdbook-linkcheck) which I gave a quick try and indeed shows the broken links (it could probably be a nice addition to the CI).

### Codegen logs with `RUST_LOG=debug` environment variable

```shell
_unrelated_
```

### To Reproduce

For example in [Simple correspondence](http://cjycode.com/flutter_rust_bridge/feature/lang_simple.html) when clicking on the link at the bottom of the page.

Same issue confirmed when generating the book locally.

### Additional context

I haven't filled most of the fields above since it's purely related to `mdbook`, from content generated locally or in the CI.
So the local context doesn't really matter here.

Roms1383:
It's a good first issue, if someone wants to get started.
Otherwise I'll address it later.

fzyzcjy:
Looks great, we may

- [ ] apply checker in CI
- [ ] fix it

thomas725:
I'm unsure if this is related but no matter what link to what docs page I try to open, I always get 404 errors now. I believe they still worked a few days ago:
For example:
https://fzyzcjy.github.io/flutter_rust_bridge/
http://cjycode.com/flutter_rust_bridge/template/setup_web.html

fzyzcjy:
@thomas725 That's a big problem #741

Roms1383:
Did I break something?! It was supposed to fix it 😅

fzyzcjy:
Well reverting that PR makes doc work now, so I guess it does accidentically break sth

Roms1383:
Ok I'll look into it 👌 Sorry for the inconvenience

fzyzcjy:
That's OK!

PR #736, opened 2022-09-24
Fix broken links in book

Roms1383:
A couple of links are broken in the docs, see #731.
It fixes them and add [mdbook-linkcheck](https://github.com/Michael-F-Bryan/mdbook-linkcheck) to the CI.

## Checklist

- [x] CI is passing.

### DIFF
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -25,8 +25,8 @@ jobs:
         with:
           mdbook-version: 'latest'
       
-      - name: Setup `mdbook-mermaid`
-        run: cargo install mdbook-mermaid
+      - name: Setup mdbook plugins (mermaid, linkcheck)
+        run: cargo install mdbook-linkcheck mdbook-mermaid
 
       - run: mdbook build ./book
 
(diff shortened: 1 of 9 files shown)


.github/workflows/gh-pages.yaml
```diff
@@ -25,8 +25,8 @@ jobs:
         with:
           mdbook-version: 'latest'
       
-      - name: Setup `mdbook-mermaid`
-        run: cargo install mdbook-mermaid
+      - name: Setup mdbook plugins (mermaid, linkcheck)
```
fzyzcjy, line 28: Wonder whether this is enough to auto check in CI? If so then OK

fzyzcjy (approved):
LGTM % CI passes, good job :)

.github/workflows/gh-pages.yaml
Roms1383: Yes, it checks automatically when `mdbook build` runs (which is the command ran just after: `mdbook build ./book`).

fzyzcjy:
Temporarily revert this to see whether #741 is ok now. Feel free to re-PR without this bug!

Roms1383:
@fzyzcjy I browsed the files again there shouldn't be any problem : could you describe more accurately what kind of bugs it generated ?

fzyzcjy:
@Roms1383 when viewing the website every page says 404

PR #764, opened 2022-10-06
Fix return for struct with methods

Zaitam:
#585 

## Checklist

- [x] CI is passing.

### DIFF
--- a/frb_codegen/src/generator/dart/mod.rs
+++ b/frb_codegen/src/generator/dart/mod.rs
@@ -327,6 +327,9 @@ fn generate_dart_implementation_body(spec: &DartApiSpec, config: &Opts) -> Acc<D
         )
     }));
 
+    lines.push(section_header("wire2api"));
+    lines.push(dart_wire2api_funcs.join("\n\n"));
+
     lines.push("}\n".into());
 
     lines.push_all(section_header("api2wire"));
@@ -339,9 +342,6 @@ fn generate_dart_implementation_body(spec: &DartApiSpec, config: &Opts) -> Acc<D
         Io | Wasm => "}\n".into(),
         Common => "".into(),
     }));
-
-    lines.push(section_header("wire2api"));
-    lines.push(dart_wire2api_funcs.join("\n\n"));
 
     if config.wasm_enabled {
         push_wasm_module(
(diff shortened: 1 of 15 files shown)


fzyzcjy:
Busy on https://github.com/fzyzcjy/flutter_smooth, will review today :)

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -317,6 +317,9 @@ fn generate_dart_implementation_body(spec: &DartApiSpec, config: &Opts) -> Acc<D
         )
     }));
 
+    lines.push(section_header("wire2api"));
+    lines.push(dart_wire2api_funcs.join("\n\n"));
+
```
fzyzcjy, line 322: nit: Looks like these lines are moved from below to here. If this is not a necessary change, maybe we should revert it? otherwise the generated code will be changed while not necessary

fzyzcjy (commented):
Ready to merge, just one nit :)

fzyzcjy:
@all-contributors please add @Zaitam for code

fzyzcjy:
https://github.com/all-contributors?type=source please add @Zaitam for code

fzyzcjy:
@all-contributors please add @Zaitam for code

fzyzcjy:
@all-contributors please add @Zaitam for code, doc

frb_codegen/src/generator/dart/mod.rs
Zaitam: Yeah, it is needed to include the wire2api into the class, so we can use the 'this'

fzyzcjy: image

I mean, shall we put it in line 333 as the old case? otherwise the generated output changes a lot while seems not necessary.

Zaitam:
Well, this was unexpected...

fzyzcjy:
Oops :(

fzyzcjy:
Luckily it is still there: https://github.com/fzyzcjy/flutter_rust_bridge/commit/57868d4fa570f827fde21429682deb2f13e934d4, you can use that commit to recover your code

frb_codegen/src/generator/dart/mod.rs
Zaitam: No, as they require to be inside the class (Closed on line 323) to use the `this` class.

PR #778, opened 2022-10-13
Add note about symbol stripping on iOS

tmpfs:
Add a note to Troubleshooting about an annoying issue with symbols being stripped only when deploying an iOS app.


fzyzcjy:
IIRC the doc already talked something like that? 👀

Issue #837, opened 2022-11-10
Opaque Dart type support


rogurotus:
adding support for opaque dart types.
Let's add the ability for rust to receive dart objects (Dart_Handle). 

The dart API(`dart_api_dl.h`) provides functions for dart object persistence.
https://github.com/dart-lang/sdk/issues/47356#issuecomment-1302306273

- Dart_NewPersistentHandle
- Dart_HandleFromPersistent
- Dart_DeletePersistentHandle

something like that
```
// internal implementation
struct DartOpaque(Dart_PersistentHandle);

impl Drop for DartOpaque {
  fn drop(&mut self) {
    Dart_DeletePersistentHandle(self.0);
  }
}

// Rust
pub fn dart_opaque(obj: DartOpaque) {
  // rust can save, delete, or use as needed.
}

// Dart
api.dartOpaque(Object obj);
```

about !Send, !Sync `Dart_Handle`:
the easiest way is to provide only `syncReturn executor`.

fzyzcjy:
> the easiest way is to provide only syncReturn executor.

hmm IIRC syncreturn is very limited and async is the main thing, so hopefully we can support async

maybe via object pool?

rogurotus:
if I'm not mistaken, we still can't use dart_handle in the thread pool (unless we save it, give it back). And so with object pool we lose compatibility with dart api \ c api. 
I also have a development for an opaque syncReturn type. This covers a wide range of needs.

rogurotus:
I understand the idea, but it would be highly desirable to maintain compatibility with the Dart_Handle api.
And for safe `drop` on the threadpool, instead of manual deletion, we can delegate this to Dart through the Dart port

fzyzcjy:
> but it would be highly desirable to maintain compatibility with the Dart_Handle api.

May I know a bit more why we need to have this compatibility? IMHO most flutter/rust dev do not use that API directly, and even if use it, I have not seen where compatibility has advantage.

> And for safe drop on the threadpool, instead of manual deletion, we can delegate this to Dart through the Dart port

looks interesting. But that will cause objects never freed until the current sync thing is finished in dart thread. For example, inside one (possibly janky) frame.

fzyzcjy:
Again, this feature is quite interesting and complex, that's why we need to consider it carefully :)

rogurotus:
[...]

PR #853, opened 2022-11-24
Extending Opaque Types

rogurotus:
This PR allows you to receive, store and pass a `Dart Object`.

Implement [Opaque Dart type support](https://github.com/fzyzcjy/flutter_rust_bridge/issues/837#top).

Additionally:
- fix `RustOpaque` doc

## Checklist

- [x] CI is passing.


fzyzcjy:
Eagerly looking forward to seeing it implemented!

rogurotus:
> Eagerly looking forward to seeing it implemented!

In general, the implementation is ready.

fzyzcjy:
Ok then let me have a first review :)

Also /cc @Desdaemon

frb_rust/src/ffi/web.rs
```diff
@@ -405,3 +408,65 @@ mod tests {
...
+}
+
+impl Drop for DartOpaque {
+    fn drop(&mut self) {
+        if self.drop {
+            let ptr = self.handle;
+            if std::thread::current().id() == self.id {
+                drop(unsafe { support::box_from_leak_ptr(ptr) });
+            } else {
+                Channel::new(PortLike::broadcast(&self.port)).post(ptr);
```
fzyzcjy, line 468: nit: we need to test both cases. and probably add some comments in those tests mentioning this impl

book/src/contributing/rust_opaque_type_safety.md
```diff
@@ -1,9 +1,9 @@
-# Opaque type safety
```
fzyzcjy, line 1: btw, would be great to add a desing doc about DartOpaque since this is a great (thus hard-to-understand for new learners) PR!

frb_rust/src/ffi/io.rs
```diff
@@ -31,3 +40,63 @@ mod tests {
...
+
+    /// Flag to cancel the drop when sending to dart.
+    drop: bool,
+
+    /// The ID of the thread on which the Dart Object was created.
+    id: ThreadId,
+}
+
+unsafe impl Send for DartOpaque {}
+unsafe impl Sync for DartOpaque {}
```
fzyzcjy, line 60: similar as before - are we sure it is really send/sync?

frb_rust/src/ffi/io.rs
```diff
@@ -31,3 +40,63 @@ mod tests {
...
     }
 }
+
+#[derive(Debug)]
+pub struct DartOpaque {
+    /// Dart object handle that won't be removed by GC.
+    handle: Dart_PersistentHandle,
+
+    /// Used to drop the `handle` on the Dart thread
+    port: Channel,
```
fzyzcjy, line 50: uber-nit: what about renaming to `drop_port` to make it match comments

frb_rust/src/ffi/io.rs
```diff
@@ -31,3 +40,63 @@ mod tests {
...
+
+    /// Flag to cancel the drop when sending to dart.
+    drop: bool,
+
+    /// The ID of the thread on which the Dart Object was created.
+    id: ThreadId,
```
fzyzcjy, line 56: uber-nit: what about thread_id, since `DartOpaque.id` looks like "id of the object"

frb_rust/src/dart_api_dl/include/dart_tools_api.h
```diff
@@ -0,0 +1,499 @@
+// Copyright (c) 2011, the Dart project authors.  Please see the AUTHORS file
```
fzyzcjy, line 1: similar as before, maybe we do not need tools I guess?

frb_rust/src/dart_api_dl/include/dart_api.h
```diff
@@ -0,0 +1,3953 @@
+/*
```
fzyzcjy, line 1: Indeed I guess we are only using one or two dart api files? So we can remove the others to keep it more maintainable

frb_rust/Cargo.toml
```diff
@@ -20,6 +20,8 @@ uuid = { workspace = true, optional = true }
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 allo-isolate = { version = "0.1.14-beta.3", features = ["anyhow"] }
+dart-sys = "2.0"
```
fzyzcjy, line 23: hmm if we are using dart-sys, is it possible we do not manually copying those .h files

frb_example/with_flutter/lib/bridge_generated.dart
```diff
@@ -220,6 +221,9 @@ class FlutterRustBridgeExampleImpl implements FlutterRustBridgeExample {
         argNames: [],
       );
 
+  void close() {
```
fzyzcjy, line 224: uber-nit: shall we use `dispose` instead of `close`, which is more Dart/Flutter-like

frb_example/with_flutter/lib/bridge_generated.io.dart
```diff
@@ -12,7 +12,26 @@ import 'package:meta/meta.dart';
 import 'dart:ffi' as ffi;
 
 class FlutterRustBridgeExamplePlatform extends FlutterRustBridgeBase<FlutterRustBridgeExampleWire> {
-  FlutterRustBridgeExamplePlatform(ffi.DynamicLibrary dylib) : super(FlutterRustBridgeExampleWire(dylib));
+  final _port = RawReceivePort();
+  NativePortType get port => _port.sendPort.nativePort;
```
fzyzcjy, line 16: uber-nit: `_dropPort`?

frb_example/with_flutter/lib/bridge_generated.io.dart
```diff
@@ -12,7 +12,26 @@ import 'package:meta/meta.dart';
...
+    }
+
+    _port.handler = (response) {
+      inner.drop_DartOpaqueFlutterRustBridgeExample(response);
+    };
+    dylib.lookupFunction<ffi.IntPtr Function(ffi.Pointer<ffi.Void>), int Function(ffi.Pointer<ffi.Void>)>(
+        'init_dart_api_dl')(ffi.NativeApi.initializeApiDLData);
+  }
+  void close() {
+    _port.close();
```
fzyzcjy, line 33: ⚠️ Problem: What if an object is dropped (and thus send object to this port) *after* someone calls this `close` method? Then it looks never be dropped and memory leak.

frb_example/with_flutter/lib/bridge_generated.io.dart
```diff
@@ -12,7 +12,26 @@ import 'package:meta/meta.dart';
...
+
+    if (initResult != 0) {
+      throw 'Failed to initialize Dart API. Code: $initResult';
+    }
+
+    _port.handler = (response) {
+      inner.drop_DartOpaqueFlutterRustBridgeExample(response);
+    };
+    dylib.lookupFunction<ffi.IntPtr Function(ffi.Pointer<ffi.Void>), int Function(ffi.Pointer<ffi.Void>)>(
+        'init_dart_api_dl')(ffi.NativeApi.initializeApiDLData);
```
fzyzcjy, line 30: maybe we can extract it to frb_dart instead of the generated code

frb_example/with_flutter/lib/bridge_generated.io.dart
```diff
@@ -12,7 +12,26 @@ import 'package:meta/meta.dart';
 import 'dart:ffi' as ffi;
 
 class FlutterRustBridgeExamplePlatform extends FlutterRustBridgeBase<FlutterRustBridgeExampleWire> {
-  FlutterRustBridgeExamplePlatform(ffi.DynamicLibrary dylib) : super(FlutterRustBridgeExampleWire(dylib));
+  final _port = RawReceivePort();
+  NativePortType get port => _port.sendPort.nativePort;
+  FlutterRustBridgeExamplePlatform(ffi.DynamicLibrary dylib) : super(FlutterRustBridgeExampleWire(dylib)) {
+    var initResult =
```
fzyzcjy, line 18: What if there are two `FlutterRustBridgeExamplePlatform` instances (or two instances of different such auto-generated classes of frb)? Will we face trouble because we init twice?

Moreover, consider hot-restart scenario: Even if users have only *one* such instance, it will be created multiple times (and without dispose at all)

Btw, do we really need it in Flutter scenarios? Since flutter should already init dart

frb_example/with_flutter/lib/bridge_generated.io.dart
fzyzcjy, line 30: We need to make "init_dart_api_dl" as well as "create the port" lazily loaded. In other words, do not do it, unless we really want to create a Dart opaque. Otherwise, for all users who do not use Dart opaque, they are paying extra price (and risking possible bugs)

frb_example/with_flutter/lib/bridge_generated.io.dart
```diff
@@ -378,6 +397,19 @@ class FlutterRustBridgeExampleWire implements FlutterRustBridgeWireBase {
       _lookup<ffi.NativeFunction<ffi.Pointer<wire_uint_8_list> Function(ffi.Int32)>>('new_uint_8_list_0');
   late final _new_uint_8_list_0 = _new_uint_8_list_0Ptr.asFunction<ffi.Pointer<wire_uint_8_list> Function(int)>();
 
+  void drop_DartOpaqueFlutterRustBridgeExample(
```
fzyzcjy, line 400: Is it something that is the same for all classes? If so, maybe we do not need to put it to *generated* code, but instead be in frb_dart/frb_rust

frb_example/with_flutter/lib/bridge_generated.web.dart
```diff
@@ -12,10 +12,22 @@ import 'package:meta/meta.dart';
...
 class FlutterRustBridgeExamplePlatform extends FlutterRustBridgeBase<FlutterRustBridgeExampleWire>
     with FlutterRustBridgeSetupMixin {
-  FlutterRustBridgeExamplePlatform(FutureOr<WasmModule> dylib) : super(FlutterRustBridgeExampleWire(dylib)) {
+  static int drop_id_port = 0;
+  final ReceivePort _port;
+  NativePortType get port => _port.sendPort.nativePort;
+  FlutterRustBridgeExamplePlatform(FutureOr<WasmModule> dylib)
+      : _port = broadcastPort('__frb_dart_opaque_drop_$drop_id_port'),
+        super(FlutterRustBridgeExampleWire(dylib)) {
+    ++drop_id_port;
```
fzyzcjy, line 21: Hmm what about

```
class _DropIdPortGenerator {
  static final instance = _DropIdPortGenerator.();
  _DropIdPortGenerator.();
  
  int nextPort = 0;
  String create() => '__frb_dart_opaque_drop_${nextPort++}';
}
```

etc

frb_example/with_flutter/rust/src/bridge_generated.io.rs
```diff
@@ -197,6 +197,15 @@ impl Wire2Api<Vec<u8>> for *mut wire_uint_8_list {
         }
     }
 }
+// Section: dart opaque related functions
+
+#[no_mangle]
+pub extern "C" fn drop_DartOpaqueFlutterRustBridgeExample(ptr: usize) {
```
fzyzcjy, line 203: same as before - is it possible we put it to frb_dart

frb_example/pure_dart/dart/lib/main.dart
```diff
@@ -692,7 +693,30 @@ void main(List<String> args) async {
     });
   });
 
-  group('opaque type', () {
+  group('dart opaque type', () {
+    f() => 'Test_String';
```
fzyzcjy, line 697:
```suggestion
    String f() => 'Test_String';
```

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -958,36 +962,56 @@ pub fn nested_id(id: [TestId; 4]) -> [TestId; 2] {
     }
 }
 
-/// Opaque types
+pub fn sync_dart_opaque(not_temp: DartOpaque) -> SyncReturn<String> {
+    SyncReturn("test".to_owned())
+}
+
+pub fn async_dart_opaque(not_temp: DartOpaque) -> String {
```
fzyzcjy, line 969: uber-nit: What about `sync_accept_dart_opaque` etc, i.e. tell readers we are using it as arg

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -958,36 +962,56 @@ pub fn nested_id(id: [TestId; 4]) -> [TestId; 2] {
...
+    "async test".to_owned()
+}
+
+pub fn loop_back(not_temp: DartOpaque) -> DartOpaque {
+    not_temp
+}
+
+pub fn exotic_drop(not_temp: DartOpaque) -> SyncReturn<i32> {
+    std::thread::spawn(move || {
+        drop(not_temp);
```
fzyzcjy, line 979: What about adding a test that `sleep` for a second before dropping, so we double test the timing as well

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -1005,43 +1029,82 @@ pub fn run_enum_opaque(opaque: EnumOpaque) -> String {
...
+pub fn unwrap_dart_opaque(opaque: DartOpaque) -> SyncReturn<String> {
+    #[cfg(target_family = "wasm")]
+    {
+        let js = opaque.try_unwrap().unwrap();
+        SyncReturn(js.as_string().unwrap())
+    }
+    #[cfg(not(target_family = "wasm"))]
+    {
+        let handle = opaque.try_unwrap().unwrap();
+        unsafe { Dart_DeletePersistentHandle_DL_Trampolined(handle) };
```
fzyzcjy, line 1093: Oops, we should never let users of frb touch these dangerous functions. (this api.rs is an example, so it is like user code not library code)

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -1005,43 +1029,82 @@ pub fn run_enum_opaque(opaque: EnumOpaque) -> String {
...
+pub fn panic_unwrap_dart_opaque(opaque: DartOpaque) -> String {
+    #[cfg(target_family = "wasm")]
+    {
+        let js = opaque.try_unwrap().unwrap();
+        js.as_string().unwrap()
+    }
+    #[cfg(not(target_family = "wasm"))]
+    {
+        let handle = opaque.try_unwrap().unwrap();
+        unsafe { Dart_DeletePersistentHandle_DL_Trampolined(handle) };
```
fzyzcjy, line 1107: same

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -1005,43 +1029,82 @@ pub fn run_enum_opaque(opaque: EnumOpaque) -> String {
...
     opaque.second.hide_data();
 }
+
+pub fn unwrap_rust_opaque(opaque: RustOpaque<HideData>) -> String {
+    let hide_data = opaque.try_unwrap().unwrap();
+    hide_data.hide_data()
+}
+
+pub fn unwrap_dart_opaque(opaque: DartOpaque) -> SyncReturn<String> {
+    #[cfg(target_family = "wasm")]
```
fzyzcjy, line 1085: for end users, it would be great not to have `cfg`

frb_example/pure_dart/rust/src/bridge_generated.io.rs
```diff
@@ -560,6 +601,15 @@ pub extern "C" fn new_BoxDartDebug() -> wire_BoxDartDebug {
     wire_BoxDartDebug::new_with_null_ptr()
 }
 
+#[no_mangle]
+pub extern "C" fn new_DartOpaque(
```
fzyzcjy, line 605: similar to before, maybe this can be put into frb_rust

frb_example/pure_dart/dart/lib/main.dart
```diff
@@ -692,7 +693,30 @@ void main(List<String> args) async {
     });
   });
 
-  group('opaque type', () {
+  group('dart opaque type', () {
```
fzyzcjy, line 696: would be great to have tests for *ephemeral* and *large* objects to detect memory leak. For example, `Uint8List` with 200MB size.

I have some boilerplate code for ensuring there is no memory leak, and I can provide it here. Ping me if I do not paste it.

frb_example/pure_dart/dart/lib/main.dart
```diff
@@ -692,7 +693,30 @@ void main(List<String> args) async {
...
+
+    test('drop', () async {
+      expect(await api.asyncDartOpaque(notTemp: f), 'async test');
+      expect(api.syncDartOpaque(notTemp: f), 'test');
+    });
+
+    test('unwrap', () async {
+      var data = 'Test';
+      expect(api.unwrapDartOpaque(opaque: data), 'Test');
+      await expectLater(() => api.panicUnwrapDartOpaque(opaque: data), throwsA(isA<FfiException>()));
```
fzyzcjy, line 715: would be great to check `data` is GCed later. (using the boilerplate I will give)

book/src/feature/lang_opaque.md
```diff
@@ -1,18 +1,18 @@
 # Arbitrary types (opaque)
 
-Any Rust type, even if it is not supported using features of this library, can be used in Dart. This is done by wrapping it with `Opaque`.
+Any Rust type, even if it is not supported using features of this library, can be used in Dart. This is done by wrapping it with `RustOpaque` also Rust can get opaque Dart type in the `DartOpaque` wrap.
```
fzyzcjy, line 3:
```suggestion
Any Rust type, even if it is not supported using features of this library, can be used in Dart. This is done by wrapping it with `RustOpaque`. In addition, Rust can get opaque Dart type in the `DartOpaque` wrap.
```

book/src/feature/lang_opaque.md
```diff
@@ -1,18 +1,18 @@
...
 
-The Dart opaque objects should be disposed *manually*, though it will also be disposed when it is GCed, that is discouraged, due to [suggestions by Dart team](https://github.com/fzyzcjy/flutter_rust_bridge/issues/775#issuecomment-1274635037). Think of it just like a lot of Flutter objects that we are familiar with, such as `ui.Image` - we have to manually dispose them as well.
+The Dart rust opaque objects should be disposed *manually*, though it will also be disposed when it is GCed, that is discouraged, due to [suggestions by Dart team](https://github.com/fzyzcjy/flutter_rust_bridge/issues/775#issuecomment-1274635037). Think of it just like a lot of Flutter objects that we are familiar with, such as `ui.Image` - we have to manually dispose them as well. The Dart opaque objects are always deleted on the parent Dart thread.
```
fzyzcjy, line 5: Hmm what is "dart rust opaque" that should be disposed manually? Looks like a typo

book/src/feature/lang_opaque.md
```diff
@@ -24,6 +24,26 @@ await api.functionThatUsesSomeOpaqueData(opaque);
...
+var fn = answer_obj as String Function();
+print(fn());
+```
+
 ## Implementation details
 
-As for how it is implemented as well as the design towards safety, please refer to [this doc](../contributing/opaque_type_safety.md).
\ No newline at end of file
+As for how it is implemented as well as the design towards safety, please refer to [this doc](../contributing/rust_opaque_type_safety.md).
+[this doc](../contributing/dart_opaque_type_safety.md).
```
fzyzcjy, line 49:
```suggestion
As for how it is implemented as well as the design towards safety, please refer to [this doc](../contributing/rust_opaque_type_safety.md) and [this doc](../contributing/dart_opaque_type_safety.md).
```

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -327,14 +328,36 @@ fn generate_dart_implementation_body(spec: &DartApiSpec, config: &Opts) -> Acc<D
         ),
         io: format!(
             "class {plat} extends FlutterRustBridgeBase<{wire}> {{
-                {plat}(ffi.DynamicLibrary dylib) : super({wire}(dylib));",
+                final _port = RawReceivePort();
```
fzyzcjy, line 331: as mentioned before - would be great to put to frb_dart/frb_rust

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -349,6 +372,12 @@ fn generate_dart_implementation_body(spec: &DartApiSpec, config: &Opts) -> Acc<D
         )
     }));
 
+    lines.push_acc(Acc {
```
fzyzcjy, line 375: (same put to frb_dart/frb_rust)

If the port is even a field in parent class in frb_dart, then this generated class even do not need to bother this

frb_codegen/src/ir/ty_delegate.rs
```diff
@@ -30,6 +30,7 @@ pub enum IrTypeDelegate {
     Array(IrTypeDelegateArray),
     String,
     StringList,
+    DartOpaque(IrTypeDartOpaque),
```
fzyzcjy, line 33: Btw curious why are we impl it as delegate?

rogurotus:
> Ok then let me have a first review :)

can't add reviewer

frb_example/pure_dart/dart/lib/bridge_generated.dart
```diff
@@ -1347,6 +1348,70 @@ class FlutterRustBridgeExampleSingleBlockTestImpl implements FlutterRustBridgeEx
         argNames: ["id"],
       );
 
+  String syncDartOpaque({required Object notTemp, dynamic hint}) {
```
fzyzcjy, line 1351: btw, curious why is it named like "notTemp"

fzyzcjy:
Still reviewing

frb_rust/src/ffi/io.rs
```diff
@@ -31,3 +40,63 @@ mod tests {
...
+        } else {
+            Err(self)
+        }
+    }
+}
+
+impl IntoDart for DartOpaque {
+    fn into_dart(mut self) -> ffi::DartCObject {
+        self.drop = false;
+        self.handle.into_dart()
```
fzyzcjy, line 88: Nice trick! Never knew Dart_PersistentHandle can be converted to DartCObject

fzyzcjy (commented):
Great job! Many comments are duplicated (because the same idea is around multiple codes) and many are nits

fzyzcjy:
TODO(fzyzcjy): examine how the pointers are passed around and whether they are safe

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -629,6 +629,10 @@ lazy_static! {
     static ref EVENTS: Mutex<Option<StreamSink<Event>>> = Default::default();
 }
 
+lazy_static! {
+    static ref DART_OPAQUE: Mutex<Vec<DartOpaque>> = Default::default();
```
fzyzcjy, line 633: seems not used?

frb_example/pure_dart/rust/src/api.rs
fzyzcjy, line 1107: imho, unwrap_dart_opaque vs panic_unwrap_dart_opaque is only different in whether it is a SyncReturn? thus not sure why "panic"

fzyzcjy (commented):
(nits)

frb_example/pure_dart/rust/src/api.rs
rogurotus: exactly in SyncReturn. Since in this case it is the same thread that the DartObject (JsValue) was created on, we can do an unwrap

frb_rust/src/ffi/io.rs
rogurotus: of course, the api checks the thread_id and delegates this to Dart when it drops

frb_example/pure_dart/rust/src/api.rs
fzyzcjy: I see. Then it looks a bit confusing for readers

frb_rust/src/ffi/web.rs
```diff
@@ -405,3 +408,65 @@ mod tests {
         assert_eq!(ns, 567_000_000);
     }
 }
+
+#[derive(Debug)]
+pub struct DartOpaque {
+    /// Dart object ptr
+    handle: *mut JsValue,
```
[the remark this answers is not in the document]
rogurotus: You can make `Box<JsValue>`, but in any case it will be about the same. This is because `JsValue` is not a `Send` (we are delegating the drop to Dart).

frb_example/with_flutter/lib/bridge_generated.io.dart
rogurotus: For this case, we can catch drop errors (writing to a closed port) and save the necessary pointers in `lazy_static`, after which, on the Dart side, clean up if an error occurs.

(The first thing that came to mind)

frb_rust/src/ffi/web.rs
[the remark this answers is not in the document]
fzyzcjy: Hmm, if it is not Send, is it safe we unsafely force it to be Send?

frb_rust/src/ffi/io.rs
fzyzcjy: Good point.

Then maybe we need to add a comment saying that for future readers

frb_example/with_flutter/lib/bridge_generated.io.dart
fzyzcjy: Hmm looks quite complicated, not very sure whether this is the correct way to do so

rogurotus: 1. how about making close() (port) asynchronous, which will wait for the execution of incomplete functions and then we will throw an error when trying to use the api

2. we can omit ports altogether and use the above mechanism, but then DartOpaque will be removed after the function is executed (for thread_pool)

frb_rust/src/ffi/web.rs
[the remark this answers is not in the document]
rogurotus: Yes, we only use how to give back Dart or unwrap on a Dart thread

frb_example/pure_dart/rust/src/api.rs
rogurotus: here we have to use `#[cfg target]` because `sleep` and `thread::spawn` are not web.

frb_example/pure_dart/rust/src/api.rs
rogurotus: how else to test? without it there will be a leak

frb_dart/lib/src/ffi.dart
```diff
@@ -15,6 +15,10 @@ abstract class FrbOpaque extends FrbOpaqueBase {
   /// Pointer to this opaque Rust type.
   PlatformPointer _ptr;
 
+  /// flag to transfer ownership of an opaque type
+  /// in Rust when used as an argument.
+  var move = false;
```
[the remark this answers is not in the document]
rogurotus: Will be in another PR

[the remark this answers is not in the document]
rogurotus: Will be in another PR

frb_codegen/src/ir/ty_delegate.rs
rogurotus: Well of course it is possible not to delegate. fixed

frb_example/pure_dart/dart/lib/bridge_generated.dart
rogurotus: forgot to change

frb_example/pure_dart/dart/lib/main.dart
rogurotus: ping

frb_rust/src/ffi/web.rs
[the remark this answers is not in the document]
fzyzcjy: Great! Then maybe we can add a comment there.

Btw, I personally prefer `Box` over `*mut` because the former is higher level and looks safer. So is it possible we migrate to it?
(I have discussed the possibility of `struct DartOpaque { inner: Option<DartOpaqueInner> }`, so double-free is not a problem as long as we properly set Option to None

frb_example/with_flutter/lib/bridge_generated.io.dart
fzyzcjy: > which will wait for the execution of incomplete functions and then we will throw an error when trying to use the api

IMHO the DartOpaque object can live for arbitrarily long time, because it can be passed to other threads etc, or even assign it to a static variable

> we can omit ports altogether and use the above mechanism, but then DartOpaque will be removed after the function is executed (for thread_pool)

That does not look quite good, since users may want to save DartOpaque somewhere and thus outlive the function

fzyzcjy: Given that Flutter users may never `close` it at all, maybe we (almost) do not need to change your original code at all: Maybe we should simply warn the user if "object dropped after close", possibly by a `log::warn!` macro?

This is imperfect solution as well indeed.

frb_example/pure_dart/rust/src/api.rs
fzyzcjy: Then what about a dead loop + check time until it passes one second, i.e. "busy wait".

Btw you are already using spawn, so I guess adding sleep is not a problem, if *both* are not web?

frb_example/with_flutter/lib/bridge_generated.io.dart
rogurotus: in the second case, we will not delete all DartOpaque, but only for which Drop

frb_example/pure_dart/rust/src/api.rs
fzyzcjy: The main idea is: This pure_dart example is the role model for every frb user. So, do we expect frb users to write down these two lines of code (surely not!).

We may wrap it to some safe operations. Not very sure what is the feature here that you want to expose to the users?

frb_example/with_flutter/lib/bridge_generated.io.dart
rogurotus: > Maybe we should simply warn the user if "object dropped after close", possibly by a log::warn! macro?
In this case, memory leaks will occur.

frb_example/pure_dart/dart/lib/main.dart
fzyzcjy: Done. https://stackoverflow.com/a/74572599/4619958 (https://gist.github.com/fzyzcjy/e68c375643d7c77942cdc8fb5f01de18)

frb_rust/src/ffi/web.rs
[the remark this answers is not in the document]
rogurotus: Yes, I already refactored

frb_example/with_flutter/lib/bridge_generated.io.dart
fzyzcjy: > in the second case, we will not delete all DartOpaque, but only for which Drop

Not exactly sure the details of the second case indeed...

> In this case, memory leaks will occur.

Yes, that's exactly what I said in "⚠️ Problem" (i.e. first comment). But after rethinking it, I guess Flutter users will *never* close it, so it is not a problem. In addition, we do warn it, so dev will know something bad (but not super critical) is happening.

frb_example/pure_dart/rust/src/api.rs
rogurotus: The idea is to get the raw Dart_Handle to use in some other Dart_Handle api.

We can make it `unsafe` for native platform and remove it from the example

fzyzcjy: I see. 

uber-nit: Shall we name it like `Box::into_raw`, since I do not expect "unwrap" to suddenly give me a raw data that I have to manage by myself (otherwise leak memory). Feel free to keep the name if the "unwrap" is seen somewhere else in std lib with similar scenario.

It seems to can be safe (instead of unsafe), since `into_raw` is safe as well (it just leaks memory if forget to dispose, so Rust style thinks it is safe).

I am thinking about `Dart_DeletePersistentHandle_DL_Trampolined`. Is it possible we mimic Box, i.e. only expose `DartOpaque::from_raw` function, and do not expose `Dart_DeletePersistentHandle_DL_Trampolined`. Then, when user wants to delete the handle, they just `DartOpaque::from_raw`, and let DartOpaque object be dropped, then DeletePersistHandle is auto called inside. This is just like how we treat raw pointers: We do `Box::from_raw(raw_pointer)` when we want to drop it. Of course this is just one way, feel free to discuss if you have other ways that is commonly used in Rust!

frb_example/with_flutter/lib/bridge_generated.io.dart
rogurotus: what about this:

give a log about the drop when the port is closed.
save statically leaked pointers and return an error with leaked pointers during the current (next) api call. on the frb side, we clear the memory and return the error further (without pointers)

thus there will be no errors if the api is closed and the api is used without dart opaque.
memory will not leak. what will change - statically store leaked pointers and add handling of such an error on the frb side

fzyzcjy: > save statically leaked pointers and return an error with leaked pointers during the current (next) api call. on the frb side, we clear the memory and return the error further (without pointers)

A slight paraphrase, hopefully making the idea clearer: during any api call, make an extra check whether "leaked pointers" pool is non-empty. If so, trigger cleaning procedure, since we are on the main thread and can safely perform the operation.

fzyzcjy: One worry: we need to have one more check (maybe atomic read) for each and every frb call.

But it seems not to be a problem, since we may be able to make DartOpaque an optional feature if someone thinks it is too slow. Anyway the slowness may not appear *now* at all, since the current implementation is sub-optimal and there should exist many other places which can be improved firstly. So no worries.

rogurotus: >A slight paraphrase, hopefully making the idea clearer: during any api call, make an extra check whether "leaked pointers" pool is non-empty. If so, trigger cleaning procedure, since we are on the main thread and can safely perform the operation.

in general, yes, but the idea was precisely at the end of the call, because by this moment all dartOpaque will be deleted (except for the exotic case when we gave dartOpaque to another thread inside the api execution)

fzyzcjy: > except for the exotic case 

IMHO we cannot detect this and this can happen

rogurotus: but that's okay - the next call will return an error and clear the memory.
Well, if there is no next call, the logs will report that there was an unsuccessful cleaning attempt.

fzyzcjy: Seems OK, but as I do not have much brain power for this currently (quite busy reading papers), I cannot say this method LGTM now. Anyway good discussions!

frb_rust/src/dart_api_dl/trampoline.c
```diff
@@ -0,0 +1,52 @@
+// Trampolines to Dynamically Linked Dart API.
```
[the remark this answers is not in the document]
rogurotus: we only need the initialization function `Dart_InitializeApiDL` (the rest can be reused from dart_sys).

frb_rust/src/dart_api_dl/include/internal/dart_api_dl_impl.h
```diff
@@ -0,0 +1,21 @@
+/*
```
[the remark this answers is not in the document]
rogurotus: This folder (`frb_rust/src/dart_api_dl/include/`) contains a copy of Dart SDK https://github.com/dart-lang/sdk/tree/master/runtime/include folder.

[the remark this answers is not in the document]
rogurotus: only dart_api_dl and its dependencies are used.

frb_rust/src/dart_api_dl/include/dart_embedder_api.h
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
rogurotus: This folder (frb_rust/src/dart_api_dl/include/) contains a copy of Dart SDK https://github.com/dart-lang/sdk/tree/master/runtime/include folder. only dart_api_dl and its dependencies are used.

frb_rust/src/dart_api_dl/include/dart_api_dl.c
```diff
@@ -0,0 +1,59 @@
+/*
```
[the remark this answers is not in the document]
rogurotus: yes, we must, because we compile it

frb_rust/Cargo.toml
rogurotus: dart_sys creates bindigs to `dart_api.h` and we need function `Dart_InitializeApiDL`  from `dart_api_dl.h`.

frb_rust/src/dart_api_dl/include/dart_api.h
rogurotus: due to dependencies we use all files except:
- dart_embedder_api.h
- dart_tools_api.h
- dart_io_api.h
- analyze_snapshot_api.h (dart api 2.1)

it's also easier to copy the entire dart-sdk/includes folder than to keep track of all the dependencies

frb_example/with_flutter/lib/bridge_generated.io.dart
rogurotus: from dart_api_dl.h
 ```
 * All symbols are postfixed with _DL to indicate that they are dynamically
 * linked and to prevent conflicts with the original symbol.
 *
 * Link `dart_api_dl.c` file into your library and invoke
 * `Dart_InitializeApiDL` with `NativeApi.initializeApiDLData`.
 ```
 made it so that initialization is called once

frb_example/with_flutter/lib/bridge_generated.io.dart
rogurotus: fixed

frb_example/with_flutter/rust/src/bridge_generated.io.rs
rogurotus: fixed

frb_example/pure_dart/rust/src/api.rs
rogurotus:
```
#[derive(Debug)]
pub struct DartHandleWrap(Option<Dart_PersistentHandle>);

impl DartHandleWrap {
    pub fn into_raw(mut self) -> Dart_PersistentHandle {
        self.0.take().unwrap()
    }
}

impl Drop for DartHandleWrap {
    fn drop(&mut self) {
        if let Some(inner) = self.0 {
            unsafe { Dart_DeletePersistentHandle_DL_Trampolined(inner) }
        }
    }
}
```

frb_codegen/src/commands.rs
```diff
@@ -226,6 +226,7 @@ fn cbindgen(
             "stdint.h".to_string(),
             "stdlib.h".to_string(),
         ],
+        after_includes: Some("typedef struct _Dart_Handle* Dart_Handle;".to_owned()),
```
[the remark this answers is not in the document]
rogurotus: added comment
```
        // copied from: dart-sdk/dart_api.h
        // used to convert Dart_Handle to Object.
```

frb_rust/src/dart_api_dl/include/dart_api_dl.c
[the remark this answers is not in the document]
fzyzcjy: Hmm, could you please add a link (e.g. point to official Dart doc) as a comment to somewhere in the codebase about how to use dart_api_dl.c? At a first glance, I feel it is buggy since if two third-party libs compile the same `.c` there can be conflicts etc.

frb_rust/src/dart_api_dl/include/dart_embedder_api.h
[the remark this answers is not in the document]
fzyzcjy: Get it.

frb_rust/src/dart_api_dl/include/internal/dart_api_dl_impl.h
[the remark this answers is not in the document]
fzyzcjy: Get it (as above)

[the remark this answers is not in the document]
fzyzcjy: So it is quite weird IMHO. Maybe we should at least file an issue at Dart repo, asking why an end user should include "internal" folders. Then after they give positive reply saying it is correct, we can paste the link into this repository's code comments, so future readers are convinced.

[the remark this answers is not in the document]
rogurotus: from dart_api_dl.h
 ```
 * All symbols are postfixed with _DL to indicate that they are dynamically
 * linked and to prevent conflicts with the original symbol.
 *
 * Link `dart_api_dl.c` file into your library and invoke
 * `Dart_InitializeApiDL` with `NativeApi.initializeApiDLData`.
 ```

frb_rust/src/dart_api_dl/include/dart_api_dl.c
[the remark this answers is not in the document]
rogurotus: from dart_api_dl.h
 ```
 * All symbols are postfixed with _DL to indicate that they are dynamically
 * linked and to prevent conflicts with the original symbol.
 *
 * Link `dart_api_dl.c` file into your library and invoke
 * `Dart_InitializeApiDL` with `NativeApi.initializeApiDLData`.
 ```
 I will search further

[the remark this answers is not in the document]
rogurotus: Theoretically, the error can only be when calling Dart_InitializeApiDL multiple times, but this method returns 0 if ok, which we check. Also I couldn't find any restrictions in the comments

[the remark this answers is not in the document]
fzyzcjy: Hmm I have not written C code for a while, but IIRC, "link" does not mean "compile"?

* "compile dart_api_dl.c" := compile the source code
* "link dart_api_dl.c" := do not compile that .c file; instead, only "link" it.

For example, we commonly say "link the `libc` library with our Rust or C code". It does not mean "download and compile the C code of the whole libc", but instead means "just fetch something like libc.a and link that precompiled thing"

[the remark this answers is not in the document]
fzyzcjy: > At a first glance, I feel it is buggy since if two third-party libs compile the same .c there can be conflicts etc.
> the error can only be when calling Dart_InitializeApiDL multiple times

Well I do not mean this. For simplicity, say there are two bare-metal C libraries and they both contain dart_api_dl.c to compile. Then, they will both expose the symbol Dart_InitializeApiDL in their .so files. Therefore, at runtime, the same symbol can come from two places. This seems to be undefined behavior or bad behavior IIRC.

frb_rust/src/dart_api_dl/include/internal/dart_api_dl_impl.h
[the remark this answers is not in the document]
fzyzcjy: (see comments above)

frb_rust/src/dart_api_dl/include/dart_api.h
fzyzcjy: (see above, where I guess we do not need that .c thus no need for a lot of .h)

frb_rust/src/dart_api_dl/include/dart_api_dl.c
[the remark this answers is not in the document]
fzyzcjy: Hmm I see https://github.com/dart-lang/sdk/issues/47356#issuecomment-1056770450, so seems the original approach is correct. But I am still quite confused about it:

1. Why we include "internal" things
2. This worry https://github.com/fzyzcjy/flutter_rust_bridge/pull/853#discussion_r1034665554

As said above, I have not written C code for a while, so please correct me if I am wrong!

frb_example/with_flutter/lib/bridge_generated.io.dart
fzyzcjy: I see. Just a remark that, the `FlutterRustBridgeExamplePlatform` constructor can be called *multiple times* because of the following scenarios for example:

1. hot restart flutter app
2. users (for whatever reason) create two FlutterRustBridgeExamplePlatform instances

frb_example/pure_dart/rust/src/api.rs
fzyzcjy: LGTM with nits: What about adding a simple line of comment mentioning "Option" is used for dropping

frb_rust/src/dart_api_dl/include/dart_api_dl.c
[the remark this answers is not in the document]
rogurotus: we can just change dart_api_dl.
instead of `Dart_InitializeApiDL`, `Dart_HandleFromPersistent` symbols, we will compile as `Dart_InitializeApiDL_FRB` (basically just change the macro). We will also remove all unnecessary.

also calling Dart_InitializeApiDL_FRB and third party Dart_InitializeApiDL won't conflict in any way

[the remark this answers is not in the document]
fzyzcjy: Ok I did a bit more research and looks like this (weird) way looks like what people really uses. Then maybe we just keep with your original simplest approach.

> we will compile as Dart_InitializeApiDL_FRB

The worry 2 is not completely removed IMHO, since some people mention they have two crates using frb in one app. But given we have not supported that case it looks acceptable.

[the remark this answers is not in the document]
fzyzcjy: By the way, I still suggest creating an issue in Dart repo, such that if the official team replies, we can know more about the problem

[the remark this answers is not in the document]
fzyzcjy: Btw, here is an example about symbol duplication (though very different from our cases): https://stackoverflow.com/questions/55191553/what-actually-happens-when-two-shared-libraries-define-the-same-symbol

> I recently encountered a crash issue when I linked two shared libraries (both made by myself) together. I eventually found it was because of one source file duplicated between the two files. In that source file a global std::vector was defined (in fact a static member of a class), and it ended up with being freed twice -- one by each library.

Again I am not expert in C given lack of using it recently, so feel free to point out my problems!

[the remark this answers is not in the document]
rogurotus: https://github.com/dart-lang/sdk/issues/47356#issuecomment-1331894796
If I understand correctly, then in our case everything is fine (if i asked what you mean)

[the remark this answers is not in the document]
rogurotus: I additionally checked in cases where different libraries call initApiDl and the case when there is a frb dependency of different versions. It didn't give me any errors.

frb_example/with_flutter/lib/bridge_generated.io.dart
rogurotus: one-time static initialization

frb_example/with_flutter/lib/bridge_generated.io.dart
rogurotus: I think really it is possible to give only warning log.

book/src/SUMMARY.md
```diff
@@ -75,7 +75,7 @@
 - [Overview](contributing/overview.md)
 - [Overall design](contributing/design.md)
 - [Submodule implementations](contributing/submodule.md)
-  - [Opaque type safety](contributing/opaque_type_safety.md)
+  - [Rust opaque type safety](contributing/rust_opaque_type_safety.md)
```
fzyzcjy, line 78:
```suggestion
  - [Rust opaque type safety](contributing/rust_opaque_type_safety.md)
  - [Dart opaque type safety](contributing/dart_opaque_type_safety.md)
```

book/src/contributing/dart_opaque_type_safety.md
```diff
@@ -0,0 +1,86 @@
+# Dart opaque type safety
+
+
+## Ownership and GC
+
+From the moment the opaque type is passed, Rust will own the persistent representation of the dart object (`Dart_PersistentHandle` or `JsValue`).
```
fzyzcjy, line 6:
```suggestion
From the moment the opaque type is passed, Rust will own a persistent representation of the dart object (`Dart_PersistentHandle` or `JsValue`).
```

book/src/feature/lang_opaque.md
```diff
@@ -1,18 +1,18 @@
 # Arbitrary types (opaque)
 
-Any Rust type, even if it is not supported using features of this library, can be used in Dart. This is done by wrapping it with `Opaque`.
+Any Rust type, even if it is not supported using features of this library, can be used in Dart. This is done by wrapping it with `RustOpaque`. In addition, Rust can get opaque Dart type in the `DartOpaque` wrap.
```
fzyzcjy, line 3:
```suggestion
On one hand, any Rust type, even if it is not supported using features of this library, can be used in Dart. This is done by wrapping it with `RustOpaque`.

On the other hand, Rust can get arbitrary Dart type as well using the `DartOpaque` opaque wrapper.
```

book/src/feature/lang_simple.md
```diff
@@ -12,7 +12,7 @@ Here is a brief glance showing what the code generator can generate (non-exhaust
 | [`enum { A(..) }`](lang_enum.md)                  | `@freezed class`            |
 | [`use ...`](lang_external.md)                     | act normally                |
 | [`Option<T>`](lang_option.md)                     | `T?`                        |
-| [Arbitrary types (opaque)](lang_opaque.md)        | `Opaque`                    |
+| [Arbitrary types (opaque)](lang_opaque.md)        | `RustOpaque`, `DartOpaque`  |
```
fzyzcjy, line 15:
```suggestion
| [Arbitrary types (opaque)](lang_opaque.md)        | `RustOpaque`                |
| `DartOpaque`                                      | [Arbitrary types (opaque)](lang_opaque.md) |
```

book/src/feature/lang_opaque.md
```diff
@@ -1,18 +1,18 @@
...
 
-The Dart opaque objects should be disposed *manually*, though it will also be disposed when it is GCed, that is discouraged, due to [suggestions by Dart team](https://github.com/fzyzcjy/flutter_rust_bridge/issues/775#issuecomment-1274635037). Think of it just like a lot of Flutter objects that we are familiar with, such as `ui.Image` - we have to manually dispose them as well.
+The Rust opaque objects in Dart should be disposed *manually*, though it will also be disposed when it is GCed, that is discouraged, due to [suggestions by Dart team](https://github.com/fzyzcjy/flutter_rust_bridge/issues/775#issuecomment-1274635037). Think of it just like a lot of Flutter objects that we are familiar with, such as `ui.Image` - we have to manually dispose them as well. The Dart opaque objects are always deleted on the parent Dart thread.
```
fzyzcjy, line 5: uber-nit: Shall we split DartOpaque and RustOpaque into two separate files (webpages)

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -327,7 +328,9 @@ fn generate_dart_implementation_body(spec: &DartApiSpec, config: &Opts) -> Acc<D
         ),
         io: format!(
             "class {plat} extends FlutterRustBridgeBase<{wire}> {{
-                {plat}(ffi.DynamicLibrary dylib) : super({wire}(dylib));",
+                {plat}(ffi.DynamicLibrary dylib) : super({wire}(dylib)) {{
+                    DartApiDl.initApi(dylib);
```
fzyzcjy, line 332: As mentioned earlier, would be great to have lazy initialization, such that people not using Dart Opaque will not have any cost at all.

(not reviewed whole code yet, so if it already have lazy init, ignore this comment)

frb_codegen/src/generator/rust/ty_dart_opaque.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::ir::*;
+use crate::target::Acc;
+use crate::type_rust_generator_struct;
+use crate::utils::BlockIndex;
+
+use super::{ExternFuncCollector, NO_PARAMS};
+
+type_rust_generator_struct!(TypeDartOpaqueGenerator, IrTypeDartOpaque);
+
+impl TypeRustGeneratorTrait for TypeDartOpaqueGenerator<'_> {
```
fzyzcjy, line 13:
```suggestion
impl TypeDartGeneratorTrait for TypeDartOpaqueGenerator<'_> {
```

frb_codegen/src/generator/rust/ty_dart_opaque.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn wire2api_body(&self) -> crate::target::Acc<Option<String>> {
+        Acc { io: Some(
+    "DartOpaque::new(self.handle as _, self.port)".to_owned(),
+), wasm: Some("let arr = self.dyn_into::<JsArray>().unwrap(); DartOpaque::new(arr.get(0), arr.get(1))".to_owned()), ..Default::default() }
```
fzyzcjy, line 17: uber-nit: format the code

frb_codegen/src/generator/rust/ty_enum.rs
```diff
@@ -273,7 +273,7 @@ impl TypeRustGeneratorTrait for TypeEnumRefGenerator<'_> {
         fn init_of(ty: &IrType) -> String {
             if ty.rust_wire_is_pointer(Io) {
                 "core::ptr::null_mut()".to_owned()
-            } else if ty.is_opaque() {
+            } else if ty.is_opaque() || ty.is_dart_opaque() {
```
fzyzcjy, line 276: uber-nit: is_opaque -> is_rust_opaque

frb_codegen/src/ir/ty_dart_opaque.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub struct IrTypeDartOpaque;
+
+impl IrTypeTrait for IrTypeDartOpaque {
+    fn visit_children_types<F: FnMut(&IrType) -> bool>(&self, _f: &mut F, _ir_file: &IrFile) {
+        // do nothing.
+    }
+
+    fn safe_ident(&self) -> String {
+        "DartObject".to_owned()
```
fzyzcjy, line 12: uber-nit: Sometimes we say DartOpaque, sometimes we say DartObject. Shall we unify them?

frb_dart/lib/src/basic.dart
```diff
@@ -25,6 +33,21 @@ abstract class FlutterRustBridgeBase<T extends FlutterRustBridgeWireBase> {
   @protected
   final T inner;
 
+  late final _dropPort = _initDropPort();
+  NativePortType get dropPort => _dropPort.sendPort.nativePort;
+
+  ReceivePort _initDropPort() {
+    var port = broadcastPort(_DropIdPortGenerator.instance.create());
+    port.listen((message) {}).onData(((response) {
```
fzyzcjy, line 41: What about

`port.listen((message) { ... the body });`

frb_dart/lib/src/ffi/io.dart
```diff
@@ -19,6 +19,24 @@ extension StoreDartPostCObjectExt on FlutterRustBridgeWireBase {
...
+
+  static void initApi(ffi.DynamicLibrary dylib) {
+    var code = _initCode ?? _internal(dylib);
+    if (code != 0) {
+      throw 'Failed to initialize Dart API. Code: $_initCode';
+    }
+  }
+
+  static int _internal(ffi.DynamicLibrary dylib) {
+    _initCode = dylib.lookupFunction<ffi.IntPtr Function(ffi.Pointer<ffi.Void>),
```
fzyzcjy, line 33: nit: Is it possible we use automatically generated signatures instead of manually writing down the lookupFunction?

frb_dart/lib/src/ffi/stub.dart
```diff
@@ -19,6 +20,19 @@ abstract class FlutterRustBridgeWireBase {
   // ignore: non_constant_identifier_names
   void store_dart_post_cobject(DartPostCObject ptr) {}
 
+  // ignore: non_constant_identifier_names
+  Object get_dart_object(int ptr) {
+    return '';
```
fzyzcjy, line 25: Given that this base class is abstract, shall we use `Object get_dart_object();` i.e. no body at all

same for all other methods

frb_dart/lib/src/ffi/web.dart
```diff
@@ -83,10 +83,34 @@ external WasmModule? get _noModules;
...
+  }
+
+  // ignore: non_constant_identifier_names
+  void drop_dart_object(int ptr) {
+    dropDartObject(ptr);
+  }
+
+  // ignore: non_constant_identifier_names
+  int new_dart_opaque(Object obj, NativePortType port) {
+    return 0;
```
fzyzcjy, line 112: Interesting, so this is never called? If so, maybe we can throw, to avoid misuse

fzyzcjy (commented):
half-reviewed, good job! mostly are nits with modifications included

.github/workflows/code_generator.yml
```diff
@@ -56,6 +56,10 @@ jobs:
         working-directory: ./frb_example/with_flutter
         run: flutter pub get
 
+      - name: Check DartApi version.
+        run: wget https://github.com/dart-lang/sdk/blob/master/runtime/include/dart_version.h?raw=true 
+          -O ./frb_rust/src/dart_api_dl/include/dart_version.h && git diff --exit-code
```
fzyzcjy, line 61: nit: would be great to check the whole `include` folder

e.g. using clone with one commit https://stackoverflow.com/questions/31278902/how-to-shallow-clone-a-specific-commit-with-depth-1

frb_example/pure_dart/dart/lib/main.dart
fzyzcjy: Hmm may I know where is it implemented?

frb_rust/src/ffi/dart_api.rs
```diff
@@ -0,0 +1,45 @@
...
+/// Initializes usage of Dynamically Linked Dart API.
+///
+/// # Safety
+///
+/// Intended to be called ONLY with [`NativeApi.initializeApiDLData`][1] from
+/// Dart.
+///
+/// [1]: https://api.dart.dev/dart-ffi/NativeApi/initializeApiDLData.html
+#[no_mangle]
+pub unsafe extern "C" fn init_frb_dart_api_dl(obj: ptr::NonNull<c_void>) -> isize {
```
fzyzcjy, line 43: uber-nit: looks like we do not need `NonNull` here, because `*mut c_void` will take the same space. In addition, NonNull is said to be (https://doc.rust-lang.org/stable/std/ptr/struct.NonNull.html) more dangerous in case of misuse, so may be good to use the `*mut`

frb_rust/src/ffi/io.rs
```diff
@@ -31,3 +38,89 @@ mod tests {
...
+    /// only to be sent to the Dart side.
+    pub unsafe fn new_non_dropable(handle: Dart_Handle) -> Self {
+        Self {
+            inner: DartHandleWrap(Some(handle)),
+            drop_port: None,
+        }
+    }
+
+    pub fn into_raw(&mut self) -> Dart_PersistentHandle {
+        self.inner.0.take().unwrap()
```
fzyzcjy, line 116:
```suggestion
        self.inner.into_raw()
```

frb_rust/src/ffi/io.rs
```diff
@@ -31,3 +38,89 @@ mod tests {
...
+        }
+    }
+
+    /// Creates a new wrapper for send to dart.
+    ///
+    /// # Safety
+    ///
+    /// Must be used for Dart objects created on the current thread
+    /// only to be sent to the Dart side.
+    pub unsafe fn new_non_dropable(handle: Dart_Handle) -> Self {
```
fzyzcjy, line 108: Btw is Dart_Handle same as Dart_PersistentHandle? If so maybe we should not mix them to make it clear

frb_rust/src/ffi/io.rs
```diff
@@ -31,3 +38,89 @@ mod tests {
...
+#[derive(Debug)]
+pub struct DartOpaqueBase {
+    inner: DartHandleWrap,
+    drop_port: Option<MessagePort>,
+}
+
+impl DartOpaqueBase {
+    pub fn new(handle: Dart_PersistentHandle, port: MessagePort) -> Self {
+        Self {
+            inner: DartHandleWrap(Some(handle)),
```
fzyzcjy, line 97:
```suggestion
            inner: DartHandleWrap::from_raw(handle),
```

frb_rust/src/ffi/io.rs
```diff
@@ -31,3 +38,89 @@ mod tests {
...
+
+    /// Creates a new wrapper for send to dart.
+    ///
+    /// # Safety
+    ///
+    /// Must be used for Dart objects created on the current thread
+    /// only to be sent to the Dart side.
+    pub unsafe fn new_non_dropable(handle: Dart_Handle) -> Self {
+        Self {
+            inner: DartHandleWrap(Some(handle)),
```
fzyzcjy, line 110:
```suggestion
            inner: DartHandleWrap::from_raw(handle),
```

fzyzcjy (commented):
Finished reviewing. Great job!

frb_example/with_flutter/lib/bridge_generated.io.dart
fzyzcjy: Have you tried the hot-restart-flutter case, because in that case, it is initialized multiple times (even though we use static variables etc), because after each hot restart, the Dart vm starts freshly, while state in native code is kept. So I am not sure whether that is OK.

frb_example/pure_dart/dart/lib/main.dart
fzyzcjy: (same as above)

frb_dart/lib/src/basic.dart
rogurotus: I will double check, but in my opinion it does not work (for some unknown reason)

frb_rust/src/ffi/io.rs
rogurotus: Dart_Handle means the object can be cleaned up by the GC while Dart_Persistent_Handle is not.

frb_dart/lib/src/ffi/stub.dart
rogurotus: it's because of dart analyzer. he does not see that we are importing web.dart from the implementation and consider that there are not enough symbols.

frb_rust/src/ffi/io.rs
fzyzcjy: Btw where is new_non_dropable used? I have done a quick search but there is no result

image

rogurotus: not used in tests. this function is for cases when we call dart from rust and want to return the received Dart_Handle to dart via frb (for direct ffi dart compatibility)

frb_example/with_flutter/lib/bridge_generated.io.dart
rogurotus: I tested and didn't get multiple initialization

frb_rust/src/ffi/io.rs
fzyzcjy: Hmm, then maybe we should add a test, otherwise it can regress in the future

fzyzcjy: Btw, how do we call Dart from Rust now? IMHO we can only do the reverse now

frb_dart/lib/src/ffi/stub.dart
fzyzcjy: Weird. Could you please take a screenshot of its complain?

rogurotus: https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/3600007687/jobs/6064198006
Analyzing dart...

  error - lib/bridge_generated.web.dart:112[8](https://github.com/rogurotus/flutter_rust_bridge/actions/runs/3600007442/jobs/6064197497#step:9:9):7 - Missing concrete implementations of 'FlutterRustBridgeWireBase.drop_dart_object', 'FlutterRustBridgeWireBase.free_WireSyncReturnStruct', 'FlutterRustBridgeWireBase.get_dart_object', 'FlutterRustBridgeWireBase.new_dart_opaque', and 1 more. Try implementing the missing methods, or make the class abstract. - non_abstract_class_inherits_abstract_member

frb_rust/src/ffi/io.rs
rogurotus: It is possible to write registration of dart functions in rust.

fzyzcjy: Sure, but how can we call it? IMHO we only have DartOpaque and RustOpaque, which is not callable

frb_dart/lib/src/ffi/stub.dart
rogurotus: the analyzer still responds the same way

frb_dart/lib/src/basic.dart
rogurotus: Hmm, it seems to work, but it may be unstable. let it be `listen`

frb_rust/src/ffi/io.rs
fzyzcjy: "on rust side use function pointer" - looks like we have to execute some unsafe code?

It is quite interesting to see Rust can call Dart callbacks. We should test it and document it!

rogurotus: new_non_dropable deleted

fzyzcjy: Looking forward to the next PR!

frb_example/pure_dart/dart/lib/main.dart
rogurotus: is that not what you need?

```
    test('drop', () async {
      expect(await api.asyncAcceptDartOpaque(opaque: _createLargeList(mb: 200)), 'async test');
      expect(api.syncAcceptDartOpaque(opaque: _createLargeList(mb: 200)), 'test');
    });
 ```

Comment (fzyzcjy) [reply frb_example/pure_dart/dart/lib/main.dart:696]: Well, we should monitor whether the list is GCed. See "usage_example" in https://gist.github.com/fzyzcjy/e68c375643d7c77942cdc8fb5f01de18#file-usage_example-dart-L80, especially https://gist.github.com/fzyzcjy/e68c375643d7c77942cdc8fb5f01de18#file-usage_example-dart-L88 where we check Weak ref is null

File: frb_dart/lib/src/ffi/stub.dart (hunk 54 above)
Comment (fzyzcjy) [reply frb_dart/lib/src/ffi/stub.dart]: Hmm looks like the problem is, in web.dart, FlutterRustBridgeExampleSingleBlockTestWire really does not implement those methods.

Thus, seems that:

1. FlutterRustBridgeWasmWireBase should implement those method (by throwing unimplemented error etc if you think it is not needed)
2. FlutterRustBridgeWireBase should be abstract methods

image

Comment (rogurotus) [reply frb_dart/lib/src/ffi/stub.dart]: the implementation from web.dart is used.

this is due to the fact that native uses the generated version, and web uses the implementation from web.dart

File: frb_example/pure_dart/dart/lib/main.dart (hunk 24 above)
Comment (rogurotus) [reply frb_example/pure_dart/dart/lib/main.dart:697]: Will this be enough?
   
   ```
    test('GC', () async {
      var vmService = await VmServiceUtil.create();
      Uint8List? strongRef = _createLargeList(mb: 300);
      final weakRef = WeakReference(strongRef);
      await api.setStaticDartOpaque(opaque: strongRef);
      strongRef = null;

      await vmService.gc();
      await Future<void>.delayed(const Duration(milliseconds: 10));
      expect(weakRef.target , isNotNull);

      await api.dropStaticDartOpaque();
      await vmService.gc();
      await Future<void>.delayed(const Duration(milliseconds: 10));
      expect(weakRef.target , isNull);

      vmService.dispose();
    });
  ```

Comment (rogurotus) [reply frb_example/pure_dart/dart/lib/main.dart:697]: on the web platform: Unsupported operation: new RawReceivePort
on the valgring: Exception: Cannot find serverUri for VmService. Ensure you run like `dart run --enable-vm-service path/to/your/file.dart`

is it worth testing this particular case without valgrind?
(dart run --enable-vm-service locally this test worked)

Comment (fzyzcjy) [reply frb_example/pure_dart/dart/lib/main.dart:697]: LGTM as long as tests pass. For a more robust test, maybe we can refer to my example and add a for loop to retry.

We may temporarily ignore web if it is too hard.

> is it worth testing this particular case without valgrind

IMHO yes

File: frb_dart/lib/src/ffi/stub.dart (hunk 54 above)
Comment (fzyzcjy) [reply frb_dart/lib/src/ffi/stub.dart]: I see. Without making it more complicated, this LGTM

File: frb_codegen/src/others.rs [hunk 62]
```diff
@@ -47,7 +55,8 @@ pub fn modify_dart_wire_content(content_raw: &str, dart_wire_class_name: &str) -
     let content = content_raw.replace(
         &format!("class {} {{", dart_wire_class_name),
         &format!(
-            "class {} implements FlutterRustBridgeWireBase {{",
+            "class {} implements FlutterRustBridgeWireBase {{
+            late final dartApi = DartApiDl(init_frb_dart_api_dl);",
```
fzyzcjy, line 59: uber-nit: Btw shall it be @internal or private?

fzyzcjy (commented):
Great job and I am ready to merge! Only two uber-nits

frb_codegen/src/others.rs
rogurotus: Yes, this is an `internal` field, but end users don't have access to it anyway.
(added attribute `internal`)

fzyzcjy (approved):

fzyzcjy:
Excellent job!

PR #875, opened 2022-12-06
support type impl trait as enum

huang12zheng:

## Checklist

- [ ] CI is passing.

## Remark for PR creator
As disscuess said, the way I automatically generate enums is ugly,
But an enum may be acceptable instead of an impl.

If something could pass, I can probably start with the pr with that part.
To be honest, I have no confidence in this pr

### DIFF
--- a/frb_codegen/src/ir/file.rs
+++ b/frb_codegen/src/ir/file.rs
@@ -1,16 +1,24 @@
+use crate::source_graph::Impl;
 use crate::target::Target;
 use crate::utils::mod_from_rust_path;
 use crate::{generator, ir::*, Opts};
 use std::collections::{HashMap, HashSet};
 
+pub type Pool<T> = HashMap<String, T>;
+pub type ImplPool = Pool<Vec<Impl>>;
+
 pub type IrStructPool = HashMap<String, IrStruct>;
 pub type IrEnumPool = HashMap<String, IrEnum>;
+pub type IrImplTraitPool = HashSet<IrTypeImplTrait>;
 
 #[derive(Debug, Clone)]
 pub struct IrFile {
     pub funcs: Vec<IrFunc>,
     pub struct_pool: IrStructPool,
     pub enum_pool: IrEnumPool,
+    pub src_impl_pool: ImplPool,
+    pub parsed_impl_traits: IrImplTraitPool,
+
     pub has_executor: bool,
 }
 
@@ -62,7 +70,11 @@ impl IrFile {
                 let contains = seen_idents.contains(&ident);
                 if !contains {
                     seen_idents.insert(ident);
-                    ans.push(ty.clone());
+                    ans.push(match ty {
+                        IrType::ImplTrait(i) => i.clone().to_enum_ir_type(),
+
+                        any => any.clone(),
+                    });
                 }
                 contains
             },
--- a/frb_codegen/src/ir/ty_type_impl_trait.rs
+++ b/frb_codegen/src/ir/ty_type_impl_trait.rs
@@ -1,0 +1,86 @@
+use crate::ir::*;
+use crate::target::Target;
+use convert_case::{Case, Casing};
+use syn::Ident;
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub struct IrTypeImplTrait {
+    pub trait_bounds: Vec<String>,
+}
+
+impl IrTypeImplTrait {
+    pub fn join(&self) -> String {
+        self.trait_bounds.join("_")
+    }
+
+    pub fn to_enum(&self) -> String {
+        format!("{}Enum", self.join()).to_case(Case::Pascal)
+    }
+
+    pub fn to_enum_ir_type(&self) -> IrType {
+        IrType::EnumRef(IrTypeEnumRef {
+            name: self.to_enum(),
+        })
+    }
+}
+
+impl std::fmt::Display for IrTypeImplTrait {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        fmt.write_str(&self.join())
+    }
+}
+
+impl IrTypeImplTrait {
+    pub fn new(raw: Vec<Ident>) -> IrTypeImplTrait {
+        IrTypeImplTrait {
+            trait_bounds: raw.iter().map(|e| e.to_string()).collect(),
+        }
+    }
+    pub fn new2(raw: Vec<String>) -> IrTypeImplTrait {
+        IrTypeImplTrait { trait_bounds: raw }
+    }
+}
+
+impl IrTypeTrait for IrTypeImplTrait {
+    fn visit_children_types<F: FnMut(&super::IrType) -> bool>(
+        &self,
+        f: &mut F,
+        ir_file: &super::IrFile,
+    ) {
+        // due to I would remove `mod bridge_generated` in lib.rs, first get ir file, can't find ImplTraitEnum
+        if ir_file.enum_pool.contains_key(&self.to_enum()) {
+            IrTypeEnumRef {
+                name: self.to_enum(),
+            }
+            .visit_children_types(f, ir_file)
+        }
+    }
+
+    fn safe_ident(&self) -> String {
+        self.to_enum().to_case(Case::Snake)
+    }
+
+    fn dart_api_type(&self) -> String {
+        self.rust_api_type()
+    }
+
+    fn dart_wire_type(&self, target: crate::target::Target) -> String {
+        if let Target::Wasm = target {
+            "List<dynamic>".into()
+        } else {
+            self.rust_wire_type(target)
+        }
+    }
+
+    fn rust_api_type(&self) -> String {
+        self.to_enum()
+    }
+
+    fn rust_wire_type(&self, target: crate::target::Target) -> String {
+        if let Target::Wasm = target {
+            "JsValue".into()
+        } else {
+            format!("wire_{}", self.to_enum())
+        }
+    }
+}
--- a/frb_codegen/src/parser/ty.rs
+++ b/frb_codegen/src/parser/ty.rs
@@ -1,6 +1,9 @@
 use std::collections::{HashMap, HashSet};
+
 use std::string::String;
 
+use itertools::Itertools;
+use quote::__private::Span;
 use syn::*;
 
 use crate::ir::IrType::*;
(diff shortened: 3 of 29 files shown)


frb_codegen/src/ir/ty_type_impl_trait.rs
```diff
@@ -0,0 +1,86 @@
...
+use convert_case::{Case, Casing};
+use syn::Ident;
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub struct IrTypeImplTrait {
+    pub trait_bounds: Vec<String>,
+}
+
+impl IrTypeImplTrait {
+    pub fn join(&self) -> String {
```
fzyzcjy, line 12: uber-nit: not very sure whether it is a good idea to have "IrTypeImplTrait.join", which does not seem very clear

frb_codegen/src/ir/ty_type_impl_trait.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+impl IrTypeImplTrait {
+    pub fn new(raw: Vec<Ident>) -> IrTypeImplTrait {
+        IrTypeImplTrait {
+            trait_bounds: raw.iter().map(|e| e.to_string()).collect(),
+        }
+    }
+    pub fn new2(raw: Vec<String>) -> IrTypeImplTrait {
```
fzyzcjy, line 39:
```suggestion
    pub fn new_raw(raw: Vec<String>) -> IrTypeImplTrait {
```

frb_codegen/src/ir/ty_type_impl_trait.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+impl IrTypeTrait for IrTypeImplTrait {
+    fn visit_children_types<F: FnMut(&super::IrType) -> bool>(
+        &self,
+        f: &mut F,
+        ir_file: &super::IrFile,
+    ) {
+        // due to I would remove `mod bridge_generated` in lib.rs, first get ir file, can't find ImplTraitEnum
```
fzyzcjy, line 50: Sorry I cannot quite get it...

frb_codegen/src/parser/ty.rs
```diff
@@ -1,45 +1,59 @@
 use std::collections::{HashMap, HashSet};
+
 use std::string::String;
 
+use itertools::Itertools;
+use quote::__private::Span;
```
fzyzcjy, line 6: Hmm "private" seems to mean we should not use it. Maybe something like quote::Span?

frb_codegen/src/parser/ty.rs
```diff
@@ -234,6 +283,9 @@ impl<'a> TypeParser<'a> {
                         Some(IrType::Opaque(IrTypeOpaque::from(path.to_string())))
                     }
                     SupportedInnerType::Unit => Some(IrType::Opaque(IrTypeOpaque::new_unit())),
+                    SupportedInnerType::ImplTrait(_) => {
+                        unreachable!("ImplTrait should not be wrapper with Opaque")
```
fzyzcjy, line 287:
```suggestion
                        unreachable!("ImplTrait should not be wrapped with Opaque")
```

frb_codegen/src/ir/ty_type_impl_trait.rs
```diff
@@ -0,0 +1,86 @@
+use crate::ir::*;
+use crate::target::Target;
+use convert_case::{Case, Casing};
+use syn::Ident;
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub struct IrTypeImplTrait {
```
fzyzcjy, line 7: Ok I see the problem: "Ir" is "intermediate representation" and means we should write down code generator for both Dart and Rust. If you want `IrTypeImplTrait` to generate an enum, maybe consider the special `ir_delegate`. For example, the underlying implementation for String is indeed `Vec<u8>`, so you can see how we do it in ir_delegate files

frb_codegen/src/source_graph.rs
```diff
@@ -31,7 +34,7 @@ pub struct Crate {
 }
 
 impl Crate {
-    pub fn new(manifest_path: &str) -> Self {
+    pub fn pre_new(manifest_path: &str) -> (String, PathBuf) {
```
fzyzcjy, line 37: Hmm the several new looks a bit confusing: pre_new vs new vs new_without_resolve. They even looks like doing different things :/

frb_codegen/src/source_graph.rs
```diff
@@ -158,10 +166,41 @@ impl Debug for Enum {
...
+pub struct ImplTrait {
+    pub ident: Ident,
+    pub src: TypeImplTrait,
+    pub bound: TraitBound,
+}
+#[derive(Debug, Clone, Eq)]
+pub struct Impl {
+    pub self_ty: Ident,
+    pub trait_: Ident,
+}
```
fzyzcjy, line 179: nit: would be great to add a few lines of comments discussing what this is. e.g.

```
/// This represents `impl {trait} for {self_ty}`
```

(not sure whether that is the case though, just an example)

frb_example/pure_dart/rust/src/lib.rs
```diff
@@ -1,5 +1,9 @@
+/// impl_trait: |Debug|fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result
+/// handle_opaque: HideData
+/// add_box: ExoticOptionals
```
fzyzcjy, line 3: Hmm maybe we should not have these (auto-generated) lines

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -24,6 +24,9 @@ pub fn simple_adder(a: i32, b: i32) -> i32 {
     a + b
 }
 
+pub fn tt(t: impl Debug) {}
```
fzyzcjy, line 27: would be great to have a more meaningful name for testing

fzyzcjy (commented):
Good job! This is an interesting feature :) Most comments are uber-nits which are trivial to solve

frb_codegen/src/ir/file.rs
```diff
@@ -62,7 +70,11 @@ impl IrFile {
                 let contains = seen_idents.contains(&ident);
                 if !contains {
                     seen_idents.insert(ident);
-                    ans.push(ty.clone());
+                    ans.push(match ty {
+                        IrType::ImplTrait(i) => i.clone().to_enum_ir_type(),
+
+                        any => any.clone(),
+                    });
```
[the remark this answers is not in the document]
huang12zheng: Because logically, I just translated the impl trait to enum, so do you have any more ideas for doing this?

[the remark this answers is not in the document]
fzyzcjy: not sure whether "delegating" (e.g. String deletates to Vec u8) works for this?

frb_codegen/src/parser/ty.rs
huang12zheng: maybe `proc_macro2::Span`
I'm going to add the proc_macro2 dependency to Cargo.tom

frb_codegen/src/source_graph.rs
```diff
@@ -158,10 +166,41 @@ impl Debug for Enum {
...
+    fn partial_cmp(&self, other: &Impl) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl PartialEq for Impl {
+    fn eq(&self, other: &Impl) -> bool {
+        self.self_ty == other.self_ty
+    }
+}
```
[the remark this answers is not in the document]
huang12zheng: I didn't find that, but here,
I just want to sort self_ty. It might be quicker to write it myself.
I would use it automatically, if you still suggest it.
Give me a reply.

> collect_impls_to_vec ' comment is not correct, I correct in my local

frb_example/pure_dart/rust/src/lib.rs
huang12zheng: Although I'd like to...
But I don't know how to solve it yet. I don't think I'll know for a long time

It's hard for me to know the sig of a Trait,
And what types are private,
And whether the Enum is too large  -:/

frb_example/pure_dart/rust/src/lib.rs
```diff
@@ -1,5 +1,9 @@
...
 mod api;
-mod bridge_generated;
+mod bridge_generated; /* AUTO INJECTED BY flutter_rust_bridge. This line may not be accurate, and you can change it according to your needs. */
+mod bridge_generated_bound;
```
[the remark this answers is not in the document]
huang12zheng: I thought it was part of api.rs.
If I add bridge_generated, I have to generate bridge_generated twice.
I also implemented ItmImpl(for trait) for Enum. This can be difficult to deal with.

Looking forward to more ideas.

frb_codegen/src/source_graph.rs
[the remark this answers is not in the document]
fzyzcjy: such handwritten things are error-prone and hard to maintain, thus IMHO we should use the auto generated thing

frb_example/pure_dart/rust/src/lib.rs
fzyzcjy: Then maybe we should move it to `bridge_generated.rs`. In other words, flutter_rust_bridge should not write to anywhere except bridge_generated.rs

fzyzcjy: Oh, do you mean these lines should be written by humans (users), or these are auto generated by flutter_rust_bridge?

If manually written, please consider using the `frb` macro.
If auto generated, see above - we should only write to bridge_generated.rs

frb_example/pure_dart/rust/src/lib.rs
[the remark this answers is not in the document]
fzyzcjy: Not sure what is the problem... Just generate bridge_generated.rs once

frb_codegen/src/ir/ty_type_impl_trait.rs
huang12zheng: I remove it, just use `to_enum`.

frb_example/pure_dart/rust/build.rs
```diff
@@ -26,6 +26,8 @@ fn main() {
     let configs = config_parse(raw_opts);
 
     // generation of rust api for ffi
+    let mut otps = OptArray::new_without_resolve(&configs);
+    otps.run_impl_trait_enum();
```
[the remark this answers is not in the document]
huang12zheng: I want to explain to you,
For other code generation, only one IrFile is needed, that is, an api.rs
But,In my case,
You need all of the api.rs,
In other word, enums are generated for the func input of all apis.

The function can be divided into modules, but the Enum of the function argument is considered to be aggregated together.
So here's a function that looks like this

frb_codegen/src/source_graph.rs
```diff
@@ -615,3 +699,24 @@ fn flatten_use_tree(use_tree: &UseTree) -> Vec<Vec<String>> {
...
+                            scope_impls.push(Impl {
+                                self_ty: self_ty.clone(),
+                                trait_,
+                            });
+                        }
+                    }
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
huang12zheng:
```rs
 if vis {
        attrs
            .iter()
            .flat_map(|a| a.tokens.clone().into_iter())
            .filter_map(|tt_a| {
                if let TokenTree::Group(g) = tt_a {
                    Some(g.stream().into_iter())
                } else {
                    None
                }
            })
            .flatten()
            .filter_map(|tt_g| {
                if let TokenTree::Ident(trait_) = tt_g {
                    Some(Impl {
                        self_ty: self_ty.clone(),
                        trait_,
                    })
                } else {
                    None
                }
            })
            .collect()
    } else {
        vec![]
    }
```

frb_example/pure_dart/rust/src/lib.rs
huang12zheng: frb macro maybe I should choose

huang12zheng: But I need a little more time, or more information

huang12zheng:
Thank you very much for your guidance, but I think I may not be able to complete this for a long time, or not be able to merge effectively.
Finally, sorry for the added noise

fzyzcjy:
@huang12zheng IMHO you are already 90% finished! so it is a pity to give up now :(

Most of the comments are uber-nits (i.e. tiny changes). In addition, if you find something weird, just comment and discuss, not all things need to be implemented as-is.

PR #882, opened 2022-12-09
Make sync mode support whatever types that classical async mode supports

### LINKED ISSUE #881: `SyncReturn` should be transparent
In many scenarios async return is overkill and create a lot of complexity:

* Non-primitive getters and setters
* Constructors
* Iterators
* Other simple/fast calls
* Use of libs that require single threaded execution because of their locking machanisms

Async can be unnecessarily painful, especially when working hand in hand with flutter build functions.

Moreover, this is an opportunity to simplify the code base significantly and facilitate maintainability.

`SyncReturn` should be transparent and support anything that would be returned otherwise (probably `IntoDart`).

Is there a drawback or any reason why `SyncReturn` is serialized to bytes so far?

I tried working with current implementation, working around with sync pointers or async structs, but I eventually figured I'd make a better use of my energy working on the bridge directly, which I did.

[You can find here my current work on it](https://github.com/fzyzcjy/flutter_rust_bridge/pull/882)

If my suggestion is not acceptable for some reason (transparent `SyncReturn`), feel free to explain me why :)

**Additional context**
I'm so happy that such a project exists and that it is active, thank you ❤️ it will be a pleasure to help as well.

ngasull:
Fixes #881

I've thought a lot through which solution could be best, and figured that matching async API was the best. It makes the code simple and efficient.

I've dug through serialization too but figured it will always need work when adding new features to the bridge. The simplest and most efficient "serialization" already exists: the FFI bindings that represents dart objects. The only issue was that these bindings are blackboxed on dart-sdk's side, which left me no better solution than reimplementing those conversions and even the FFI itself.

However the cost of these changes it pretty low and comparable to implementing custom bytes deserialization.

I'm pretty happy of this solution and I call your attention in case I missed something important, esp. @rogurotus who recently made work that could overlap mine (I already read it all and I think we're fine)

## Checklist

- [x] Validate the solution with the team
- [x] Reimplement `DartCObject` on the dart side to seamlessly match async API
- [x] Simulate `WireSyncReturnStruct` being opaque in generated code, which avoids keeping it synchronized with the actual rust code. Its FFI bindings are already written in `frb_dart` anyways.
- [x] Preserve memory safety (double checks are welcome)

<details>
<summary>Remark for PR creator</summary>

</details>


fzyzcjy:
Looks interesting!

Also cc @Desdaemon who implemented web and various parts

fzyzcjy:
Feel free to ping me when you need a review (either formal or just a brief review for the overall direction)

ngasull:
This PR is ready for review! @fzyzcjy 

Please note that most of the diff is generated code and test entries. Except that, more stuff is actually deleted than added: I removed `ty_sync_return.rs` from rust and dart generator because they are now transparent and identical to default case (could be different for dart but I prefer preserving the code currently relying on IR)

The IR still explicitly preserves the `SyncReturn` information of course.

I also touched `wasm.rs` in dart generator, removing the `void` exception, now typed as dynamic return and allowing it to be treated like any other return type. This is important in sync calls because they are wrapped in `WireSyncReturnStruct`, holding the success of the call. Only `callFfi`'s signature changes; the public API remains `void` of course.

Running valgrind on my machine revealed 8 bytes leaking but I realized it was the same on master, so I must be running it wrong. CI is OK though!

Sync zero copy seems to be working no problem based off the tests.

Please let me know if anything seems odd or needs adjustments :wave: and thank you for reading :slightly_smiling_face:

fzyzcjy:
Great! I will probably review within a few hours :)

fzyzcjy:
(Doing a review now. Below are notes for myself to summarize data flow.)

[TODO]

fzyzcjy (commented):
Great job! I have just started reviewing a part of it, and then realize we may use Dart_Handle instead of Dart_CObject. See my comments below.

fzyzcjy:
💡  Hey I realize a thing: There is `DartHandle` in addition to `DartCObject`. @rogurotus did some great job using DartHandle in flutter_rust_bridge (for opaque types). The DartHandle, IIRC, can directly be translated to objects in Dart automagically by Dart at runtime, and we do not need to write any code for that.

Therefore, can we utilize DartHandle, so we do not need to manually deal with all these DartCObjects? We may refer to @rogurotus's recent code for an idea.

frb_codegen/src/generator/rust/ty.rs
```diff
@@ -115,7 +114,7 @@ impl<'a> TypeRustGenerator<'a> {
             StructRef(ir) => TypeStructRefGenerator { ir, context }.into(),
             Boxed(ir) => TypeBoxedGenerator { ir, context }.into(),
             EnumRef(ir) => TypeEnumRefGenerator { ir, context }.into(),
-            SyncReturn(ir) => TypeSyncReturnGenerator { ir, context }.into(),
+            SyncReturn(ir) => TypeRustGenerator::new(ir.into_inner(), ir_file, config),
```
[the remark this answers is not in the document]
ngasull: So here's the thing: I realized I switched patterns here and I removed `TypeSyncReturnGenerator` on purpose in oder to use `TypeRustGenerator` recursively instead.

Keeping `TypeSyncReturnGenerator` and implementing it only with transparent methods into inner type looks like a lot of "noise", of unnecessary code I think.

Would you suggest a better way to preserve the previous pattern while having a transparent struct?

Same for Dart generator

frb_rust/src/lib.rs
```diff
@@ -21,9 +20,13 @@ mod wasm_bindgen_src;
...
-pub struct SyncReturn<T>(pub T)
-where
-    WireSyncReturnData: From<T>;
+pub struct SyncReturn<T: IntoDart>(pub T);
+
+impl From<()> for SyncReturn<()> {
+    fn from(_: ()) -> Self {
+        SyncReturn(())
+    }
+}
```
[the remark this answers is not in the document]
ngasull: Totally agree! Added it at some point and forgot it there, my bad.

rogurotus:
> bulb Hey I realize a thing: There is `DartHandle` in addition to `DartCObject`. @rogurotus did some great job using DartHandle in flutter_rust_bridge (for opaque types). The DartHandle, IIRC, can directly be translated to objects in Dart automagically by Dart at runtime, and we do not need to write any code for that.
> 
> Therefore, can we utilize DartHandle, so we do not need to manually deal with all these DartCObjects? We may refer to @rogurotus's recent code for an idea.

I haven't read much into the implementation, but I find it more attractive to use 
transparent DartCObject (dart ffi struct) than an opaque Dart_Handle

rogurotus:
I'm talking about the file `frb_dart/lib/src/ffi/io_dartcobject.dart` implementing and getting a transparent structure `extends ffi.Struct`. When I look into the details, I will be able to say more precisely.

`Dart_Handle` is used in ffigen. It replaces it with `dart.ffi.Handle` aka `dart Object`. We use this for opaque Dart types in Rust.

frb_dart/lib/src/ffi/io.dart
```diff
@@ -51,26 +37,134 @@ class DartApiDl {
...
-  /// Not to be used by normal users, but has to be public for generated code
-  @ffi.Int32()
-  external int len;
+  @ffi.Bool()
+  external bool isSuccess;
 
-  /// Not to be used by normal users, but has to be public for generated code
-  @ffi.Uint8()
-  external int success;
+  external ffi.Pointer<ffi.Void> ptr;
```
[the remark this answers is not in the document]
ngasull: Yes, it is used in `WireSyncReturnStruct::free`.

`WireSyncReturnStruct` copies values instead of wrapping a pointer, which is slighly more simple and efficient (right?). Keeping the pointer to the wrapped data allows to `ManuallyDrop` later (`free_WireSyncReturnStruct`) thus letting `allo_isolate`'s `Drop` impl correctly drop the returned struct.

Do you feel we should use `ptr` directly to access value and not copied values?

frb_codegen/src/generator/rust/ty.rs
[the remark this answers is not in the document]
fzyzcjy: Hmm, are we sure SyncReturn will always *purely* delegate, and never develop its own logic? If not, it may be reasonable to do the delegating (though it is verbose).

As for verbose code for delegate: https://crates.io/crates/delegate maybe we can use a crate

ngasull:
All of this handle stuff sound great but I admit I'm stuck. The code generation works well and generate dart object return types! However I can't figure how to actually get a handle pointing to a fresh Dart object created from Rust side. I'm not even sure it's possible and there's very little info on the web about it.

@rogurotus Do you know if it's possible to allocate a handle into dart vm from rust's side? Or if there is a well defined mechanism to transfer ownership to dart?

So far I tried to `Box` a  `DartCObject` and return its pointer as the handle but it returns unclear stuff to dart. I then read from `dart-sys` that handles mustn't be allocated on the head so I suppose I'm the wrong path... but I didn't get any clues to go further 🛟 🙏

rogurotus:
> @rogurotus Do you know if it's possible to allocate a handle into dart vm from rust's side? Or if there is a well defined mechanism to transfer ownership to dart?

well we can call a dart function from rust which will return a Dart object aka Dart_Handle. But for this you need to get a pointer Dart function

But I still think it's a dead end. We need to see how the conversion takes place via DartPort. If I'm not mistaken, your original implementation is essentially this

ngasull:
Okay, this might add a lot more complexity than I expected... Thanks for the insight.

I've been reading [dart-sys docs](https://docs.rs/dart-sys) further and figured the following:

- Dart handles can be created in Rust realm by hand, allocation by allocation ([for example](https://docs.rs/dart-sys/latest/dart_sys/fn.Dart_NewList.html))
- There is no existing conversion function from `DartCObject` to `Dart_Handle`. `DartCObject` is only meant to be sent through an isolate port. Existing implementations thus are useless by themselves.
_NB: `allo_isolate` somehow [reimplement it themselves](https://github.com/sunshine-protocol/allo-isolate/blob/430623bc777cc0fb4d99fb9c065cb5d81762e3ee/src/ffi.rs)._

This leaves us with the following options:

1. Stop relying on `IntoDart` and define our own trait to have a requirement that works both for async (`IntoDart`) and sync (`Dart_Handle`).
_Pretty impactful but potentially the most rigorous option._
2. Keep `IntoDart` requirement and convert generated C struct into manually generated Dart handle from Rust.
_Way less impactful but requires converting from an arbitrary struct, unwrapping unsafe pointer access in Rust_
3. Remain with this PR's current implementation: pass the `DartCObject` to dart as an FFI struct and let `frb_dart` generate the dart object naturally.
_As impactful as 2., read struct from Dart's realm without unwrapping pointers _a la_ Rust, but requires redeclaring `DartCObject` FFI in `frb_dart` because dart sdk keeps it private_

In any case, we will have to implement conversions somewhere.

I would exclude `1.` in order to avoid any major impact on the lib and risk unnecessary bugs (and energy spent). Then, I'd feel safer making use of FFI structs in Dart realm because of error handling instead of potential random segfaults with `2.`. So I'd stay with `3.`, the current solution. I **think** I'm not too biased :sweat_smile:

frb_dart/lib/src/ffi/io.dart
```diff
@@ -51,26 +37,134 @@ class DartApiDl {
...
+        return value.as_int64;
+      case 4: // DartDouble
+        return value.as_double;
+
+      case 5: // DartString
+        int len = 0;
+        while (value.as_string.elementAt(len).value != 0) {
+          len++;
+        }
+        return utf8.decode(value.as_string.cast<ffi.Uint8>().asTypedList(len));
```
fzyzcjy, line 69: nit: Are we sure the string is `\0`-terminated? 
If so, we should add a test about it to ensure. For example, deliberately create a Rust `String` with `\0` *in the middle* and assert results in the dart side.
In addition, maybe add a line of comment pointing to where the "0-terminated" is specified
If not,  we may need to change the way

frb_dart/lib/src/ffi/io.dart
```diff
@@ -51,26 +37,134 @@ class DartApiDl {
...
+        return copy ? Int64List.fromList(view) : view;
+      case 9: // Uint64
+        final view = typedValues.cast<ffi.Uint64>().asTypedList(nValues);
+        return copy ? Uint64List.fromList(view) : view;
+      case 10: // Float32
+        final view = typedValues.cast<ffi.Float>().asTypedList(nValues);
+        return copy ? Float32List.fromList(view) : view;
+      case 11: // Float64
+        final view = typedValues.cast<ffi.Double>().asTypedList(nValues);
+        return copy ? Float64List.fromList(view) : view;
```
fzyzcjy, line 144: nit: we can extract the common pattern to make it a bit cleaner

fzyzcjy (commented):
(uber-nits when looking at code)

fzyzcjy:
I see. The "3" looks safest given it only adds some Dart code.

Is it possible we do more *auto-generation* and less *manual* things? For example, the `case 0, case 1, ...` looks like magic numbers so it may not be a perfect code style (with auto generated io_dartcobject we may have those constants so no magic number). The io_dartcobject.dart would also be great if it is auto generated instead of manually written.
If so, I guess "3" LGTM

frb_example/pure_dart/dart/lib/main.dart
```diff
@@ -160,6 +160,73 @@ void main(List<String> args) async {
...
+    expect(api.handleSyncTypedU64ZeroCopy(input: Uint64List.fromList([10, 20, 30])), Uint64List.fromList([10, 20, 30]));
+  }, skip: skipWeb('Not supported by dart2js'));
+  test('dart call handle_sync_typed_f32_zero_copy', () async {
+    expect(api.handleSyncTypedF32ZeroCopy(input: Float32List.fromList([10.1, 20.2, 30.3])),
+        Float32List.fromList([10.1, 20.2, 30.3]));
+  });
+  test('dart call handle_sync_typed_f64_zero_copy', () async {
+    expect(api.handleSyncTypedF64ZeroCopy(input: Float64List.fromList([10.1, 20.2, 30.3])),
+        Float64List.fromList([10.1, 20.2, 30.3]));
+  });
```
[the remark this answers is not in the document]
fzyzcjy: For example, structs, enums, nested structs, ...

frb_dart/lib/src/ffi/io.dart
ngasull: [We are sure yes](https://github.com/sunshine-protocol/allo-isolate/blob/71b9760993d64ef46794176ca276d1cc637b2599/src/into_dart.rs#L106)! Will add the comment. [std::ffi::CString](https://doc.rust-lang.org/nightly/std/ffi/struct.CString.html) asserts this guarantee.

Because of `allo_isolate`'s implementation, a string containing the _nul_ character will be unwrapped as default (empty). Is this really our job to test this?

frb_dart/lib/src/ffi/io.dart
ngasull: Actually I considered it and figured probably not: type casts are specific and let us access dart extensions to typed lists based on ffi type. Moreover, even though this looks redundant, it's clear and statically well typed.

However if you have a better solution that preserves the same level of type checking I sure am taking the change!

ngasull:
I have to admit the file is originally auto generated and I had to adjust some types because some were broken :grimacing: now that you notice this I realize it's because of the `CPATH` issue I had with the rest of the code generation too.

I'll give another look into the code generation and how to make sure it remains up to date.

frb_dart/lib/src/ffi/io.dart
fzyzcjy: LGTM. Maybe just add a very simple test: `api.rs` has sth like `fn f() -> String { String::new("abc\0def") }` and `main.dart` has `expect(f(), "")`. Just a few lines and we are more sure;)

frb_dart/lib/src/ffi/io.dart
fzyzcjy: e.g.

```
      case 9: // Uint64
        return f<ffi.Uint64, Uint64List>(typedvalues, nvalues, copy, Uint64List.fromList);
      case 10: // Float32
        return f<ffi.float32, Float32list>(typedvalues, nvalues, copy, float32List.fromList);

S f<T, S>(typedValues, S Function() copyer, bool copy) {
          final view = typedValues.cast<T>().asTypedList(nValues);
        return copy ? copyer(view) : view;
}
```

ngasull: Won't be enough: `asTypedList` extension only exists for the specific type of each case. I doubt dart's type system is strong enough to let us abstract the union of these types + the corresponding copyer correctly. Especially that `fromList` is static and that dart won't let us interface static functions.

I'd still buy working implementation though :smile:

fzyzcjy: Anyway, as long as you add one test for each case in `main.dart` and verify the types and content, then we are safe

frb_dart/lib/src/ffi/io.dart
ngasull: Added the test for async as well

frb_example/pure_dart/dart/lib/main.dart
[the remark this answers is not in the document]
ngasull: So I removed the tests I added and copied each existing async test into a sync one, side by side, to indicate maintainers that we should systematically test both.

This involves a lot of copy pasta but I don't feel bad about it in unit testing, especially that this shows both features work the same but independently. Does this look fine to you too @fzyzcjy ?

frb_example/pure_dart/dart/lib/main.dart
```diff
@@ -160,6 +160,73 @@ void main(List<String> args) async {
   test('dart call handle_sync_string', () async {
     expect(api.handleSyncString(input: "Hello Rust!"), "Hello Rust!");
   });
+  test('dart call handle_sync_unit', () async {
+    api.handleSyncUnit();
+  });
+  test('dart call handle_sync_typed_i8', () async {
```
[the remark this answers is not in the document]
ngasull: Fixed by reusing the same pattern as async tests!

frb_dart/lib/src/ffi/io.dart
ngasull: These new tests failed for web as this nul-character behavior is specific to native env. In wasm, the string returned still contains the nul-character.

This should be the ideal scenario for native as well but it's out of the scope of this PR. Just reproducing the existing async behavior.

If it's believed to be important, an issue could be created!

frb_codegen/src/generator/rust/ty.rs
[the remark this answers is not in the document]
ngasull: So I made use of delegate, restoring the existing pattern. Howver couldn't exactly replicate the exact same pattern in the enum pattern: `SyncReturn` is transparent which requires it to instantiate the inner generator at some point.

I considered involving lazy field like a `RefCell` but even this one involves defining it for the `ty.rs`s. Doing it in the `new` makes sense to me, and we could use a new `new` constructor for all other types, but I'm limiting the impact of the changes for now.

What do you think?

ngasull:
All the reviews should be treated!

The big change to be noticed is the creation of a crate. This sounds a bit scary but it's just a cbindgen requirement in order to statically generate `io_dartcobject.dart`. It could actually be used to generate more code at compile time (meaning at "library release time").

I tried to work around cbindgen to generate `DartCObject` and its dependencies from `frb_rust` directly but this involved a lot of other C exports automatically detected. cbindgen taking the whole top level crate for granted, we only have a limited control over the filtering of the exports. Moreover, having a separate crate clearly defines which bindings we generates, and wraps `allo_isolate` only.

I do my best to be the least impactful on the code base, but this for sure will need some validation by the core team :pray:

fzyzcjy:
To add a new crate, we have to do some more things, e.g.:

- [ ] modify `justfile`, ensure the crate is auto updated like other crates (e.g. its version)
- [ ] modify `.github/*`, such as CI, to ensure crate is auto *tested* (e.g. cargo test) and linted (e.g. cargo clippy)

frb_rust_bindgen_io/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "flutter_rust_bridge_bindgen_io"
```
fzyzcjy, line 2: uber-nit: "bindgen io" looks very specific to the underlying tech it uses. For example, maybe one day there is a lib called awesome_new_bindgen, then we have to rename this crate. So shall we name it more general (well naming is hard ;) )

ngasull: Sooo I agree naming is difficult and I could change to any other name, no problem! It's just that we don't need to publish this crate, it's a simple internal tool to regenerate allo_isolate bindings. It's as you wish really!

Will add some details to the reason of this crate

ngasull:
Just to be clear: this new crate is purely internal and only serves to regenerate static dart bindings.

These bindings are committed into `frb_dart`, just like we do for generated examples. I added the generation step to the CI/`.github` so that any diff with the committed state will be spotted.

I didn't touch the `justfile` because regenerating the bindings is no mandatory step for our maintainers, unless we bump `allo_isolate` version.

The new crate is not meant to be published.

Does it make more sense? :slightly_smiling_face:

frb_rust_bindgen_io/Cargo.toml
ngasull: If it durably only treats allo_isolate, it could be called `allo_isolate_bindgen` actually :sweat_smile:

frb_dart/lib/src/ffi/io.dart
```diff
@@ -51,26 +37,137 @@ class DartApiDl {
 /// This class is only for internal usage.
 class WireSyncReturnStruct extends ffi.Struct {
```
fzyzcjy, line 38: Is it possible we also generate this `extends ffi.Struct` info *automatically* via our newly added frb_rust_bindgen_io?

frb_dart/lib/src/ffi/io.dart
```diff
@@ -51,26 +37,137 @@ class DartApiDl {
...
+            externalTypedData, value.as_external_typed_data);
+        return externalTypedData;
+
+      case DartCObjectType.DartSendPort:
+      case DartCObjectType.DartCapability:
+      case DartCObjectType.DartNativePointer:
+      case DartCObjectType.DartUnsupported:
+      case DartCObjectType.DartNumberOfTypes:
+      default:
+        throw "Can't read invalid data type $ty";
```
fzyzcjy, line 103:
```suggestion
        throw Exception("Can't read invalid data type $ty");
```

frb_dart/lib/src/ffi/io.dart
```diff
@@ -51,26 +37,137 @@ class DartApiDl {
...
+  static dynamic _typedDataIntoDart(
+    int ty,
+    ffi.Pointer<ffi.Uint8> typedValues,
+    int nValues, {
+    required bool copy,
+  }) {
+    switch (ty) {
+      case DartTypedDataType.ByteData:
+        final view = typedValues.cast<ffi.Uint8>().asTypedList(nValues);
+        final bytes = copy ? view : Uint8List.fromList(view);
```
fzyzcjy, line 116: ⚠️ looks like it is flipped? if copy should use fromList, but it is the converse

That's why I do not like this style :( too easy to make a mistake

frb_dart/lib/src/ffi/io.dart
```diff
@@ -51,26 +37,137 @@ class DartApiDl {
...
+    }
+  }
+
+  static final _externalTypedDataFinalizer =
+      Finalizer<DartNativeExternalTypedData>((externalTypedData) {
+    final handleFinalizer = externalTypedData.callback
+        .cast<ffi.NativeFunction<NativeExternalTypedDataFinalizer>>()
+        .asFunction<DartExternalTypedDataFinalizer>();
+    handleFinalizer(externalTypedData.data, externalTypedData.peer);
+  });
```
fzyzcjy, line 163: We need to ensure the finalizer part is really executed (in our valgrind environment), because GC may not happen at all (if memory pressure low).

To do this, maybe have a look at https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_example/pure_dart/dart/lib/main_with_vm_service.dart

fzyzcjy (commented):
Great job! Most are nits here, and I cannot wait to merge it :)

frb_rust_bindgen_io/src/main.rs
```diff
@@ -0,0 +1,62 @@
+use std::{fs, path::Path, process::Command};
+
+use dunce::canonicalize;
+
+fn main() -> std::io::Result<()> {
+    let frb_dart_dirname = "../frb_dart";
+    // tempfile won't work because ffigen relies on the file having .h extension
```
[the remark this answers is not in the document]
ngasull: It's named indeed, but I can't give it a cute name! (unless I missed something?)

frb_rust_bindgen_io/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+        name: 'DartCObject'
+        description: 'generated by flutter_rust_bridge_bindgen_io'
+        headers:
+          entry-points:
+            - '{}'
+          include-directives:
+            - '{}'
+        preamble: |
+          // ignore_for_file: camel_case_types, non_constant_identifier_names, avoid_positional_boolean_parameters, annotate_overrides, constant_identifier_names, unused_field
+        ",
```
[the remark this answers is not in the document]
ngasull: `cbindgen.toml` is used to configure `cbindgen`. The conf above configures `ffigen` :)

frb_rust_bindgen_io/Cargo.toml
ngasull: This crate might more or less be dedicated to `frb_dart` bindings generation so...

I felt like it would a better practice to keep crates on the same flat level but here this one is clearly a requirement of `frb_dart`, so I'll look into nesting it in `frb_dart` directly. Simpler, clearer, not impacting workspace.

I see you suggesting having common code generation functions though :) I just feel that they already exist in `frb_codegen` which could be used as a library. This seems a bit out of the scope of this PR isn't it?

frb_rust/src/handler.rs
```diff
@@ -312,30 +304,18 @@ impl ErrorHandler for ReportDartErrorHandler {
...
-                success,
-            }
-        }
+        WireSyncReturnStruct::new(data, success)
     }
 
     #[cfg(wasm)]
     {
-        vec![data.0.into_dart(), success.into_dart()].into_dart()
+        vec![data.into_dart(), success.into_dart()].into_dart()
```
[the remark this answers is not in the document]
ngasull: Probably yes because of zero copy, we need to manually drop allocated memory later in time.

frb_rust/src/support.rs
```diff
@@ -64,69 +65,38 @@ pub fn slice_from_byte_buffer<T: bytemuck::Pod>(buffer: Vec<u8>) -> Box<[T]> {
...
+/// and its name is used in the code generator. Renaming would break the generated bindings.
 #[repr(C)]
 #[cfg(not(wasm))]
 pub struct WireSyncReturnStruct {
-    pub ptr: *mut u8,
-    pub len: i32,
+    pub ty: DartCObjectType,
+    pub value: DartCObjectValue,
     pub success: bool,
+    pub ptr: *mut ManuallyDrop<DartCObject>,
```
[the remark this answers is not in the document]
ngasull: Absolutely. I hesitated quite a bit about this, just dug further and realized there might be no benefit to copied values in this case. Just my bad, will keep it simple.

frb_rust/src/support.rs
```diff
@@ -64,69 +65,38 @@ pub fn slice_from_byte_buffer<T: bytemuck::Pod>(buffer: Vec<u8>) -> Box<[T]> {
...
         }
     }
-}
 
-/// Bool will be converted to u8 where 0 stands for false and 1 stands for true.
-impl From<bool> for WireSyncReturnData {
-    fn from(data: bool) -> Self {
-        u8::from(data).into()
+    pub fn free(self) {
+        let _ = std::mem::ManuallyDrop::into_inner(*unsafe { Box::from_raw(self.ptr) });
```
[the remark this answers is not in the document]
ngasull: So I wanted to return WireSyncReturn by value, which would have dropped it on return. Instead I opted for an internal drop that also allowed keeping both the allocation and the free in the same place in the code.

Anyway, I will change this in order to return a pointer which will directly be dropped just like on master :)

frb_example/with_flutter/macos/Runner/bridge_generated.h
```diff
@@ -38,9 +38,7 @@ typedef struct wire_list_size {
 } wire_list_size;
 
 typedef struct WireSyncReturnStruct {
-  uint8_t *ptr;
-  int32_t len;
-  bool success;
+  uint8_t _opaque[0];
 } WireSyncReturnStruct;
```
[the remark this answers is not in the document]
ngasull: Because the dart bindings were manually coded in `io.dart`, I figured this header definition only serves indicating the existence of the type, whatever is inside! I will remove this anyway with the pointer change :smile: :bomb:

frb_rust_bindgen_io/src/main.rs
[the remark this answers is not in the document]
fzyzcjy: Maybe just create a temp dir (using `tempfile`) and then create whatever filename you like inside it?

frb_rust_bindgen_io/Cargo.toml
fzyzcjy: > they already exist in frb_codegen which could be used as a librar

* frb_codegen: users use it at/before compile time to generate user code
* frb_dart/frb_rust: user use it at runtime
* this_new_crate: *we* use it at/before compile time, to generate *code in frb_dart/frb_rust*, i.e. generate code *inside the flutter_rust_bridge* repo

so they are somehow quite different ;)

fzyzcjy: > This seems a bit out of the scope of this PR isn't it?

No, it is just a naming problem (change a few characters) ;)

IMHO we can put everything related to "generating code in frb_dart/frb_rust" inside this new crate, that's the main point why I want it to have a more general name

frb_rust/src/handler.rs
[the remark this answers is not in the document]
fzyzcjy: Well my question is, can we directly use WireSyncReturnStruct in wasm? It just simply changes how the data is represented, we can still write whatever deallocation code we like

frb_dart/lib/src/ffi/io.dart
ngasull: Well spotted! Okay you made your point :+1: 

I didn't see conditional values as unordered parameters of the same type, will now avoid them when possible (especially repeatedly)

frb_dart/lib/src/ffi/io.dart
ngasull: Fixed, added memory-checking test and comment.

frb_dart/lib/src/ffi/io.dart
ngasull: This one gave me some struggle. I came out with a solution involving some testing code inside the lib itself... Feels dirty but I really didn't find anything better.

Because tests are `dart compile/run`'d, I couldn't check for `Platform` test setting, which means this code won't be stripped down at compile time =/ Even though impact is minimal, any better option is welcome.

frb_rust/src/handler.rs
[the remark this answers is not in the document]
ngasull: Refactored this return type quite a bit, let me know if this looks better once I push all this :)

frb_rust_bindgen_io/src/main.rs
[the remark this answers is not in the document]
ngasull: I eventually opted for the most straightforward usage of cbindgen and ffigen: static `cbindgen.toml` and ffigen config in `pubspec.yaml`. This is the actual recommended use of both libs.

Moreover I committed the generated header file, just like all the other committed generated files.

This options gives very clear understanding of the flow, with minimal code. I feel a lot a trouble around understanding of the flow, so this should help contributors. Sounds good?

frb_dart/lib/src/ffi/io.dart
ngasull: Just pushed latest changes, only this should be missing, I'm trying something.

ngasull:
In order to refactor de external/zero copy typed data switch case in a non-uber-complex way, I had to use _constructor-tearoffs_ Dart feature. Before 2.15, it's an experimental feature, so I tried to enable it in analysis options but it's plain not working.

So for now I bumped the sdk version (:cold_sweat:) while we discuss. It made me notice the docs/troubleshooting were partially outdated because latest `ffigen` actually requires 2.17 now.

What should I do? Replace all tearoffs by anonymous functions? Just stay with 2.15? Btw, if the refactored solution better to you now @fzyzcjy ? :sweat_smile:

fzyzcjy:
Given dart 2.17 has been there [for half a year](https://dart.dev/guides/whats-new), I guess it is ok to be 2.17.

However, I suggest *firstly* create another PR which does nothing but bumping to 2.17, and after merging that PR (should be trivial) we continue on this PR. This is because, when migrating to 2.17, I guess the analyzer will complain many things and we need to fix them. Such fix would be very clear if it can be separated from this PR :)

ngasull:
As we don't need sdk>2.15 yet, I only published bump to 2.15. Half a year is pretty short in the lifetime of a project, and some users might not be able to bump their sdk.

@fzyzcjy  Should I update the PR to bump to 2.17?

frb_rust_bindgen_io/Cargo.toml
ngasull: >    - frb_codegen: users use it at/before compile time to generate user code
>    - frb_dart/frb_rust: user use it at runtime
>    - this_new_crate: we use it at/before compile time, to generate code in frb_dart/frb_rust, i.e. generate code inside the flutter_rust_bridge repo

Sums it up really well! :ok_hand: 

Not sure if you noticed but I moved this crate _inside_ `frb_dart` in order to avoid confusion with the rest of the code base and because it shares configuration with `frb_dart` for simplicity and maintainability. I kept the "bindgen" naming for now, because I think it is plain bingen of an external crate. Shall I rename this folder to "source_generator"? Seems ok to make it live in `frb_dart` (unpublished of course)

fzyzcjy:
Nice suggestion, the 2.15 LGTM (and I have merged it) :)

ngasull:
Hi there ! Anything I can do to get this PR merged @fzyzcjy ? Any end action on the naming of bindgen/sourcegen ?

fzyzcjy:
Ah I just did not realize I need action as there is no notifications ;) Anyway feel free to ping me / request a review whenever you need one!

I will have a review now

justfile
```diff
@@ -100,6 +100,7 @@ refresh_all:
     just gen-bridge 
     (cd frb_rust && cargo clippy -- -D warnings)
     (cd frb_macros && cargo clippy -- -D warnings)
+    (cd frb_dart/bindgen && cargo run)
```
fzyzcjy, line 103:
```suggestion
    (cd frb_dart/bindgen && cargo run)
    (cd frb_dart/bindgen && cargo clippy -- -D warnings)
```

frb_dart/lib/src/ffi/io.dart
```diff
@@ -42,35 +29,177 @@ class DartApiDl {
...
-  external ffi.Pointer<ffi.Uint8> ptr;
+typedef WireSyncReturn = ffi.Pointer<DartCObject>;
+
+dynamic wireSyncReturnIntoDart(WireSyncReturn syncReturn) =>
+    syncReturn.ref.intoDart();
 
-  /// Not to be used by normal users, but has to be public for generated code
-  @ffi.Int32()
-  external int len;
+extension DartCObjectExt on DartCObject {
```
fzyzcjy, line 42: uber-nit: About the code that converts DartCObject into dart data - shall we extract it into a separate file, so it is clearer

frb_dart/lib/src/ffi/io.dart
```diff
@@ -42,35 +29,177 @@ class DartApiDl {
...
   }
 
-  bool get isSuccess => success > 0;
+  static final _externalTypedDataFinalizer =
+      Finalizer<_ExternalTypedDataFinalizerArgs>((externalTypedData) {
+    final handleFinalizer =
+        externalTypedData.callback.asFunction<DartExternalTypedDataFinalizer>();
+    handleFinalizer(externalTypedData.length, externalTypedData.peer);
+
+    if (Platform.environment.containsKey('FRB_TEST')) {
```
fzyzcjy, line 162: uber-nit: if we want best performance, would be great to use https://api.flutter.dev/flutter/dart-core/bool/bool.fromEnvironment.html which is compile time constant (instead of the Platform.environment which is runtime "environment variable"). By doing so, if someone do not define FRB_TEST, then this `if` will be removed at compile time so zero runtime cost.

p.s. defining that constant needs `--dart-define` instead of normal env var

frb_dart/bindgen/io_dartcobject.h
```diff
@@ -0,0 +1,124 @@
+#include <stdbool.h>
```
fzyzcjy, line 1: Wait... I remember someone (you or maybe @rogurotus IIRC) already copies Dart SDK source code into our frb source code (and setup CI to ensure it is not stale). So maybe we should reuse that infrastructure to make it much easier

frb_dart/lib/src/ffi/web.dart
```diff
@@ -114,25 +113,9 @@ abstract class FlutterRustBridgeWireBase {
...
-}
-
-Tuple2<int, int> parseOpaquePtrAndSizeFrom(Uint8List data) {
-  return Tuple2(
-    ByteData.view(data.buffer).getUint32(_webPointerLength),
-    ByteData.view(data.buffer)
-        .getUint32(_webPointerLength + syncReturnPointerLength),
-  );
-}
+dynamic wireSyncReturnIntoDart(WireSyncReturn syncReturn) => syncReturn;
```
fzyzcjy, line 118:
```suggestion
List<dynamic> wireSyncReturnIntoDart(WireSyncReturn syncReturn) => syncReturn;
```

frb_dart/lib/src/ffi/io.dart
```diff
@@ -42,35 +29,177 @@ class DartApiDl {
...
 
-// NOTE for maintainer: Please manually keep in sync with [WireSyncReturnStruct] in Rust
-/// This class is only for internal usage.
-class WireSyncReturnStruct extends ffi.Struct {
-  /// Not to be used by normal users, but has to be public for generated code
-  external ffi.Pointer<ffi.Uint8> ptr;
+typedef WireSyncReturn = ffi.Pointer<DartCObject>;
+
+dynamic wireSyncReturnIntoDart(WireSyncReturn syncReturn) =>
+    syncReturn.ref.intoDart();
```
fzyzcjy, line 40: uber-nit

```suggestion
List<dynamic> wireSyncReturnIntoDart(WireSyncReturn syncReturn) =>
    syncReturn.ref.intoDart() as List<dynamic>;
```

frb_dart/lib/src/basic.dart
```diff
@@ -74,22 +73,26 @@ abstract class FlutterRustBridgeBase<T extends FlutterRustBridgeWireBase> {
...
-    if (raw.isSuccess) {
-      final result = task.parseSuccessData(buffer);
-      inner.free_WireSyncReturnStruct(raw);
-      return result;
-    } else {
-      final errMessage = utf8.decode(buffer!);
-      inner.free_WireSyncReturnStruct(raw);
-      throw FfiException('EXECUTE_SYNC', errMessage, null);
+    try {
+      final syncReturnAsDartObject = wireSyncReturnIntoDart(syncReturn);
```
fzyzcjy, line 83:
```suggestion
      final syncReturnAsDartObject = wireSyncReturnIntoDart(syncReturn);
      assert (syncReturnAsDartObject.length == 2);
```

fzyzcjy (commented):
Just nits and hopefully I can merge it today and release it!

fzyzcjy:
Btw I realized you commented a lot above, but again github does not give me notifications about this :( replying now

frb_example/pure_dart/dart/lib/main.dart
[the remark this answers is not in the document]
fzyzcjy: Sure, unit tests can be "WET" instead of "DRY" when we want to make it more clear

(you can mark this conversation "resolved" after seeing the comment ;) )

frb_dart/lib/src/ffi/io.dart
fzyzcjy: > These new tests failed for web as this nul-character behavior is specific to native env. In wasm, the string returned still contains the nul-character.

Anything reasonable is accpetable, as long as we do not see crashes, memory leaks, invalid memory access, etc.

frb_dart/lib/src/ffi/io.dart
ngasull: Looks like you gave the missing piece of my puzzle, thanks :)

fzyzcjy: You are welcome!

frb_dart/bindgen/io_dartcobject.h
ngasull: Correct me if I'm wrong but I think we're not doing the same thing: this infrastructure copies dart sdk into `frb_rust` to include headers at user codegen time. My problem is to have `DartCObject` bindings at maintainer (us) coding time, in order to generate the dart object with correct static typing.

I'd love to reuse such logic but it probably can't be done as-is. However if you meant to use original dart sdk rather than generating bindings from `allo_isolate` well I sure can look into that.

Adding another `cp` in `code_generator.yml:66` to copy those headers in `frb_dart` as well sound fine to you?

ngasull:
I went through your comments @fzyzcjy, feel free to check if you have a few minutes :pray:

.github/workflows/code_generator.yml
```diff
@@ -63,10 +63,15 @@ jobs:
           git sparse-checkout set runtime/include
           cd ..
           cp -rf ./sdk/runtime/include/* ./frb_rust/src/dart_api_dl/
+          cp -rf ./sdk/runtime/include/* ./frb_dart/dart_api_dl/
```
fzyzcjy, line 66: Well I am hesitate whether copying the whole dir *twice* is a good way or not

What about this: Instead of copying to frb_dart, can we throw away it, and *points to* the frb_rust copy whenever we need those files? Given that we only need it at *our* develop time (instead of user's develop time), this should be ok

frb_dart/pubspec.yaml
```diff
@@ -23,7 +23,20 @@ dependencies:
...
   flutter_rust_bridge_serve: serve
+ffigen:
+  output: 'lib/src/ffi/dart_native_api.dart'
+  name: 'DartCObject'
+  description: 'generated by flutter_dart'
+  headers:
+    entry-points:
+      - 'dart_api_dl/include/dart_native_api.h'
+    include-directives:
+      - 'dart_api_dl/include/dart_native_api.h'
```
fzyzcjy, line 39: i.e. just these two lines

fzyzcjy (commented):
Ready to merge it, just one nit (separated as two comments) :)

fzyzcjy:
Great! Will review within an hour

fzyzcjy (approved):

fzyzcjy:
Excellent job!

dbsxdbsx:
> All suggested changes should be pushed! Feel free to tell me if there's any more issue @fzyzcjy :)
With 1.57.0，I found [my template](https://github.com/dbsxdbsx/frb_opencv) can't build app successfully on win10. One of issues is:
```
Undefined class 'WireSyncReturnStruct'.
Try changing the name to the name of an existing class, or creating a class with the name 'WireSyncReturnStruct'.
```

fzyzcjy:
@dbsxdbsx Hi, could you please create a new issue with more details (for searchability and debuggability)

dbsxdbsx:
> @dbsxdbsx Hi, could you please create a new issue with more details (for searchability and debuggability)

I would do so if I can't solve it myself, otherwise, I would pr directly.  
By the way,  since the [testing APIs](https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_example/pure_dart_multi/rust/src/api_1.rs) for multi-blocks are not complete enough, may I update the APIs there first---of course, the CI test would fail, but at least when a new feature is merged, test for multi-blocks would be covered more thoroughly.

fzyzcjy:
Happy to hear that, and looking forward to your updates!

> By the way, since the [testing APIs](https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_example/pure_dart_multi/rust/src/api_1.rs?rgh-link-date=2022-12-24T06%3A07%3A00Z) for multi-blocks are not complete enough, may I update the APIs there first---of course, the CI test would fail, but at least when a new feature is merged, test for multi-blocks would be covered more thoroughly.

anything that *improves* code quality / feature are encouraged, definitely

ngasull:
Code correctness and more tests are always welcome :muscle: 

I re-checked the whole code base for `WireSyncReturnStruct` and I can confirm there is no more reference it except in historical doc.

@dbsxdbsx  Are you sure you've been running all the necessary code generation? Is your `flutter_rust_bridge_codegen` cargo crate installed with the latest version? If the issue persists, please open up a dedicated issue :pray: thanks!

dbsxdbsx:
> Code correctness and more tests are always welcome 💪
> 
> I re-checked the whole code base for `WireSyncReturnStruct` and I can confirm there is no more reference to it except in historical doc.
> 
> @dbsxdbsx Are you sure you've been re-running code generation? Is your `flutter_rust_bridge_codegen` cargo crate installed with the latest version? If the issue persists, please open up a dedicated issue 🙏 thanks!

@ngasull, yes, I am using the latest frb(currently v1.59.0).  It is still not compilable, due to issue:
issue1

issue2

And I've uploaded the [issued example project](https://github.com/dbsxdbsx/frb_opencv), I test it on win10.

Before posting a new issue, could you please test it, to see if it could be compilable with your machine?(Just download, run `just clean && just`) Hopefully, I didn't set something wrong.

ngasull:
@dbsxdbsx Pulled and checked your code and I can affirm that there's no more reference to `WireSyncReturnStruct` after running:

```bash
# In `native` folder
flutter_rust_bridge_codegen \
    --wasm \
    --rust-input src/api_1.rs \
    --c-output ../macos/Runner/bridge_generated.h \
    --dart-output ../lib/gened_api_1.dart
    
flutter_rust_bridge_codegen \
    --wasm \
    --rust-input src/api_2.rs \
    --dart-output ../lib/gened_api_2.dart
```

:bulb: I actually could debug your issue by trying out your `just` script: the `--wasm` option isn't passed to `flutter_rust_bridge_codegen`. Indeed try to remove the comments or move `--wasm` flag at the beginning of the command. `just` just silently ignores the issue :upside_down_face:

PR #890, opened 2022-12-14
Delete dart_sys

rogurotus:
fix build for windows #877

## Checklist

- [x] CI is passing.

### DIFF
--- a/frb_rust/src/ffi/dart_api.rs
+++ b/frb_rust/src/ffi/dart_api.rs
@@ -4,7 +4,16 @@
 
 use std::ffi::c_void;
 
-use dart_sys::{Dart_Handle, Dart_PersistentHandle};
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct _Dart_Handle {
+    _unused: [u8; 0],
+}
+
+#[allow(non_camel_case_types)]
+pub type Dart_Handle = *mut _Dart_Handle;
+#[allow(non_camel_case_types)]
+pub type Dart_PersistentHandle = Dart_Handle;
 
 #[link(name = "trampoline")]
 extern "C" {
--- a/frb_rust/src/ffi/dart_api.rs
+++ b/frb_rust/src/ffi/dart_api.rs
@@ -5,7 +5,6 @@
 use std::ffi::c_void;
 
 #[repr(C)]
-#[derive(Debug, Copy, Clone)]
 pub struct _Dart_Handle {
     _unused: [u8; 0],
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -344,12 +344,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7a81dae078cea95a014a339291cec439d2f232ebe854a9d672b796c6afafa9b7"
 
 [[package]]
-name = "dart-sys"
-version = "2.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42fe64308f2072bb94ecd8f7557f9ac01b69d86e6199bac24be0f091ad74b092"
-
-[[package]]
 name = "either"
 version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


frb_rust/src/ffi/dart_api.rs
```diff
@@ -4,7 +4,16 @@
 
 use std::ffi::c_void;
 
-use dart_sys::{Dart_Handle, Dart_PersistentHandle};
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
```
fzyzcjy, line 8: uber-nit: should it be copy/clone? (I know we will only use the `*mut` pointer, but not sure about this derive)

frb_rust/src/ffi/dart_api.rs
```diff
@@ -4,7 +4,16 @@
...
+pub struct _Dart_Handle {
+    _unused: [u8; 0],
+}
+
+#[allow(non_camel_case_types)]
+pub type Dart_Handle = *mut _Dart_Handle;
+#[allow(non_camel_case_types)]
+pub type Dart_PersistentHandle = Dart_Handle;
```
fzyzcjy, line 16: so, is it possible we somehow generate the rust code from c code? (also ok if not)

fzyzcjy (commented):
LGTM % nits

gutenfries:
> @\gjf2a I was very surprised that you had not seen the issue on Windows in version +1.53 as it seemed to be a fairly certain bug. I have created a PR for my suggestion of a fix in the dart-sys library but it's a third party library that hasn't been touched in three years so we will have to see how that goes...
> 
> If it's not successful then the owners of FRB can make a call on whether they fork that library or include it's code directly. It is clear that the problem exists for anyone developing or compiling on windows and it would be great to get it fixed properly

Is it worth considering forking [dart-sys](https://github.com/OptimisticPeach/dart-sys) and/or integrating it into the repository?

The whole library is very lightweight (59KB including tests, docs, and examples) and unlicensed.

PR Review from https://github.com/fzyzcjy/flutter_rust_bridge/pull/890/#discussion_r1048408705:
> so, is it possible we somehow generate the rust code from c code? (also ok if not)

I believe it would be best, _at least for now, if we could use the underlying API from dart-sys, as it bypasses cross-compiling to C entirely.

We may consider C cross-compilation in the future for performance reasons, however, that will require a _Lot_ of development time and resources to get off the floor (in my opinion).


dart-sys is very well documented and though we are currently having issues with build.rs, as mentioned in #877, and the library is rather unorganized, it still is a lot of legwork that, _in my opinion_, is not worth discarding.

I am willing to contribute to this project.

@rogurotus 
@fzyzcjy 
Let me know your thoughts. Perhaps a new Issue & PR?

fzyzcjy:
@gutenfries Looks interesting and feel free to create issue/PR! Totally agree that our goal is: cleaner and more maintainable code (e.g. by using auto-generation instead of manually writing, good testing, etc) + easy-to-use (e.g. do not have complicated compilation)

rogurotus:
The main thing is to be safe. dart_sys contains functions not from dart_api_dl.h and we need guarantees to port them (like https://github.com/dart-lang/sdk/issues/47356#issuecomment-1302253513).

frb_rust/src/ffi/dart_api.rs
rogurotus: I don't think it will be possible to correctly generate _Dart_Handle

frb_rust/src/ffi/dart_api.rs
rogurotus: Yes, it is not necessary

rogurotus:
I think we need to merge this fix and then integrate dart-sys because it might not be fast.

gutenfries:
> I think we need to merge this fix and then integrate dart-sys because it might not be fast.

I agree. The underlying issues in dart-sys are larger than just what is breaking prod in #877. I think it would be best to delete dart-sys for now, as being progressed in this PR, and in the next semver we can include dart-sys again with the appropriate changes & adoptation to the code.

fzyzcjy:
Ok, so from your discussions, I should merge this PR first.

PR #891, opened 2022-12-14
Migrate to modern Android environment variables

svenstaro:
As can be seen at https://developer.android.com/studio/command-line/variables, the only modern variables are `ANDROID_HOME` and `ANDROID_USER_HOME`, the former of which I think is applicable in this case.

I removed the issue checklist as I don't think it's required for a doc change.


Desdaemon:
LGTM

Issue #1, opened 2021-10-04
Missing license.


cldershem:
**Describe the bug**
This code is not have a license and is therefore closed source.

**Expected behavior**
A LICENSE file in the repository.

**Additional context**
Posting the repository publicly [does not](https://choosealicense.com/no-permission/) make it open source.

Anything you write is automatically copyrighted unless you explicitly license it. The vast majority of Rust projects use the MIT license, but you can do whatever you feel is best. I think the website I linked above can walk you through choosing a license and adding a LICENSE file to your codebase.

derekdreery:
Thanks again for making this code public, it's awesome and will help so many people.

But yes it won't be possible for people to use it in production/at work unless it has a license that allows them to. The parts of the library that are used in users final binaries would need to be permissive (e.g. MIT), but the codegen parts could be more copyleft if that is your preference, since they do not appear in the binary downstream.

fzyzcjy:
Nice catch! I will add one definitely.

fzyzcjy:
Done. Thanks for the suggestions!

PR #894, opened 2022-12-14
(feat/bugfix): Integrate & update dart-sys crate

### LINKED ISSUE #893: (feat/bugfix): Integrate dart-sys crate
# (feat/bugfix): Integrate dart-sys crate

## Main Goal: 

Integrate the [dart-sys](https://github.com/OptimisticPeach/dart-sys) crate directly into the repository, as well as organizing and updating much of the code.

## Related:

[Bug] Compiles with 1.51, does not compile with 1.53. #877
Delete dart_sys #890

## Tasks:

- [x] Integrate the crate directly into the repository & use as a cargo path-managed dependency
	- completed in [`599e941` (#894)](https://github.com/fzyzcjy/flutter_rust_bridge/pull/894/commits/599e941bc304ebf05acc21467ec5f1c286455d0c)
- [ ] Organize and restructure the crate, following rust best practices, in order to produce _clean_, _maintainable_, and _comprehensive_ code.
- [x] fix the issues with `build.rs` #877.
- [ ] rewrite unsafe code https://github.com/fzyzcjy/flutter_rust_bridge/pull/890#issuecomment-1351692892
- [ ] write better tests for the crate
	- [ ] Integration
	- [ ] Unit

## See Also:

[discussion on #890](https://github.com/fzyzcjy/flutter_rust_bridge/pull/890/#issuecomment-1351533716)

gutenfries:
# (feat/bugfix): Integrate & update dart-sys crate

*Fixes #893*

## Main Goal:
Integrate the [dart-sys](https://github.com/OptimisticPeach/dart-sys) crate directly into the repository, as well as organizing and updating much of the code.

## Related:
[Bug] Compiles with 1.51, does not compile with 1.53. #877 Delete dart_sys #890

## Tasks:
* [x]  Integrate the crate directly into the repository & use as a cargo path-managed dependency
* [ ]  Organize and restructure the crate, following rust best practices, in order to produce _clean_, _maintainable_, and _comprehensive_ code.
* [x]  fix the issues with `build.rs` [[Bug] Compiles with 1.51, does not compile with 1.53. #877](https://github.com/fzyzcjy/flutter_rust_bridge/issues/877).
* [ ]  rewrite unsafe code [Delete dart_sys #890 (comment)](https://github.com/fzyzcjy/flutter_rust_bridge/pull/890#issuecomment-1351692892)
* [ ]  write better tests for the crate
  * [ ]  Integration
  * [ ]  Unit
* [ ] write better documentation
 
## See Also:
[discussion on #890](https://github.com/fzyzcjy/flutter_rust_bridge/pull/890/#issuecomment-1351533716)


gutenfries:
@fzyzcjy What do you think of this?

dart_sys/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "dart_sys"
+version = "2.0.1"
+# authors = ["Patrik Buhring <patrikbuhring@gmail.com>"]
+edition = "2018"
+build = "build.rs"
+description = "Bindings to the dart native extensions api"
+# repository = "https://github.com/OptimisticPeach/dart-sys"
```
fzyzcjy, line 8: Well I realize the original repo does *not* have a license! From what I have learnt (https://github.com/fzyzcjy/flutter_rust_bridge/issues/1), "no license" means "nobody is allowed to use it". So unless they put sth like a MIT license or explicitly allow us to fork it, we should not use it :/

dart_sys/src/tests.rs
```diff
@@ -0,0 +1,1698 @@
+#[allow(unused_imports)]
```
fzyzcjy, line 1: we need to setup `ci.yaml` to test these as well (should be quite simple, just a few more lines)

dart_sys/src/lib.rs
```diff
@@ -0,0 +1,4395 @@
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+#![allow(unused_variables)]
+#![allow(dead_code)]
+
+mod tests;
+
+/* automatically generated by rust-bindgen */
```
fzyzcjy, line 9: how is this file generated? IMHO we should:

1. include the file generation script as a part of source
2. in `ci.yaml`, run file generation script automagically, to ensure lib.rs does not change. To do this, just mimic current CI, where we run flutter_rust_bridge codegen automatically to ensure generated codes in frb_examples are up to date

dart_sys/Cargo.toml
gutenfries: Check the Cargo.toml at the source repo
https://github.com/OptimisticPeach/dart-sys/blob/master/Cargo.toml#L12

It says that the _crate_ is MIT, but I think it would probably still be best to ask the author...

dart_sys/src/lib.rs
gutenfries: I truly have no idea how it was generated... Take a look at the crate's `Cargo.toml`.
https://github.com/OptimisticPeach/dart-sys/blob/master/Cargo.toml#L22

`bindgen` was included as a dev-dependency and at some point commented out. 

I think that it was probably originally generated but is now hand maintained...?

There is a directory at the root named `bindgen/` with a single file, `wrapper.h`.

Wrapper.h:
```c
#include "dart_api.h"
#include "dart_native_api.h"
#include "dart_tools_api.h"
```

There is nothing in the `build.rs` related to code generation, so my best guess is that these are all remnants from code that at some point was generated, but is now just maintained by hand.

Thoughts?

dart_sys/src/tests.rs
gutenfries: proposed `ci.yml` changes:
https://github.com/fzyzcjy/flutter_rust_bridge/pull/894/commits/5907df4f5cc48fde84d998f18a970997ffa18297#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133dd

fzyzcjy: LGTM, roughly it is like that

dart_sys/src/lib.rs
fzyzcjy: > my best guess is that these are all remnants from code that at some point was generated, but is now just maintained by hand.

I also think so. However, to be a part of flutter_rust_bridge, we should ensure the quality of code. Thus, maybe try to firstly write down a simple script to invoke bindgen, see how the output differs, try to modify bindgen config, see again output, etc. if we have to modify sth by hand, would you please describe it here so we may find out a solution to avoid modifying by hand.

gutenfries: please have a thorough look at the diff in this commit:
https://github.com/OptimisticPeach/dart-sys/commit/ecdb7ffc4fa3515dee163172e0e703c28914d964

this should answer questions on where the apparently generated parts of files came from, as well as why they are no longer needed to be generated.

dart_sys/Cargo.toml
gutenfries: In progress

dart_sys/src/tests.rs
gutenfries: For the purposes of getting started on this project, I am going to close this review and then as we work on the `dart_sys/` crate we can update the tests are write better ones as well as test anything we add :)

dart_sys/src/lib.rs
gutenfries: this means we can safely remove `bindgen/wrapper.h`

fzyzcjy: Good spot :)

> why they are no longer needed to be generated.

Well IMHO we should still keep the generation script (which you have already found), and ensure the generated content is up-to-date via CI. This is analogy to: Say someone have a .rs file and compile it to .so, then we just throw away the .rs (given that it rarely changes) and only keep .so in the codebase - surely we should not do that ;)

fzyzcjy: If it is hard to trigger the generation script, maybe wait for https://github.com/fzyzcjy/flutter_rust_bridge/pull/882 - which adds another crate to solely run bindgen

gutenfries: Agreed 🤝

PR #922, opened 2022-12-24
Add API for testing type String for multi-blocks

dbsxdbsx:
Adding an API with type `String` for testing multi-blocks.  Since there are more and more features, to avoid potential issues when using multi-blocks, this API is essential, since `String` is quite a common type and not a built-in type for Dart and Rust, it can reflect issues quite common.

ps: the CI test may fail.


fzyzcjy:
LGTM after CI passes :)

fzyzcjy:
👀

dbsxdbsx:
> 👀
This pr is closed on purpose. I would try to fix issues related to multi-block in another pr.

fzyzcjy:
I see. Take your time!

PR #924, opened 2022-12-25
fix: force library bundling for iOS and macOS in release mode

GregoryConrad:
Hi, I just noticed that in release mode the static library was getting stripped for iOS and macOS.

I can't quite explain why, but this exact change with separating a method from a variable seems to do the trick. Note: setting the variable directly to the result of `dummy_...` does *not* work.


fzyzcjy:
I think so, the previous method may still be stripped if the compiler is smart (or silly) enough ;)

Issue #909, opened 2022-12-18
Multiple files binding on macOS


rhobro:
### Describe the bug

I've been following the guide to binding multiple files from Rust to dart. As part of this, my binding gen taskrunner looks like this:
```
gen:
    flutter_rust_bridge_codegen \
        --rust-input "native/src/api_discover.rs" "native/src/api_extratest.rs" \
        --dart-output "lib/native/bridge_generated_api_discover.dart" "lib/native/bridge_generated_api_extratest.dart" \
        --rust-output "native/src/generated_api_discover.rs" "native/src/generated_api_extratest.rs" \
        --c-output "macos/Runner/generated_api_discover.h" "macos/Runner/generated_api_extratest.h" \
        --class-name ApiDiscover ApiExtratest
```

However when I run it, it gives the following error:
```
error: Found argument 'macos/Runner/generated_api_extratest.h' which wasn't expected, or isn't valid in this context
```

When I keep only the first one and set the output to `macos/Runner/bridge_generated.h`, it only seems to create the bindings for the first Rust file.

### Codegen logs with `RUST_LOG=debug` environment variable

```shell
in previous ^
```

### OS

macOS

### Version of `flutter_rust_bridge_codegen`

1.55

### Version of `clang++`

14.0.0

### Version of `ffigen`

7.2.4

fzyzcjy:
Is this related: https://github.com/fzyzcjy/flutter_rust_bridge/pull/862

rhobro:
Hi there, I think so. But just adding one header file seems to only include methods from the first input file but not the second.

fzyzcjy:
Well I mean whether that PR solves this issue?

rhobro:
Yes I think this is tackling my problem exactly. When can it be merged by?

fzyzcjy:
current status: for @dbsxdbsx's review :)

rhobro:
Right now, methods in the first rust file seem to be ok but when calling ones from the second one, stuff like this happens:
```
Failed to lookup symbol 'wire_discover': dlsym(RTLD_DEFAULT, wire_discover): symbol not found
```

rhobro:
Just to confirm, for each file are we required to take the dylib and create a separate impl on it?

fzyzcjy:
Would be great to ask in the PR since this part is implemented by them :)

fzyzcjy:
[...]

PR #862, opened 2022-11-27
support header files with multi files

huang12zheng:
Close #909

## Checklist
fix #781
- [x] An issue to be fixed by this PR is listed above.
- [x] New tests are added to ensure new features are working. End-to-end tests are usually in the `./frb_example/pure_dart` example, more specifically, `rust/src/api.rs` and `dart/lib/main.dart`.
- [x] The code generator is run and the code is formatted (e.g. via `just refresh_all`).
- [ ] If this PR adds/changes features, documentations (in the `./book` folder) are updated.
- [ ] CI is passing.

Be dependent on #862

just check config.rs, lib.rs and some(two) test output : header files

Issue #781, opened 2022-10-14
[Bug] Headers files for macos/ios are not generated as expected when following the multiple files article


AhyaanTech:
### Describe the bug

http://cjycode.com/flutter_rust_bridge/article/generate_multiple_files.html 

Following that article i created two rust files and tried to do code gen. but the generated headers file only have functions that was present in the last file to be provided to the argument

### Codegen logs with `RUST_LOG=debug` environment variable

```shell
Tools • Dart 2.18.2 • DevTools 2.15.0
[2022-10-14T03:00:54Z INFO  lib_flutter_rust_bridge_codegen] Picked config: Opts { rust_input_path: "/Users/jawadkalia/Downloads/flutter_rust_bridge_template-main/native/src/release_mode.rs", dart_output_path: "/Users/jawadkalia/Downloads/flutter_rust_bridge_template-main/lib/bridge_generated_releasemode.dart", dart_decl_output_path: None, c_output_path: ["/Users/jawadkalia/Downloads/flutter_rust_bridge_template-main/ios/Runner/bridge_generated.h", "/Users/jawadkalia/Downloads/flutter_rust_bridge_template-main/macos/Runner/bridge_generated.h"], rust_crate_dir: "/Users/jawadkalia/Downloads/flutter_rust_bridge_template-main/native", rust_output_path: "/Users/jawadkalia/Downloads/flutter_rust_bridge_template-main/native/src/generated_releasemode.rs", class_name: "ReleaseModeClass", dart_format_line_length: 80, skip_add_mod_to_lib: false, llvm_path: ["/opt/homebrew/opt/llvm", "/usr/local/opt/llvm", "/usr/lib/llvm-9", "/usr/lib/llvm-10", "/usr/lib/llvm-11", "/usr/lib/llvm-12", "/usr/lib/llvm-13", "/usr/lib/llvm-14", "/usr/lib/", "/usr/lib64/", "C:/Program Files/llvm", "C:/msys64/mingw64"], llvm_compiler_opts: "", manifest_path: "/Users/jawadkalia/Downloads/flutter_rust_bridge_template-main/native/Cargo.toml", dart_root: Some("/Users/jawadkalia/Downloads/flutter_rust_bridge_template-main"), build_runner: true, block_index: BlockIndex(1), skip_deps_check: false, wasm_enabled: false, inline_rust: false }
[...]

PR #930, opened 2022-12-26
Fix unexpected content in c output header

dbsxdbsx:
## What's new 
Currently, when explicitly setting flag `c-output`,  **only the content of the last one blocks of rust API would be written in**, no matter how many fields(paths) are within `c-output`. 

The reason is that frb would combine all fields(paths) in `c-output` for each block. Therefore, when frb finishes generating, only the content for the last block is contained(https://github.com/fzyzcjy/flutter_rust_bridge/pull/862#issuecomment-1362495197).

The way to solve this is to make the number of `c-output` the same as the number of `rust-input`.  If not consistent, the generation would fail. But specifically for the flag `c-output`, it could be omitted --- in this case, frb would still generate c output files as before, but now the number of files would still be consistent.

This issue is not obvious, since the generated c files are temporary, and each block of APIs is generated sequentially. Thus it doesn't harm generating dart files afterward. **But once users want to hold them, the issue appears**(related issues are #781 and #909).

The only drawback of this modification is that it is no longer possible to generate multiple c files (in different paths) which contain the same content (for a certain block of API). But I think the tradeoff is worth it. 

NOTE: If this pr is approved, then #862 should be rejected.

## Checklist
- [x] New tests are added to ensure new features are working. End-to-end tests are usually 
- [ ] CI is passing.
- [ ] document is updated.


fzyzcjy:
Good job! I will review soon (probably today) when having

fzyzcjy:
> The only drawback of this modification is that it is no longer possible to generate multiple c files (in different paths) which contain the same content (for a certain block of API). But I think the tradeoff is worth it.

Well I am not sure whether we can do this. For example, when Flutter app targets ios and macos (and other platforms), IIRC, we need to put generated_headers.h into *both* macos and ios folders. Thus this feature is needed. Please correct me if I am wrong - I may remembered it wrongly

justfile
```diff
@@ -29,10 +29,11 @@ gen-bridge: build
                 -d {{frb_flutter}}/lib/bridge_generated.dart \
                 --dart-decl-output {{frb_flutter}}/lib/bridge_definitions.dart \
                 -c {{frb_flutter}}/ios/Runner/bridge_generated.h \
-                -c {{frb_flutter}}/macos/Runner/bridge_generated.h \
+                # -c {{frb_flutter}}/macos/Runner/bridge_generated.h \
```
fzyzcjy, line 32: Ah this is what I mean. We should really generate to both ios and macos!

fzyzcjy:
However, it is possible we somehow come up with a more complicated configuration option. For example (may be good or bad), for 2 api.rs and 2 platforms (ios and macos), users may need to specify 2x2=4 locations

justfile
dbsxdbsx: If it is necessary, why CI tests for IOs and MacOs are passed?

dbsxdbsx: > Ah this is what I mean. We should really generate to both ios and macos! So this line cannot be commented - otherwise, the macos header file will be stale soon, and cause problems

If it is necessary, why CI tests for IOs and MacOs are passed?

fzyzcjy: because there are *stale* bridge_generated.h on that location :/

dbsxdbsx: First, I still have no idea what "stale" means here. What kind of problem would occur, would this result in a failed compilation?

Second, I read [documentation ](https://cjycode.com/flutter_rust_bridge/integrate/ios_headers.html)here, it seems that the c header file has to be put in path `ios/Runner/` for building IOS APP or `macos/Runner/` for MacOS APP, and users have to specify the `c-output` manually, right (please correct me if I am wrong)?  

If this is the case, my suggestion is to add a new flag `apple` while following the logic of this pr. Then, once the flag `apple` is provided:

1. if no `c-output` flag is provided, frb could generate c header(s) in `ios/Runner/` and `macos/Runner/` respectively with the same name in each APPLE path, which corresponds to the same rust block--- Say, there are 2 rust blocks, then `c_output_01.h` and `c_output_02.h` would be provided in both `ios/Runner/` and `macos/Runner/`(altogether 4 files).
2. if the `c-output` flag is provided, then like for other target platforms(Android, Windows),  generation would still fail if the provided number of fields in  `c-output`  is not consistent with that of `rust-input`. Otherwise, after generating the c file of path/name specified in `c-ouput`, frb could then generate (ALL) c file(S) in the above stated 2 APPLE paths with the specified name(s).

The logic here is:
no matter what platform, each field of  `c-output` should always have the same meaning. 
For special platforms (IOS/MacOS), just tell frb with a new flag, then frb would handle all other dirty work.

dbsxdbsx: > 

just 1 thing to clarify, to build IOS or macos APP, the coding platform has to be MacOS, right?  If this is the case, then the new flag is no needed, rust could detect running OS by :
```rust
use std::env;

fn main() {
    let os_string = env::consts().os;
    println!("Current OS String: {}", os_string);
}
```

fzyzcjy: > First, I still have no idea what "stale" means here. What kind of problem would occur, would this result in a failed compilation?

Suppose user changes api.rs or something like that, and trigger frb codegen. Then, if you comment out the macos one, we will see ios/bridge.h updated, **but macos/bridge.h is still old and not updated at all**. Then, when you build macos app, you will be using the new api.rs **and the old bridge.h** so it can cause trouble

fzyzcjy: > If this is the case, my suggestion is to add a new flag apple while following the logic of this pr.

Well IMHO it may not be perfect. Because frb should not be bound to Flutter and platforms too tightly. It should work well with even pure dart, and whatever folder structure that people like. It may even generate bridge.h into android/windows/whatever platform as long as the user needs it.

huang12zheng: > If it is necessary, why CI tests for IOS and MacOS are passed?

Since we only pass for non-multiple APIs before, when there was only one API, we needed to send the header to ios and mac separately
Of course, you can give it to Linux if you want.

Just my understanding.
z_z

dbsxdbsx: > It may even generate bridge.h into android/windows/whatever platform as long as the user needs it.

With the design of this pr,  it is feasible for users to generate the c header file into any custom path by flag `c-output`. The issue is just about MacOS and IOS, for which the c head file is essential for building app with flutter/dart, if I am not mistaken.

> Because frb should not be bound to Flutter and platforms too tightly. It should work well with even pure dart

Here some things I need to clarify:
1. For platforms like Windows, web, and Android, the c head file during generating is not needed after generating the corresponding dart file? 
2. for building a flutter app targeting IOS/MacOS, the c head file is essential, right? If so, is the path `ios/Runner/` or `macos/Runner/` the compulsory path?
3. for building an app with pure dart? Does the c head file necessary? If so, where the file should be put?

fzyzcjy: > For platforms like Windows, web, and Android , is the c head file during generating still needed after generating the corresponding dart file? (In my test, I think the answer is no)
> for building an app with pure dart? Does the c head file necessary? If so, where the file should be put?

Well my point is that, even if *currently our scenario* does only require macos/ios and not other platforms to have .h, we may not rely on it. If we rely on it, we are essentially writing the frb codebase in a way such that *frb mandatory knows what is Flutter platforms*, while I have claimed in README that this package is pure-dart compatible.

In other words, "high cohesion, low coupling" - IMHO we should not couple our frb with flutter platform knowledge

dbsxdbsx: > > For platforms like Windows, web, and Android , is the c head file during generating still needed after generating the corresponding dart file? (In my test, I think the answer is no)
> > for building an app with pure dart? Does the c head file necessary? If so, where the file should be put?
> 
> Well my point is that, even if _currently our scenario_ does only require macos/ios and not other platforms to have .h, we may not rely on it. If we rely on it, we are essentially writing the frb codebase in a way such that _frb mandatory knows what is Flutter platforms_, while I have claimed in README that this package is pure-dart compatible.
> 
> In other words, "high cohesion, low coupling" - IMHO we should not couple our frb with flutter platform knowledge

So what is your suggestion to solve these issues?

fzyzcjy: For example (may be good or bad), for 2 api.rs and 2 platforms (ios and macos), let users specify 2x2=4 locations

dbsxdbsx: > For example (may be good or bad), for 2 api.rs and 2 platforms (ios and macos), let users specify 2x2=4 locations

what about this:
1. take flag `c-ouput` follows the logic of this pr--each filed represent one API block;
2. take a new flag, say  `extra-c-output-path`, for bak all c files into specific path;

Example 1:
there are 2 API blocks, and users want to put c files in `pathA`, `pathB` and `pathC`,
then command could be:
```
...
-c-output "pathA/api1.h" "pathA/api2.h"
-extra-c-output-path  "pathB"  "pathC"
...
```
Here in `pathA`, `pathB`, and `pathC`, `api1.h` and `api2.h` are contained relatively.

Example 2:
there are 2 API blocks, the name of c file is no cared, but users want to put them in  `pathA`,
then command could be:
```
...
# no -c-output
-extra-c-output-path  "pathA"
...
```
In this case, all generated files would bak into `pathA`, though the names are  arbitrary.

fzyzcjy: LGTM

dbsxdbsx: > LGTM :)

I've commit accordingly in ba004911c64c850118c4ca5fcc914761cae51a08, but it is weird that the CI test is not triggerred.

fzyzcjy:
weird ci does not trigger... could you please commit arbitrarily to trigger it again

dbsxdbsx:
> weird ci does not trigger... could you please commit arbitrarily to trigger it again

I still can't trigger it. I am gonna recreate an pr._

PR #934, opened 2022-12-27
Fix `SyncReturn<RustOpaque<T>>`

rogurotus:
New SyncReturn system ignores `SyncReturn<RustOpaque<T>>`
learn more - `frb_sync_generator_test function`
This PR fixes it

## Checklist

- [x] CI is passing.


rogurotus:
@fzyzcjy  ping just in case

fzyzcjy:
Nice fix :) will probably review today

frb_codegen/src/ir/ty.rs
```diff
@@ -94,6 +94,14 @@ impl IrType {
         matches!(self, RustOpaque(_))
     }
 
+    #[inline]
+    pub fn is_sync_rust_opaque(&self) -> bool {
+        match self {
+            SyncReturn(sync) => sync.clone().into_inner().is_rust_opaque(),
```
fzyzcjy, line 100: would be great not to clone, but anyway this is just uber nit

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -1277,6 +1277,13 @@ pub fn frb_generator_test() -> RustOpaque<FrbOpaqueReturn> {
     panic!("dummy code");
 }
 
+/// Structure for testing the SyncReturn<RustOpaque> code generator.
+/// FrbOpaqueSyncReturn must be only return type.
+/// FrbOpaqueSyncReturn must should be without wrapper like Option<> Vec<> etc.
```
fzyzcjy, line 1282:
```suggestion
/// FrbOpaqueSyncReturn must be without wrapper like Option<> Vec<> etc.
```

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -1277,6 +1277,13 @@ pub fn frb_generator_test() -> RustOpaque<FrbOpaqueReturn> {
...
+pub fn frb_sync_generator_test() -> SyncReturn<RustOpaque<FrbOpaqueSyncReturn>> {
```
fzyzcjy, line 1283: well, we should add a few lines in the rust side to check it does return what we want

fzyzcjy (commented):
LGTM % nits :)

frb_example/pure_dart/rust/src/api.rs
rogurotus: I did not quite understand. Can you explain in more detail? This is a test for the code generator to generate the necessary functions and finalizer

fzyzcjy: Yes I agree. But without anything in the dart side https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_example/pure_dart/dart/lib/main.dart, we do not know whether, for example, those generated code panics

frb_codegen/src/ir/ty.rs
rogurotus: This is due to the specifics of the new implementation of SyncReturn. 
Since it delegates functions to the internal type. RustOpaque not in `distinct output_types`  
and have to check `SyncReturn<RustOpaque<T>>`

frb_example/pure_dart/rust/src/api.rs
rogurotus: In theory, we know when we run dart tests. If no methods are generated dart will just crash.
This is exactly how I found this error.

fzyzcjy: Yes, the question is, does it work *in runtime*. For example, will it happen that, when calling this function, it never returns the desired rustopaque, but somehow panic? (I know it won't - but would be great to test it)

just as simple as:

```
test('...'), () {
  await api.frb_sync_generator_test();
});

Comment (rogurotus) [reply frb_example/pure_dart/rust/src/api.rs:1283]: no problem
```
expect(instanceof(api.frbSyncGeneratorTest(), FrbOpaqueSyncReturn), isTrue);
```

fzyzcjy (approved):
Good fix, waiting for CI to land :)

rogurotus:
@fzyzcjy  CI ready

fzyzcjy:
Good job!

PR #937, opened 2022-12-27
Fix unexpected content in c output header (refine #930)

dbsxdbsx:
This is a refined pr based on #930, where CI test was not triggered.

## What's new

1. The logic of the flag `c-output` is modified. Now each field within this flag should only respond to one `rust-input` API block at that place.
2. new flag `extra-c-output-path` is introduced. All fields within this flag are paths to bak ALL c output files, whatever flag `c-output` is provided or not.flag `c-output`.

## Checklist
- [x] New tests are added for local and online CI tests.
- [x] CI is passing.
- [x] document is updated.


frb_example/pure_dart_multi/rust/build.rs
```diff
@@ -15,12 +15,24 @@ const RUST_OUTPUT_2: &str = "src/generated_api_2.rs";
...
+    #[cfg(feature = "c-output")]
+    return true;
+    false
+}
+
+fn use_extra_c_output_path_flag() -> bool {
+    #[cfg(feature = "extra-c-output-path")]
+    return true;
+    false
+}
```
fzyzcjy, line 28: https://doc.rust-lang.org/rust-by-example/attribute/cfg.html

```
cfg!(target_os = "linux") 
```

is a bool

frb_codegen/src/config.rs
```diff
@@ -187,10 +191,57 @@ pub fn parse(raw: RawOpts) -> Vec<Opts> {
...
+                    .map(|s| s.to_owned())
+                    .collect::<Vec<_>>();
+                first.extend(iter);
+            }
+
+            first
+        })
+        .collect::<Vec<_>>();
+    //↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑ c output path(s) ↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑
+
```
fzyzcjy, line 244: would be great to extract to a separate function if possible

fzyzcjy (commented):
Good job! Only two nits

frb_example/pure_dart_multi/rust/build.rs
dbsxdbsx: I don't get it.

frb_codegen/src/config.rs
dbsxdbsx: It is moved into a new function called `get_refined_c_output`, cd09bcb737e03e8875b477db2a0a474837bc7c7e.

frb_example/pure_dart_multi/rust/build.rs
dbsxdbsx: Fixed in [cd09bcb](https://github.com/fzyzcjy/flutter_rust_bridge/pull/937/commits/cd09bcb737e03e8875b477db2a0a474837bc7c7e).

dbsxdbsx:
The commit  [ff3061f](https://github.com/fzyzcjy/flutter_rust_bridge/pull/937/commits/ff3061fcfbbc77d0df99b36a3b67bfd33c7cfa89) is used to make `just gen-help` behavior consistent across all platforms. Otherwise, the output of section `USAGE:` in `book/src/help.txt` would be : `flutter_rust_bridge_codegen.exe [OPTIONS]` instead of `flutter_rust_bridge_codegen [OPTIONS]` on windows.

frb_codegen/src/config.rs
```diff
@@ -242,6 +240,72 @@ pub fn parse(raw: RawOpts) -> Vec<Opts> {
         .collect()
 }
 
+fn get_refined_c_output(
```
fzyzcjy, line 243: last uber-nit: would be great to add unit tests for this function, since it is so isolated and has a bit of complexity.

for example:

1. single api file, no extra
2. single api file, has extra
3. multi api files, no extra
4. multi api files, has extra

fzyzcjy (commented):
last uber-nit and ready to merge :)

frb_codegen/src/config.rs
dbsxdbsx: Tests for 1,3 and 4 have already existed, so I only add tests for 2 in 47bbcec14a175d1490d0f0fb7d1fc8195f3c8f8c.
And for all examples, I didn't add case with flag `extra-c-output-path` but without the flag `c-output`, in that:
1. this case is quite rare in use;
2. with this case added, there would be new files produced with arbitrary names after running the test both locally by `just precommit` or online by CI test. It is difficult to check correctness.

fzyzcjy: Well I mean `unit test` - https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html. The unit tests are a good way to ensure this function never breaks, and also documents the function (by providing examples)

The 47bbcec end-to-end test is not required (but if you like surely you can keep it)

dbsxdbsx: > Well I mean `unit test` - https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html. The unit tests are a good way to ensure this function never breaks, and also documents the function (by providing examples)
> 
> The [47bbcec](https://github.com/fzyzcjy/flutter_rust_bridge/commit/47bbcec14a175d1490d0f0fb7d1fc8195f3c8f8c) end-to-end test is not required, but if you like surely you can keep it, as long as it does not cause problems

Actually, before adding test for CI and justfile, I've come up the unit test idea for `pure_dart` and `pure_dart_multi`.  But question is I didn't find a way to test following the procedure in `build.rs` with no syntax error, say I create path `tests\test.rs` at the same level of `src`/`build.rs` in project `pure_dart`, the content like this:
```rust
#[path = "../build.rs"]
// mod module_name;
use lib_flutter_rust_bridge_codegen::{
    config_parse, frb_codegen, get_symbols_if_no_duplicates, RawOpts,
};

#[cfg(test)]
mod test {
    use super::*;

    #[test]
    fn test_1() {
        println!("cargo:rerun-if-changed={}", RUST_INPUT);
        // Options for frb_codegen
        let mut raw_opts = RawOpts {
            // Path of input Rust code
            rust_input: vec![RUST_INPUT.to_string()],
            // Path of output generated Dart code
            dart_output: vec![DART_OUTPUT.to_string()],
            wasm: true,
            dart_decl_output: Some("../dart/lib/bridge_definitions.dart".into()),
            dart_format_line_length: 120,
            // for other options use defaults
            ..Default::default()
        };

        // ...

        let all_symbols = get_symbols_if_no_duplicates(&configs).unwrap();
        for config in configs.iter() {
            frb_codegen(config, &all_symbols).unwrap();
        }
    }

    #[test]
    fn test_2() {}
}
```
I don't know how to make it compliable.

fzyzcjy: Well for unit test I mean unit test the single function - `get_refined_c_output`. Just test this one function, and pretend the rest of all code *does not exist at all*. (That's why it is called "unit" testing - only test the unit, i.e. the function)

dbsxdbsx: Ok, I found the way with no syntax error, just change the head into this:
```rust
#[path = "../build.rs"]
mod lib_flutter_rust_bridge_codegen;
use lib_flutter_rust_bridge_codegen::{
    config_parse, frb_codegen, get_symbols_if_no_duplicates, RawOpts,
};
```
Then it is compilable. 
In addition, if this is not the unit test you want, please tell me.

dbsxdbsx: > Well for unit test I mean unit test the single function - `get_refined_c_output`. Just test this one function, and pretend the rest of all code _does not exist at all_. (That's why it is called "unit" testing - only test the unit, i.e. the function)

I see. But I wonder how to catch panic for:
```rust
fn bail(err: clap::ErrorKind, message: Cow<str>) {
    RawOpts::command().error(err, message).exit()
}
```
The `bail` would be called in a `should_panic` test.

fzyzcjy: LGTM, if you want to test it just catch the panic

dbsxdbsx: I don't know how, since

> LGTM, if you want to test it just catch the panic

I mean I don't know how to catch the panic from ` RawOpts::command().error(err, message).exit()`. From commit aa11a608bf2f0b801b2d306ec6a636202f9fef66, if uncomment test case `test_coutput_with_inconsistent_number_of_input_block_api`, then the test routine can't proceed.

frb_codegen/src/config.rs
```diff
@@ -482,3 +548,113 @@ impl Opts {
         .with_extension("freezed.dart")
     }
 }
+
+#[cfg(test)]
+mod test {
```
fzyzcjy, line 553: Well, sorry but this may not be unit test... Unit test looks like:

```
    fn test_coutput_with_single_input_block_api() {
        let output = get_refined_c_output(vec!["api_1.rs"], vec!["./pathA/c_output.h"]);
        assert_eq!(output, vec!["the_value_that_you_expect_here"]);
  }
```

and write a few such tests, asserting scenarios like no extra output, have extra output, ...

dbsxdbsx: I've enumerated all cases in 6be714a6711f3dc08a3b6fcf1586bee077b3bfc3.

fzyzcjy: Great! I will review soon

frb_example/pure_dart/rust/build.rs
```diff
@@ -22,6 +22,18 @@ fn main() {
...
+        raw_opts.c_output = Some(vec!["./c_output_path/c_output.h".into()]);
+    }
+
+    if cfg!(feature = "extra-c-output-path") {
+        raw_opts.extra_c_output_path = Some(vec![
+            // For test, the below 2 paths format are made a little different
+            "./extra_c_output_path_1/".into(),
+            "extra_c_output_path_2".into(),
+        ]);
+    }
```
fzyzcjy, line 36: Ready to merge! Given that we have thorough unit tests, maybe we can remove these in pure_dart example (indeed I have not required this one - I originally meant unit tests), and only remain such end-to-end tests in pure_dart_multi. Because pure_dart is already quite huge, and adding such tests will make generated content tripled.

fzyzcjy (approved):
Ready to merge after uber-nit deletion!

frb_example/pure_dart/rust/build.rs
dbsxdbsx: Done since d84c77b18632b1eabc4957cb2a3d203531c89baa.

fzyzcjy (approved):

fzyzcjy:
Great job!

PR #949, opened 2023-01-04
Refactors for `Boxed` and `Option`

Desdaemon:
- Refactors `EnumRef` to make `Option<FieldlessEnum>` work
- (Potentially breaking) `Box<primitive>` and `Option<Box<primitive>>` on WASM no longer allocates a Box, but receives a (nullable) value directly from Dart. Makes the implementation cleaner.

Closes #828.

## Checklist

- [x] An issue to be fixed by this PR is listed above.
- [x] New tests are added to ensure new features are working. End-to-end tests are usually in the `./frb_example/pure_dart` example, more specifically, `rust/src/api.rs` and `dart/lib/main.dart`.
- [x] The code generator is run and the code is formatted (e.g. via `just refresh_all`).
- [x] If this PR adds/changes features, documentations (in the `./book` folder) are updated.
- [x] CI is passing.

PR #950, opened 2023-01-04
feat: properly handle `List<DateTime>`, `List<Duration>`

Desdaemon:
Also fixes a bug where WASM wire functions incorrectly accepted the `Int64List` wrapper instead of the inner `BigInt64Array` typed array.

Depends on #949, so should not be reviewed/merged before it.

## Checklist

- [ ] CI is passing.


Desdaemon:
To be split after #949 is unblocked.

PR #963, opened 2023-01-09
feat: allow opt-out of WASM initializer

Desdaemon:
## Checklist

- [x] CI is passing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ lazy_static = "1.4.0"
 uuid = "1.1.2"
 
 flutter_rust_bridge_macros = { path = "frb_macros", version = "1" }
-flutter_rust_bridge = { path = "frb_rust", version = "1" }
+flutter_rust_bridge = { path = "frb_rust", version = "1", default-features = false }
 flutter_rust_bridge_codegen = { path = "frb_codegen", version = "1" }
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ lazy_static = "1.4.0"
 uuid = "1.1.2"
 
 flutter_rust_bridge_macros = { path = "frb_macros", version = "1" }
-flutter_rust_bridge = { path = "frb_rust", version = "1", default-features = false }
+flutter_rust_bridge = { path = "frb_rust", default-features = false, version = "1" }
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ lazy_static = "1.4.0"
 uuid = "1.1.2"
 
 flutter_rust_bridge_macros = { path = "frb_macros", version = "1" }
-flutter_rust_bridge = { path = "frb_rust", version = "1" }
+flutter_rust_bridge = { path = "frb_rust", version = "1", default-features = false }
```
fzyzcjy, line 28: Hmm, so we are never testing the wasm-start feature in frb? Because IIRC we only test web for pure-dart, and now that one has wasm-start feature disabled.

fzyzcjy (commented):
Good job, except for one small suggestion:

Cargo.toml
Desdaemon: It's that way because flutter_rust_bridge is a workspace dependency, so specifying default-features on pure-dart has no effect. I could try to change pure-dart to depend by path, but not sure if that has any unintended effects.

fzyzcjy: Totally agree. However, future PRs may make wasm-start more fancy etc, so it would be great to make it testable.

For example, is it possible to create one extra example as a sibling of pure_dart example, such that one example uses wasm-start (probably the default?) and another example (say example_without_wasm_start) forbids wasm start. Then, our pure_dart example captures the default case, while the example_without_wasm_start example tests when wasm-start disabled

Desdaemon: Another example crate might be overkill; we could use custom profiles in Cargo.toml to do this.

fzyzcjy:
Good job! I will review it probably today and merge it :)

fzyzcjy:
LGTM, except for one nit - please correct me if I am wrong since quite busy recently! Seems that I have not seen any change in GitHub CI code. So I guess we do not test both behaviors (i.e. opt-out and non-opt-out)?

Desdaemon:
> LGTM, except for one nit - please correct me if I am wrong since quite busy recently! Seems that I have not seen any change in GitHub CI code. So I guess we do not test both behaviors (i.e. opt-out and non-opt-out)?

That's fine, I'll put in a test for CI as well.

fzyzcjy:
Good job!

Issue #966, opened 2023-01-12
Possible Solution for Async Rust Functions


anlumo:
**Is your feature request related to a problem? Please describe.**

According to [the documentation Section 3.9](https://cjycode.com/flutter_rust_bridge/feature/async_rust.html), having async functions is currently not possible due to issues with the Rust runtime.

Async functions are very important especially on the Web target, since all I/O is inherently async on the Web. So if you're trying to do anything other than data structures or pure compute, it won't work at all. [Section 18.1](https://cjycode.com/flutter_rust_bridge/article/async_in_rust.html) does provide some possible solutions, but they all don't work on the Web target, because there's no way to block.

**Describe alternatives you've considered**

There is a solution that's surprisingly simple. Note that I haven't actually tested it, this is purely based on the documentation.

There's Stream support [according to Section 3.3](https://cjycode.com/flutter_rust_bridge/feature/stream.html). So, an async function can simply take a `StreamSink<T>` as the first parameter, then spawn a new task in whatever way the runtime wants (like [`wasm_bindgen_futures::spawn_local`](https://docs.rs/wasm-bindgen-futures/latest/wasm_bindgen_futures/fn.spawn_local.html)), return immediately, and once the result is available, call `.add(T)` on the StreamSink and then close the stream.

On the Dart side, the caller can call `.single` on the returned Stream and use that as the Future.

**Describe the solution you'd like**

Ideally, this should just work as a solution for async as well:

Couldn't the Stream feature be adjusted so that there's a similar solution for just returning a Future? So take a `FutureSink<T>` as the first parameter and then whenever you feel like, call `.resolve(T)` (which automatically closes the internal Stream). On the Dart side, this could be an actual Future that was created from the Stream (preceded with the Future created by the sync function call).

fzyzcjy:
Well, the solution is indeed simpler:

The stream, and normal rust function, indeed supports it naturally. This is because, their implementation looks like:

```
// normal function
fn generated_function() {
  let ans = your_rust_sync_function_in_api_rs();
  dart_utility_send_data(ans);
}

// stream
impl StreamSink {
  pub fn add(data) { dart_utility_send_data(data); }
}
```

So, to support Rust async, we already have all things needed:

```
async fn generated_function() {
  let ans = your_rust_sync_function_in_api_rs().await; // JUST ADD AWAIT
  dart_utility_send_data(ans);
}
```

fzyzcjy:
[...]

PR #989, opened 2023-01-17
RustOpaque::Clone has Incorrect Bounds

### LINKED ISSUE #988: Let RustOpaque Expose the Arc

Working with async code (via `Stream`s) with this project means that I have to keep objects around even after the function call itself returns. Due to the ownership rules of Rust, I have to use something like `Arc` for this, because both my closure and Dart have to keep a reference to that object around and lifetimes are not transferred over the bridge. `RustOpaque` is already halfway there by using an `Option<Arc<_>>` internally. However, the Rust code doesn't have access to that field.

`RustOpaque` should implement a function that returns a clone of the inner `Arc`.

It's possible to do it right now by using an `Arc` as part of the type wrapped by RustOpaque, but then it's an unnecessary double-`Arc`.

**Additional context**

The PR is incoming, this ticket is just there because you require one for the PR.

anlumo:
Fixes #988

## Checklist

- [X] An issue to be fixed by this PR is listed above.
- [X] New tests are added to ensure new features are working. End-to-end tests are usually in the `./frb_example/pure_dart` example, more specifically, `rust/src/api.rs` and `dart/lib/main.dart`.
- [X] The code generator is run and the code is formatted (e.g. via `just refresh_all`).
- [x] CI is passing.

Not sure if documentation is needed for this, since it should be self-explanatory.

### DIFF
--- a/frb_example/pure_dart/rust/src/api.rs
+++ b/frb_example/pure_dart/rust/src/api.rs
@@ -13,7 +13,9 @@ use anyhow::{anyhow, Result};
 use flutter_rust_bridge::*;
 use lazy_static::lazy_static;
 
-use crate::data::{EnumAlias, Id, MyEnum, MyStruct, NonCloneData, StructAlias, UserIdAlias};
+use crate::data::{
+    EnumAlias, Id, MyEnum, MyStruct, NonCloneData, NonCloneData, StructAlias, UserIdAlias,
+};
 pub use crate::data::{FrbOpaqueReturn, FrbOpaqueSyncReturn, HideData, NonSendHideData};
 use crate::new_module_system::{use_new_module_system, NewSimpleStruct};
 use crate::old_module_system::{use_old_module_system, OldSimpleStruct};
--- a/frb_example/pure_dart/rust/src/api.rs
+++ b/frb_example/pure_dart/rust/src/api.rs
@@ -13,7 +13,7 @@ use anyhow::{anyhow, Result};
 use flutter_rust_bridge::*;
 use lazy_static::lazy_static;
 
-use crate::data::{EnumAlias, Id, MyEnum, MyStruct, StructAlias, UserIdAlias};
+use crate::data::{EnumAlias, Id, MyEnum, MyStruct, NonCloneData, StructAlias, UserIdAlias};
 pub use crate::data::{FrbOpaqueReturn, FrbOpaqueSyncReturn, HideData, NonSendHideData};
 use crate::new_module_system::{use_new_module_system, NewSimpleStruct};
 use crate::old_module_system::{use_old_module_system, OldSimpleStruct};
@@ -1251,13 +1251,6 @@ pub fn run_opaque(opaque: RustOpaque<HideData>) -> String {
     opaque.hide_data()
 }
 
-#[allow(clippy::redundant_clone)]
-pub fn run_opaque_inner(opaque: RustOpaque<HideData>) -> String {
-    // Tests whether `.clone()` works even without the generic type wrapped by it
-    // implementing Clone.
-    opaque.clone().hide_data()
-}
-
 pub fn run_opaque_with_delay(opaque: RustOpaque<HideData>) -> String {
     sleep(Duration::from_millis(1000));
     opaque.hide_data()
@@ -1268,6 +1261,17 @@ pub fn opaque_array() -> [RustOpaque<HideData>; 2] {
         RustOpaque::new(HideData::new()),
         RustOpaque::new(HideData::new()),
     ]
+}
+
+pub fn sync_create_non_clone() -> SyncReturn<RustOpaque<NonCloneData>> {
+    SyncReturn(RustOpaque::new(NonCloneData::new()))
+}
+
+#[allow(clippy::redundant_clone)]
+pub fn run_non_clone(clone: RustOpaque<NonCloneData>) -> String {
+    // Tests whether `.clone()` works even without the generic type wrapped by it
+    // implementing Clone.
+    clone.clone().hide_data()
 }
 
 pub fn create_sync_opaque() -> RustOpaque<NonSendHideData> {
--- a/frb_example/pure_dart/rust/src/api.rs
+++ b/frb_example/pure_dart/rust/src/api.rs
@@ -1252,7 +1252,7 @@ pub fn run_opaque(opaque: RustOpaque<HideData>) -> String {
 }
 
 pub fn run_opaque_inner(opaque: RustOpaque<HideData>) -> String {
-    opaque.as_inner().hide_data()
+    opaque.clone().hide_data()
 }
 
 pub fn run_opaque_with_delay(opaque: RustOpaque<HideData>) -> String {
--- a/frb_example/pure_dart/rust/src/api.rs
+++ b/frb_example/pure_dart/rust/src/api.rs
@@ -1251,6 +1251,10 @@ pub fn run_opaque(opaque: RustOpaque<HideData>) -> String {
     opaque.hide_data()
 }
 
+pub fn run_opaque_inner(opaque: RustOpaque<HideData>) -> String {
+    opaque.as_inner().hide_data()
+}
+
 pub fn run_opaque_with_delay(opaque: RustOpaque<HideData>) -> String {
     sleep(Duration::from_millis(1000));
     opaque.hide_data()
--- a/frb_example/pure_dart/rust/src/api.rs
+++ b/frb_example/pure_dart/rust/src/api.rs
@@ -1251,7 +1251,7 @@ pub fn run_opaque(opaque: RustOpaque<HideData>) -> String {
     opaque.hide_data()
 }
 
-#[allow(clippy::redundant-clone)]
+#[allow(clippy::redundant_clone)]
 pub fn run_opaque_inner(opaque: RustOpaque<HideData>) -> String {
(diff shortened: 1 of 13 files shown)


fzyzcjy:
Given it is draft, I will review after it is marked as ready for review :)

fzyzcjy:
cc @rogurotus - do you think this operation (as_inner) is safe, or will we somehow violate some implicit assumptions?

rogurotus:
> Due to the ownership rules of Rust, I have to use something like Arc for this, because both my closure and Dart have to keep a reference to that object around and lifetimes are not transferred over the bridge.

I didn't quite understand what was going on.
RustOpaque can live entirely on the dart side. (If you pass an object to dart, delete it on the rust side, it will continue to be valid and can be passed back to rust (until dispose is called))

In general, I don’t see any problems, I just don’t quite understand why this is necessary.

anlumo:
> In general, I don’t see any problems, I just don’t quite understand why this is necessary.

Maybe I should supply an example.

Let's say I'm implementing a database in Rust to be accessed from Dart (which is what I'm actually writing right now). All access to the database is asynchronous, because it's using I/O.

`api.rs`:

```rust
pub struct Entry {
    value: String,
}

pub struct Database {
   db: RustOpaque<DatabaseEngine>,
}

impl Database {
    pub fn fetch(&self, sink: StreamSink<Entry>, key: String) {
        let db = self.db.as_inner();
        tokio::spawn(async move {
            let entry = db.get(&key).await.unwrap();
            sink.add(entry);
            sink.close();
        });
    }
}
```

I need to keep `db` around while the closure is running, but I can't do that without having some kind of `Arc` or similar wrapper.

(Note that this obviously isn't my code directly, since there's also error handling and wasm handling and other stuff in there. It's just the structure.)

fzyzcjy:
@anlumo btw will the problem disappear when rust async (https://github.com/fzyzcjy/flutter_rust_bridge/issues/966) is implemented? The rust async looks general and can be widely used

anlumo:
> @anlumo btw will the problem disappear when rust async (#966) is implemented? The rust async looks general and can be widely used

Depends. The Rust borrow checker would be able to handle this case in an async function, but it depends on the lifetime requirements that are required by the executor. For example, if the Future returned from the function would have to be `'static`, it wouldn't work. If the lifetime can be tied to `&self`, it would.

rogurotus:
RustOpaque implements Clone , isn't that what you need?

anlumo:
> RustOpaque implements Clone , isn't that what you need?

I might be able to just use RustOpaque directly, yeah… I have to test that.

anlumo:
> RustOpaque implements Clone , isn't that what you need?

Tried it. It only implements `Clone` when `T` implements `Clone`, so that doesn't work.

The problem is that the code is autoderiving `Clone`, and that's a know issue with the macro that can't be fixed as far as I know.

However, I could fix that within this PR.

rogurotus:
> Tried it. It only implements Clone when T implements Clone, so that doesn't work.

Yes, it looks like a bug. Behavior should be like Arc::clone

anlumo:
> Yes, it looks like a bug. Behavior should be like Arc::clone

I renamed this PR to this instead and also pushed the change to my branch.

anlumo:
I have no idea why my generated files are different. It contains a few new functions where I have no idea what they do.

fzyzcjy:
> I have no idea why my generated files are different. It contains a few new functions where I have no idea what they do.

which ones? the only failure seems only to be a clippy error

image

anlumo:
Never mind, forgot that I added a new test function `run_opaque_inner` which of course generates some more functions.

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -1156,6 +1156,7 @@ pub fn run_opaque(opaque: RustOpaque<HideData>) -> String {
     opaque.hide_data()
 }
 
+#[allow(clippy::redundant-clone)]
```
fzyzcjy, line 1159: Well I would suggest just fix that linter error

anlumo: The clone is the whole point of that test though.

fzyzcjy: Oh I see. Then would be great to add a line of comment to explain it

dbsxdbsx:
> I have no idea why my generated files are different. It contains a few new functions where I have no idea what they do.

Have you tried to run `just precommit` locally before pr?

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -1156,6 +1156,13 @@ pub fn run_opaque(opaque: RustOpaque<HideData>) -> String {
     opaque.hide_data()
 }
 
+#[allow(clippy::redundant_clone)]
+pub fn run_opaque_inner(opaque: RustOpaque<HideData>) -> String {
+    // Tests whether `.clone()` works even without the generic type wrapped by it
+    // implementing Clone.
+    opaque.clone().hide_data()
```
fzyzcjy, line 1163: uber-nit: To make it explicit and avoid future programmers accidentically adding Clone to HideData, what about this: We can create a *new* struct with *explicit* name saying it is not clone. e.g. `struct NonCloneableData { a: String }`, and use it here

fzyzcjy (commented):
Good job! I am ready to merge it, if @rogurotus also thinks this PR is OK

frb_example/pure_dart/rust/src/api.rs
anlumo: Well, I made sure to check that HideData isn't Clone.

fzyzcjy: How did you do that? If it is done automatically then also OK.

This is because, we have to imagine that a future programmer may do some refactoring or adding features/tests that may break current code if there is no *automated* tests

anlumo: I just did it manually. You're right, if it ever got added, the whole test wouldn't work any more (well, it wouldn't fail, but it'd work with the old code as well).

fzyzcjy: Yes that's the problem - then the test will not test what you want to test anymore

rogurotus:
> Good job! I am ready to merge it, if @rogurotus also thinks this PR is OK

Yes, that's how it's supposed to work.

fzyzcjy:
Great!

Then I will merge it after CI is green

fzyzcjy (approved):

fzyzcjy:
👀

anlumo:
Small update: I got reassigned to a different project for the time being, so I can't continue my work for this.

fzyzcjy:
Take your time! (Btw this one is already OK so after CI is green it can be merged)

satvikpendem:
Any way I can help contribute to this? I was trying to do something similar and would be willing to try fixing the CI. Anything I should do @anlumo and @fzyzcjy?

anlumo:
I'd welcome any help to getting this thing working, but I don't know the best course of action. Should I just give you push access to my fork that this PR is based on? Do you want to have your own fork based on mine and then make a new PR?

satvikpendem:
Push access should be fine, we don't need a new PR since this has relevant details already

fzyzcjy:
Feel free to ping me if need any operations (e.g. make a review) :)

satvikpendem:
Main CI issue looks like the bridge generated files want `use crate::data::NonCloneData;` imported at the top, but they're auto-generated, so I'm not sure where to add this use statement if I'd want the generated files to pick them up. I tried in `api.rs` but it just says that the use statement is unused while still throwing the CI error.

fzyzcjy:
> the bridge generated files want use crate::data::NonCloneData; imported at the top

Not checked the details, but is it possible that, we avoid the need of `use ...` by generating code of *full* names. For example, instead of generating code with name NonCloneData, we generate `flutter_rust_bridge::data::NonCloneData` everywhere

satvikpendem:
Yeah that might solve a lot of issues, with fully qualified names.

anlumo:
@fzyzcjy please merge this PR!

fzyzcjy:
Sure!

satvikpendem:
Awesome, glad to see this happen!

fzyzcjy:
@all-contributors please add @satvikpendem for code

satvikpendem:
Hey @fzyzcjy no need, I didn't actually contribute any code or commits to this PR, it was all @anlumo

fzyzcjy:
@satvikpendem Oh I just checked the discussion history and thought your co-authored. But looks like you submitted at least 1 lines of code? Then you are indeed contributor ;) Since people who only submit 1-char doc fix is contributor as well. At the same time, I document "who did what" in the README, so it is clear what is done by each contributor.

image

satvikpendem:
Haha, alright sounds good then, didn't realize I pushed a commit haha!

satvikpendem:
Ah wait that code seems wrong:

```rust
use crate::data::{
    EnumAlias, Id, MyEnum, MyStruct, NonCloneData, NonCloneData, StructAlias, UserIdAlias,
};
```

The NonCloneData use definition occurs twice, should be a clippy lint warning and removed.

anlumo:
> The NonCloneData use definition occurs twice, should be a clippy lint warning and removed.

I did fix that during my rework yesterday.

PR #998, opened 2023-01-26
Migration to dart-sys

rogurotus:
This PR removes manual bindings to `dart_api_dl` in favor of `dart-sys` (Thanks @gutenfries).
 #893

## Checklist

- [ ] CI is passing.


fzyzcjy:
Looks interesting, looking forward to ready-for-review :)

fzyzcjy:
Hmm the ios test is failing, maybe we can fix it before reviewing?

rogurotus:
> Hmm the ios test is failing, maybe we can fix it before reviewing?

this emulator won't start

fzyzcjy:
Looks like github ci has been updated, thus maybe change 16.0 to 16.2

image

rogurotus:
@fzyzcjy, CI passed

fzyzcjy:
Good job! I will probably review it today (since quite busy with my research)

.github/workflows/code_generator.yml
```diff
@@ -58,11 +58,11 @@ jobs:
 
       - name: Check DartApi
         run: |
-          git clone --depth 1 --filter=blob:none --sparse https://github.com/dart-lang/sdk.git
+          git clone --depth 1 --filter=blob:none --sparse --branch stable https://github.com/dart-lang/sdk.git
           cd sdk
           git sparse-checkout set runtime/include
           cd ..
-          cp -rf ./sdk/runtime/include/* ./frb_rust/src/dart_api_dl/
+          cp -rf ./sdk/runtime/include/* ./frb_rust/src/dart_api/
```
fzyzcjy, line 65: uber-nit: curious why rename dart_api_dl to dart_api? And the title suggests that we use dart-sys, so shall we remove the whole dart_api?

fzyzcjy, line 65: Btw, would be great if dart-sys could use the same style of checking (e.g. at https://github.com/dart-sys/dart-sys/blob/main/.github/workflows/ci.yml). Then we are sure that dart-sys is also never stale

fzyzcjy (commented):
Good job! Only two nits and ready to merge

.github/workflows/code_generator.yml
rogurotus: renamed because only `dart_api.h` is used.
It is used to generate 
https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_dart/lib/src/ffi/dart_native_api.dart

Unfortunately, there is still a dependence on the `.h`. Getting rid of this addiction is not included in the PR

rogurotus: There is something similar
https://github.com/dart-sys/dart-sys/blob/main/update-dart-sdk/src/main.rs

fzyzcjy: Great! Is it run in CI for every commit?

rogurotus: as far as I remember, @gutenfries has an idea to do this, but I'm not sure if it has been implemented now

fzyzcjy: I see. Made an issue there -> https://github.com/dart-sys/dart-sys/issues/55

gutenfries: hi! 

im planning on this and will work on it when I get back from vacation next week 👍🏼

fzyzcjy: @gutenfries Hi, take your time!

gutenfries: Hi folks, @fzyzcjy @rogurotus

I am back and beginning to work on the assorted problems surrounding Dart-sys.

When I first took over as maintainer, I simply updated the old code, which used dynamic generation and had a lot of unprotected edge cases.

I believe that it is time to reconsider the cleanest and most maintainable way to have bindings to the dart-sdk in rust and believe that _static_ generation should be preferred to dynamic generation.

pros of static generation:
- better IDE compatibility
- better platform integration (windows is a pain because of dynamic linking)
- better maintainability & independence (dart-sys will not have external dependencies. currently, you need OpenSSL, a working C and CC compiler, as well as a _full_ installation of the dart sdk on windows.) 
- reduced ship weight. (dart-sys currently needs to ship with a heavy build script, as well as a lot of boilerplate code that is likely to go unused. With static generation, the code footprint will be larger, but the bin size will be MUCH smaller.)

cons:
- potential platform integration issues due to dart SDK version conflicts

fzyzcjy: Looks great (as mentioned in https://github.com/dart-sys/dart-sys/issues/48#issuecomment-1409534097)

> potential platform integration issues due to dart SDK version conflicts

Hmm that is a problem. However, IIRC, dart sdk is backward compatible, and is not upgraded very frequently? Thus, we may be able to stick to a *a bit older* version of dart sdk, such as the Dart for Flutter 3.3?

Issue #999, opened 2023-01-26
how to get debug log for  project `frb_codegen` ?


dbsxdbsx:
I am trying to add some feature to `frb_codegen` using vs code, but the debugging is a little frustrating. Since I don't know where the output is after changing code, then `build.rs` run automatically, though there are many `info!`, `debug!` in code.  

I guess it is related to:
```rust
 let raw_opts = RawOpts::parse();
    env_logger::Builder::from_env(Env::default().default_filter_or(if raw_opts.verbose {
        "debug"
    } else {
        "info,cbindgen=error"
    }))
    .init();
```

I hope there would be log output on local disk after  `build.rs` running.

fzyzcjy:
`build.rs` is a standard thing in Rust ecosystem, so I guess maybe search how to see logs of a standard rust build.rs

dbsxdbsx:
There is nothing with `build.rs`, `build.rs` of some sub examples are just triggered when code in `frb_codegen` is changed.    I  wonder how to catch output info when running `frb_codegen`.
Currently,  I just write some debug code like:
```rust
   let filepath = "./output.txt";
    std::fs::write(filepath, data).expect("Unable to write file");
```
then after `build.rs` auto-triggered, modified `frb_codegen` is running for those examples, so that I can catch data on local disk, but it is still a little inconvenient.

fzyzcjy:
> how to catch output info when running frb_codegen

Not very sure about the question. Maybe you mean `RUST_LOG=debug flutter_rust_bridge_codegen your_args`?

Related: bug issue template

![image](https://user-images.githubusercontent.com/5236035/214838397-9cbe582e-1455-402a-a441-1c523cb4fe64.png)

dbsxdbsx:
I solved it by using crate `fern`.

PR #1019, opened 2023-02-01
Add feature: log to disk 

dbsxdbsx:
## What's new
add crates `fern` and `chono` for `frb_codegen` to support recording debug information to local disk with the existing crate `log`---yes, this is a feature only for developers.

Now, with this feature, it is easy to record information in an example project, whenever `frb_codegen` is changed. The only thing to do is to add:
```rust
use lib_flutter_rust_bridge_codegen::init_logger;
...
fn main(){
init_logger("./logs/").expect("failed to initial log");
...
}
```
to the `main.rs` of the example needed to catch log.

Like I did for `pure_dart` and `pure_dart_multi`.  Now, whenever the 2 examples are trigger to rebuild, there would be a directory called `logs`, where logs would be recorded in the format of date, like `20230201`.

Hopefully, this feature could let maintaining frb become easier.
## log picture
Snipaste_2023-02-01_19-59-32
image

This log is shown in vscode with a certain extension, so it is colored.

## related issue
https://github.com/fzyzcjy/flutter_rust_bridge/issues/999


fzyzcjy:
That's interesting!

My main concern that we should not add this init_logger into frb_codegen is that, it is not related to frb_codegen itself, instead it is the user who can decide how and where to log. For example, one user may decide to log to disk while another one may decide to log to somewhere else. 
However, this may be acceptable, since otherwise we have to duplicate code. What do you think?

fzyzcjy:
By the way, would be great to update doc to mention this a little bit

dbsxdbsx:
> My main concern that we should not add this init_logger into frb_codegen is that, it is not related to frb_codegen itself, 

At present,  the function `init_logger`  is added in `logs.rs` in project `frb_codegen` as an extra module, which implies that it is intended to log to disk for `frb_codegen`, not for other projects in frb, since `frb_codegen` is the most important one.  If you mean  `init_logger` should not only serve `frb_codegen` but also others like `frb_rust` and `frb_dart`,  I have no idea where to put the `init_logger` into.

>For example, one user may decide to log to disk while another one may decide to log to somewhere else.

First, the function `init_logger` was just added to examples `pure_dart_multi` and `pure_dart`, which are used for debugging only, so it would not ruin the behavior of crate `log` when using frb for final users. Second, for developers, the extra statement `init_logger("./logs/").expect("failed to initial log");` in the 2 examples is what I think ok for default, if else where someone what to log, only changing the path from `logs` to other paths is enough.

dbsxdbsx:
> By the way, would be great to update doc to mention this a little bit

Which section of the doc book should be mentioned with this feature?

frb_codegen/src/logs.rs
```diff
@@ -0,0 +1,37 @@
...
+        .apply()?;
+
+    #[cfg(not(debug_assertions))]
+    d.level(log::LevelFilter::Info)
+        .chain(fern::DateBased::new(path, "%Y-%m-%d.log"))
+        .apply()?;
+
+    std::panic::set_hook(Box::new(|m| {
+        log::error!("{}", m);
+    }));
```
fzyzcjy, line 34: Hmm set_hook "replacing any that was previously registered". So would be better to also call the old hook first

fzyzcjy:
> Which section of the doc book should be mentioned with this feature?

Have not come up with a good place. Maybe just https://cjycode.com/flutter_rust_bridge/feature/misc.html? A bit better than nothing ;)

dbsxdbsx:
It is strange, the doc update triggered CI test, and one of them failed.

fzyzcjy:
No worries, let me run it

dbsxdbsx:
It is still failing, maybe a bug with GitHub action.

fzyzcjy:
weird, let's ignore it for now

dbsxdbsx:
@fzyzcjy, I've done it. Ready for a review.

frb_codegen/Cargo.toml
```diff
@@ -29,6 +29,8 @@ extend = "1.1.2"
 itertools = "0.10.3"
 lazy_static = {workspace = true}
 log = "0.4"
+fern = { version = "0.6.0", features = ["date-based"] }
```
fzyzcjy, line 32: nit: We already have `env_logger` and has setup that one; now we also use `fern`. Will we have conflicts, or can we reduce such two duplicated crates into one?

fzyzcjy (commented):
Ready to merge, just one nit!

dbsxdbsx:
@fzyzcjy, please restart the CI, it got several issues with the rustup.

fzyzcjy:
If we are going to use fern instead of env_logger, we may have to update https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/.github/ISSUE_TEMPLATE/bug_report.yaml as well: Originally there are several places saying:

> ... with `RUST_LOG=debug` ...

but now seems we need to follow fern's style of specifying "output debug log level"

dbsxdbsx:
> 

No need to change it, because I've set checking `RUST_LOG` in `init_logger(...)`.

fzyzcjy:
Great! Forget to check the init_logger impl just now

frb_codegen/src/logs.rs
```diff
@@ -1,37 +1,63 @@
...
+        };
+        out.finish(format_args!("{}", format))
+    });
 
-    std::panic::set_hook(Box::new(|m| {
-        log::error!("{}", m);
-    }));
+    let log_level = match std::env::var("RUST_LOG") {
+        Ok(val) => val,
+        Err(_) => if verbose { "debug" } else { "info" }.to_owned(),
```
fzyzcjy, line 45: uber-nit: `unwrap_or`

frb_codegen/src/logs.rs
```diff
@@ -1,37 +1,63 @@
...
+            .chain(fern::DateBased::new(path, "%Y-%m-%d.log"))
+            .chain(std::io::stdout())
+            .apply()?;
+    } else {
+        // #[cfg(not(debug_assertions))]
+        d.level(LevelFilter::Info)
+            .level_for("cbindgen", LevelFilter::Error)
+            .chain(std::io::stdout())
+            .apply()?;
+    }
```
fzyzcjy, line 60: uber-nit: maybe

```
match log_level {
debug => ...
info => ...
_ => panic!("only allow debug and info")
}

**Comment by dbsxdbsx** [comment]
@fzyzcjy, all check is passed except for the one with the rustup issue.😀

File: .github/ISSUE_TEMPLATE/bug_report.yaml [hunk 5]
```diff
@@ -22,8 +22,8 @@ body:
     id: codegen-logs
     attributes:
       label: Codegen logs with `RUST_LOG=debug` environment variable
-      description: Run code generator with `RUST_LOG` environment variable set to `debug`. For example, `RUST_LOG=debug flutter_rust_bridge_codegen your_args`.
-      placeholder: Run code generator with `RUST_LOG` environment variable set to `debug`. For example, `RUST_LOG=debug flutter_rust_bridge_codegen your_args`.
+      description: Run code generator with `RUST_LOG` environment variable set to `debug`. For example, `RUST_LOG=debug flutter_rust_bridge_codegen your_args`.Alternatively, add `--verbose` to command `flutter_rust_bridge_codegen` would get the same output.
+      placeholder: Run code generator with `RUST_LOG` environment variable set to `debug`. For example, `RUST_LOG=debug flutter_rust_bridge_codegen your_args`.Alternatively, add `--verbose` to command `flutter_rust_bridge_codegen` would get the same output.
```
fzyzcjy, line 26:
```suggestion
      description: Run code generator with `RUST_LOG` environment variable set to `debug`. For example, `RUST_LOG=debug flutter_rust_bridge_codegen your_args`. Alternatively, add `--verbose` to command `flutter_rust_bridge_codegen` would get the same output.
      placeholder: Run code generator with `RUST_LOG` environment variable set to `debug`. For example, `RUST_LOG=debug flutter_rust_bridge_codegen your_args`. Alternatively, add `--verbose` to command `flutter_rust_bridge_codegen` would get the same output.
```

book/src/feature/misc.md
```diff
@@ -36,3 +35,21 @@ flutter_rust_bridge = { version = "..", default-features = false, features = [..
 ```
 
 The `wasm-start` feature governs this behavior and is enabled by default.
+
+## Logging for developers
+
+For developers who want to contribute to this project, here is the feature logging that needs to mention.
+
+When the code in `frb_codegen` is modified, usually developers want to build and run it locally for testing. Now with the `./logs.rs/init_logger` from `frb_codegen`, it is easy to do so.  Take `frb_example/pure_dart` as an example, in `./rust/build.rs`, with:
```
fzyzcjy, line 43:
```suggestion
When the code in `frb_codegen` is modified, usually developers want to build and run it locally for testing. Now with the `init_logger` in `logs.rs` from `frb_codegen`, it is easy to do so.  Take `frb_example/pure_dart` as an example, in `./rust/build.rs`, with:
```

fzyzcjy (approved):

Issue #1014, opened 2023-01-31
[Bug] About Multiple files `.h` redefinition


WangJiaJun8922:
### Describe the bug

hi, I want to use multiple files to define separately.
I followed the method of the `build.rs` file in the `pure_dart_multi` example, but the `.h` header file generated by export will repeat the definition.

- api_1.rs
```rust
pub fn add(left: usize, right: usize) -> usize {
    left + right
}
```

- api_2.rs
```rust
pub fn sub(left: usize, right: usize) -> usize {
    left - right
}
```

- build.rs
```rust
use lib_flutter_rust_bridge_codegen::{
    config_parse, frb_codegen, get_symbols_if_no_duplicates, RawOpts,
};

/// Path of input Rust code
const RUST_INPUT_1: &str = "src/api_1.rs";
const RUST_INPUT_2: &str = "src/api_2.rs";
/// Path of output generated Dart code
const DART_OUTPUT_1: &str = "../dart/lib/bridge_generated_api_1.dart";
const DART_OUTPUT_2: &str = "../dart/lib/bridge_generated_api_2.dart";
/// Path of output Rust code
const RUST_OUTPUT_1: &str = "src/generated_api_1.rs";
const RUST_OUTPUT_2: &str = "src/generated_api_2.rs";
/// Class name to use in dart, corresponding to each Rust block
const CLASS_NAME_1: &str = "ApiClass1";
const CLASS_NAME_2: &str = "ApiClass2";

fn main() {
    // Tell Cargo that if the input Rust code changes, to rerun this build script.
    println!("cargo:rerun-if-changed={}", RUST_INPUT_1);
    println!("cargo:rerun-if-changed={}", RUST_INPUT_2);
    // Options for frb_codegen
    let mut raw_opts = RawOpts {
        // Path of input Rust code
        rust_input: vec![RUST_INPUT_1.to_string(), RUST_INPUT_2.to_string()],
        // Path of output generated Dart code
        dart_output: vec![DART_OUTPUT_1.to_string(), DART_OUTPUT_2.to_string()],
        // Path of output Rust code
        rust_output: Some(vec![RUST_OUTPUT_1.to_string(), RUST_OUTPUT_2.to_string()]),
        wasm: true,
        // Class name of each Rust block of api
        class_name: Some(vec![CLASS_NAME_1.to_string(), CLASS_NAME_2.to_string()]),
        dart_format_line_length: 120,
        // for other options use defaults
[...]

Issue #987, opened 2023-01-17
[Bug] - Mirroring External Types with multiple files - generates Redundant types in the dart code


dinbtechit:
### Describe the bug

Mirroring External Types with multiple files - generates Redundant types in the dart code

**my_types.rs**
```rs
#[frb(mirror(Configuration))]
#[derive(Default, Clone)]
pub struct _Configuration {
    pub id: u32,
    pub name: String
}
```

**api1.rs**
```rs
pub fn update_configuration(config: Configuration) -> anyhow::Result<()> {
   
}
```

**api2.rs**
```rs
pub fn perform_some_operaton(config: Configuration) -> anyhow::Result<()> {
   
}
```

```rs
// build.rs contains all rust input and output files
cargo build
```

### Codegen logs with `RUST_LOG=debug` environment variable

```shell
Finished dev [unoptimized + debuginfo] target(s) in 10.40s
```

### Generated binding code

Api1.dart

```dart
abstract Api1Class {
 . . .
}
 class Configuration {     <----- this class is generated in both Api1 and Api2
   . . .
 } 

class Api1ClassImpl implements Api1Class {
 
   . . . 
  Future<void> updateConfiguration (Configuration config) {

  }
  
}
```

**Api2.dart**
```dart
abstract Api2Class {
 . . .
}
 class Configuration {     <----- this class is generated in both Api1 and Api2
   . . .
 } 

class Api2ClassImpl implements Api2Class {
 
   . . . 
  Future<void> performSomeOtherAction(Configuration config) {
    . . .
  }
   
  
}
```

### OS

MacOS

### Version of `flutter_rust_bridge_codegen`

1.61.*

fzyzcjy:
Looks like a bug. cc @dbsxdbsx who implemented the multi-file feature :)

dbsxdbsx:
> Looks like a bug. cc @dbsxdbsx who implemented the multi-file feature :)

Just as I stated [here](https://github.com/fzyzcjy/flutter_rust_bridge/issues/956#issuecomment-1381778631), I wonder how to handle the customized struct.   
Since the customized struct would be used in both rust and dart code over different API blocks(files), duplicately defining them by adding suffixes to the customized struct name is not a good idea. Then, if this kind of structs is defined only once, where to place them?   Putting them in an auto-generated file say, `auto_genearated_shares.rs`?  If following this way, is `auto_genearated_shares.dart` needed?

dinbtechit:
[...]

PR #1024, opened 2023-02-04
Fix the duplicated "dummy_method_to_enforce_bundling" in C header within multi-blocks

dbsxdbsx:
This PR fixes https://github.com/fzyzcjy/flutter_rust_bridge/issues/1014. And I think the [idea](https://github.com/fzyzcjy/flutter_rust_bridge/issues/1014#issuecomment-1410657254) @WangJiaJun8922 proposed is correct (thanks~), since `dummy_method_to_enforce_bundling` has nothing to do with the shared stuff in the context of multi-blocks. In any case, giving `dummy_method_to_enforce_bundling` a suffix of API block/class name makes sense, unless `--class-name` is not specified.

In addition, the `bail` function is refined to make log output consistent.


dbsxdbsx:
I realized that for Mac/Ios, the name for presetting `dummy_method_to_enforce_bundling` should be changed, too.

fzyzcjy:
> I realized that for Mac/Ios, the name for presetting dummy_method_to_enforce_bundling should be changed, too.

Why? would be great to know more details

> By the way, where does FlutterRustBridgeExample come from for example with_flutter?

Which FlutterRustBridgeExample - may need to point out the exact line of code

dbsxdbsx:
>Why? would be great to know more details

Referring to the [CI issue](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/4089272432/jobs/7051798455)from macOS, the signature is still `dummy_method_to_enforce_bundling`, but not ` dummy_method_to_enforce_bundling_FlutterRustBridgeExample`. Maybe the swift file need to be changed before generation.

>Which FlutterRustBridgeExample - may need to point out the exact line of code
In example `with_flutter`,  `let c_dummy_code = generator::c::generate_dummy(&effective_func_names, &config.class_name);
`, the `config.class_name` is `FlutterRustBridgeExample`, but I didn't find where the specific name is pushed in for this example(I checked CI test yaml, justfile, build.rs).

fzyzcjy:
Hmm, so is it possible to generate one extra function, `dummy_method_to_enforce_bundling`, which calls the (possibly multiple) `dummy_method_to_enforce_bundling_*`?

For example, I see you generate two functions:

image


Can we generate one more:

```
static int64_t dummy_method_to_enforce_bundling(void) {
      int64_t dummy_var = 0;
    dummy_var ^= dummy_method_to_enforce_bundling_apiclass1();
    dummy_var ^= dummy_method_to_enforce_bundling_apiclass2();
    return dummyvar
```

etc.

This function should only be generated once (say, in the first file), but should call *all* dummy_method_to_enforce_bundling_*

fzyzcjy:
By doing so, we avoid doing breaking changes :)

dbsxdbsx:
Within https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/a9b56702e26d3f829e10550238c2fcc5508496ae,  there are 3 things needed to mention:
1. the signature for `frb_codegen` is refined to include the whole configs, which is essential for catching global information when generating for a specific config. Also, this modification paves the way for future refinement of [the shared stuff](https://github.com/fzyzcjy/flutter_rust_bridge/issues/987).
2. replace crate `extend` with trait mechanism, since it can't be used if it is defined in another module.
3. For test purpose, add/refine the output of c-header for both "pure_dart" and "pure_dart_multi".

dbsxdbsx:
@fzyzcjy, running "just precommit" on my win10 machine with flutter 3.7.1.  The `pubspec.lock` would output `dart: ">=2.18.0 <3.0.0`, But CI would output `>=2.18.0 <4.0.0`. Is there something need to change?

fzyzcjy:
> The pubspec.lock would output dart: ">=2.18.0 <3.0.0, But CI would output >=2.18.0 <4.0.0. Is there something need to change?

Hmm maybe check the flutter and dart versoin in the CI? I am not sure why that happens

dbsxdbsx:
> 

I did so at first, but then there are inconsistent format issues with dart format. By the way, is it ok to use the latest flutter/dart in CI? I saw flutter 3.7 appear in the previous pr.

fzyzcjy:
> I did so at first, but then there are inconsistent format issues with dart format

Hmm...

> is it ok to use the latest flutter/dart in CI?

IMHO yes. What is your opinion?

dbsxdbsx:
> IMHO yes. What is your opinion?
At present, what about making the flutter version of CI to be 3.7.1.

fzyzcjy:
Sure!

fzyzcjy:
CI is fixed in another PR so by merging the master maybe we are ready to go :)

dbsxdbsx:
The format issue still exists for ubuntu. Maybe I need to do `just precommit` in a Linux system.  Anyway, @fzyzcjy, could you please check if the  rust logic is ok?

fzyzcjy:
Sure, I will probably review it within a day!

temeddix:
Right now `just precommit` is a bit broken. Post-PR https://github.com/fzyzcjy/flutter_rust_bridge/pull/1051 is waiting to be merged, as this allows `just precommit` to work again.

fzyzcjy:
> waiting to be merged

Done ;) (given it is green)

dbsxdbsx:
@temeddix， thanks for your fixed CI issue. But here is one thing to mention [this format issue](https://github.com/dbsxdbsx/flutter_rust_bridge/actions/runs/4142325893/jobs/7162872500). on win10, with flutter 3.7.3:
This line in `with_flutter` during `just precommit` would first format into:
```dart
external dynamic /* void */ wire_off_topic_memory_test_output_zero_copy_buffer(NativePortType port_, int len);
```
Then, in afterward task, it would be reformatted back to 
```dart
external dynamic /* List<dynamic> */
      wire_handle_zero_copy_vec_of_primitive_sync(int n);
```

fzyzcjy:
I am also reproducing this problem. I will fix it in https://github.com/fzyzcjy/flutter_rust_bridge/issues/1053 - just wait until I fix and update master :)

temeddix:
Good :) I will be busy for a while..

fzyzcjy:
@temeddix take your time :)

frb_example/pure_dart/rust/c_output_path_bak/c_output.h
```diff
@@ -0,0 +1,1149 @@
+#include <stdbool.h>
```
fzyzcjy, line 1: uber-nit: what is "c_output_path_bak"? e.g. rename to "c_output_path_extra"?

frb_codegen/src/lib.rs
```diff
@@ -40,10 +40,19 @@ mod parser;
...
 mod target;
 mod transformer;
-mod utils;
+pub mod utils;
 use error::*;
 
-pub fn frb_codegen(config: &config::Opts, all_symbols: &[String]) -> anyhow::Result<()> {
+pub fn frb_codegen(
+    configs: &[config::Opts],
+    index: usize,
```
fzyzcjy, line 48: Hmm shall we put `index` here? Or is it possible to be inside `configs` etc

frb_codegen/src/utils.rs
```diff
@@ -89,10 +89,59 @@ impl Display for BlockIndex {
...
+/// let output_1 = relative_path("./a/b/c", "./a/d/e");
+/// assert_eq!(std::path::PathBuf ::from(output_1), std::path::PathBuf ::from("../../d/e"));
+/// let output_2 = relative_path("./a/d/e", "./a/b/c");
+/// assert_eq!(std::path::PathBuf ::from(output_2), std::path::PathBuf ::from("../../b/c"));
+/// let output_3 = relative_path("./a/b/c", "./a/b/c/d");
+/// assert_eq!(std::path::PathBuf ::from(output_3), std::path::PathBuf ::from("d"));
+/// let output_4 = relative_path("./a/b/c", "./a/b/c/");
+/// assert_eq!(std::path::PathBuf::from(output_4),std::path::PathBuf::from(""));
+/// ```
+pub fn relative_path(from: &str, to: &str) -> String {
```
fzyzcjy, line 121: would be great to have some `unit tests` for this utils.rs's functions (if we really want to manually write down this)

fzyzcjy, line 121: uber-nit: Shall we use `Path` instead of str/string

fzyzcjy, line 121: uber-nit: I guess there should already exists some crate that does this, so we do not need to write by hand?

frb_codegen/src/generator/c/mod.rs
```diff
@@ -1,14 +1,84 @@
...
+use crate::utils::{relative_path, BlockIndex, PathExt};
+
+pub fn generate_dummy(
+    config: &config::Opts,
+    configs: &[config::Opts],
+    func_names: &[String],
+    c_path_index: usize,
+) -> String {
+    match configs.len() > 1 {
+        true => {
```
fzyzcjy, line 13: uber-nit: if-else

frb_codegen/src/generator/c/mod.rs
```diff
@@ -1,14 +1,84 @@
-pub fn generate_dummy(func_names: &[String]) -> String {
+use std::path::Path;
+
+use crate::config;
+use crate::utils::{relative_path, BlockIndex, PathExt};
+
+pub fn generate_dummy(
+    config: &config::Opts,
+    configs: &[config::Opts],
```
fzyzcjy, line 8: hmm quite confused about this... maybe we can refactor to a better function signature

frb_codegen/src/generator/c/mod.rs
```diff
@@ -1,14 +1,84 @@
...
+                    .skip(1)
+                    .map(|e| {
+                        // get directory only from paths
+                        let src_p = Path::new(&config.c_output_path[c_path_index]);
+                        let src_dir = src_p.directory_name_str().unwrap();
+                        let dst_p = Path::new(&e.c_output_path[c_path_index]);
+                        let dst_dir = dst_p.directory_name_str().unwrap();
+                        // get reletive path and head file name
+                        let relative_p = relative_path(src_dir, dst_dir);
+                        let head_file_name = dst_p.file_name_str().unwrap();
```
fzyzcjy, line 32: uber-nit: "head" -> "header" (occurred in many places, would be great to fix all of them)

frb_example/pure_dart_multi/rust/c_output_path/c_output_1.h
```diff
@@ -34,4 +34,11 @@ static int64_t dummy_method_to_enforce_bundling(void) {
     dummy_var ^= ((int64_t) (void*) drop_dart_object);
     dummy_var ^= ((int64_t) (void*) new_dart_opaque);
     return dummy_var;
+}
+#include "c_output_2.h"
+static int64_t dummy_method_to_enforce_bundling(void) {
+    int64_t dummy_var = 0;
+    dummy_var ^= ((int64_t) (void*) dummy_method_to_enforce_bundling_ApiClass1);
+    dummy_var ^= ((int64_t) (void*) dummy_method_to_enforce_bundling_ApiClass2);
+    return dummy_var;
```
fzyzcjy, line 43: I am not an expert in C, so the question is: c_output_2.h has dummy_method_to_enforce_bundling_ApiClass2 in *header* file with *static*, and here we are using its *function pointer* instead of actually calling it. So,

1. is it sufficient to forbid the function be stripped? (I guess yes?)
2. is it safe or dangerous to do this? (I guess safe?)

fzyzcjy (commented):
Good job! Mostly are nits :)

frb_codegen/src/utils.rs
dbsxdbsx: I didn't find someone to do so.

frb_example/pure_dart_multi/rust/c_output_path/c_output_1.h
dbsxdbsx: I got similar wonders like you, that is why the CI test is essential.

frb_codegen/src/lib.rs
dbsxdbsx: We know there is a field `block_index` in `Opts`,  so there is already a field in charge of the index. And the reason why using  ` index: usize,` instead of `index: BlockIndex`(or just omit the parameter) is that I want to double-check the order of the current config is consistent with  that in `all_symbols` in `main.rs`, therefore  here comes  the double-check code:
```rust
pub fn frb_codegen(
    configs: &[config::Opts],
    index: usize,
    all_symbols: &[String],
) -> anyhow::Result<()> {
    let config = configs
        .iter()
        .find(|each| each.block_index == BlockIndex(index))
        .unwrap();
```
If the `unwrap()` doesn't make panic, that means it is safe to use `config.block_index`  afterwards for this config routine.

frb_codegen/src/utils.rs
dbsxdbsx: Actually, the doc example here:
```rust
/// # Examples
/// ```
/// use lib_flutter_rust_bridge_codegen::utils::relative_path;
/// let output_1 = relative_path("./a/b/c", "./a/d/e");
/// assert_eq!(std::path::PathBuf ::from(output_1), std::path::PathBuf ::from("../../d/e"));
/// let output_2 = relative_path("./a/d/e", "./a/b/c");
/// assert_eq!(std::path::PathBuf ::from(output_2), std::path::PathBuf ::from("../../b/c"));
/// let output_3 = relative_path("./a/b/c", "./a/b/c/d");
/// assert_eq!(std::path::PathBuf ::from(output_3), std::path::PathBuf ::from("d"));
/// let output_4 = relative_path("./a/b/c", "./a/b/c/");
/// assert_eq!(std::path::PathBuf::from(output_4),std::path::PathBuf::from(""));
/// ```
```
is already unit tests,  rust compiler is so restricted that the compilation would not be passed if just omitting `use lib_flutter_rust_bridge_codegen::utils::relative_path;` here.

fzyzcjy: What about: 

1. https://stackoverflow.com/questions/39340924/given-two-absolute-paths-how-can-i-express-one-of-the-paths-relative-to-the-oth ie https://crates.io/crates/pathdiff
2. https://docs.rs/relative-path/latest/relative_path/struct.RelativePath.html#method.relative ie https://crates.io/crates/relative-path

fzyzcjy: Hmm pathdiff already in the cargo.toml (https://github.com/fzyzcjy/flutter_rust_bridge/blob/864675c5ba5269f0f665045c43cb0818bb9b9f97/frb_codegen/Cargo.toml#L34) so let's use it

frb_codegen/src/lib.rs
fzyzcjy: Hmm... The API looks a bit less than ideal IMHO :/

frb_example/pure_dart_multi/rust/c_output_path/c_output_1.h
dbsxdbsx: > I am not an expert in C, so the question is: c_output_2.h has dummy_method_to_enforce_bundling_ApiClass2 in _header_ file with _static_, and here we are using its _function pointer_ instead of actually calling it. So,
> 
> 1. is it sufficient to forbid the function be stripped? (I guess yes?)
> 2. is it safe or dangerous to do this? (I guess safe?)

I do think this concern is the most important one for this pr.  Frankly, I am not sure what `dummy_method_to_enforce_bundling` acts here.  Intuitionally, I treat it as a "bridge" that tells dart and rust how to translate information between each other.  

And since https://github.com/fzyzcjy/flutter_rust_bridge/releases/tag/v1.59.0, I notice the c headers are only used for compilation within in MacOs or Ios, and it is not used once generation files have been produced by frb in other platforms. So I guess this concern is really something related to MacOs and Ios.  

In addition, I wonder why no issue is posted for multi-blocks for MacOs or Ios, since https://github.com/fzyzcjy/flutter_rust_bridge/releases/tag/v1.59.0 until https://github.com/fzyzcjy/flutter_rust_bridge/issues/1014.

dbsxdbsx: Actually, I am thinking of something similar when designing for the shared task. I am wondering if  it is essential and feasible to define an auto-generated header file `shared_auto_c_header.h` to contains only  
```c
#include "c_output_1.h"
#include "c_output_2.h"
static int64_t dummy_method_to_enforce_bundling(void) {
    int64_t dummy_var = 0;
    dummy_var ^= ((int64_t) (void*) dummy_method_to_enforce_bundling_ApiClass1);
    dummy_var ^= ((int64_t) (void*) dummy_method_to_enforce_bundling_ApiClass2);
    return dummy_var;
...
```
for a multi-blocks

fzyzcjy: > I do think this concern is the most important one for this pr. 

Yes - as I mentioned most are nits ;)

> Frankly, I am not sure what dummy_method_to_enforce_bundling acts here.

Shortly speaking, it avoids "symbol stripping" in ios. When ios thinks a function is not used at all, it may throw it away - then Flutter can no longer use that C function. By creating such a weird dummy_method_to_enforce_bundling, we make fake use of those generated C functions, then ios will not strip them.

> So I guess this concern is really something related to MacOs and Ios.

Yes, symbol stripping seems only a problem on ios/macos (if using static lib IIRC)

> In addition, I wonder why no issue is posted for multi-blocks for MacOs or Ios, since [v1.59.0 (release)](https://github.com/fzyzcjy/flutter_rust_bridge/releases/tag/v1.59.0) until https://github.com/fzyzcjy/flutter_rust_bridge/issues/1014.

I guess maybe people do not use this feature that often?

fzyzcjy: stripping: https://docs.flutter.dev/development/platform-integration/ios/c-interop#stripping-ios-symbols , https://cjycode.com/flutter_rust_bridge/integrate/ios_headers.html etc

fzyzcjy: > shared_auto_c_header.h to contains only

Then that may be a breaking change

dbsxdbsx: After a second thought, I answer is:
> is it sufficient to forbid the function be stripped? (I guess yes?)

Yes. I think the "strip" issue should be thought of as something related to optimization for c/c++ code related to a specific c/c++ compiler in use. In this way, as long as there is no optimization in the single-block case, the only concern left is the `static` mark for a function symbol.  

As far as I know, when a c/cpp function symbol is marked as `static` excluding the case in a class/struct, it  only means 2 things:
1, there is only 1 instance of such a thing when using it;
2, it is only visible in the current file which defined the static stuff.

That means, with static, there could be functions with the same names in each respective visible scope, like the no-conflict "Parallel universes"([Reference](https://www.scaler.com/topics/static-function-in-c/)). But this is not the case when using FFI here---there is a conflict.

It seems that the `static` has nothing to do with code optimization since the relative address in memory of a function would not be changed whether it is defined as `static` or not. But anyway, I suggest not using `static` in any case(single/multi-blocks), unless it is necessary.

Here, for the multi-block example, I think that marking `dummy_method_to_enforce_bundling_*(void)` without `static` can make them more like a manually defined function API for easier analysis, even for the single-block case.

>is it safe or dangerous to do this? (I guess safe?)

In the aspect of memory, if you are concerning the "dangling pointer", I think it is safe. Because the heap addresses used to initialize/store those API functions would not be changed until APP termination, as long as these addresses are recognized(not optimized out by the c/cpp compiler) to initiate.

fzyzcjy: > marking dummy_method_to_enforce_bundling_*(void) without static

Then if users include it in multi places I am worried maybe it will cause problems? Not an expert in C, but IIRC a function definition in header files should be static b/c this

dbsxdbsx: > Then if users include it in multi places I am worried maybe it will cause problems? Not an expert in C, but IIRC a function definition in header files should be static b/c this

I guess [this](https://stackoverflow.com/questions/10847176/should-i-define-static-inline-methods-in-header-file) is the answer.  Roughly, both using `inline` and `static` could avoid conflicts, but the first one is doing so by linking and the latter one is doing so by copying.

So I think it is appropriate to add `inline` for any external linkable function when the function is both declare and defined together in the same file. 

Do I miss something else?

frb_codegen/src/lib.rs
dbsxdbsx: >  a bit less than ideal 
what do you want for this function signature? What is missing?

frb_codegen/src/generator/c/mod.rs
dbsxdbsx: If you think too many "config" is miss leading, then what if change  it  to :
```rust
pub fn generate_dummy(
    configs: &[config::Opts],
    index: usize,
    func_names: &[String],
    c_path_index: usize,
) -> String {
```
But here, it would once again try to fetch current config in side the function. I would suggest just add a doc to clarify what each parameters are used for.

frb_codegen/src/lib.rs
fzyzcjy: Hopefully we can avoid passing such `index`. e.g. Is it possible we just pass in `config: config::Opts` etc

dbsxdbsx: > Is it possible we just pass in config: config::Opts

IMHO, no,  the configs are essential to catch global info. The global info is in used in `generate_dummy`. And the global info would be especially useful when doing [the shared task.](https://github.com/fzyzcjy/flutter_rust_bridge/issues/987#issuecomment-1407940083)

frb_codegen/src/generator/c/mod.rs
dbsxdbsx: fixed in https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/cbd8da41d3f47156452430b5a2774aa471e2ddb2.

frb_codegen/src/generator/c/mod.rs
dbsxdbsx: fixed in https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/cbd8da41d3f47156452430b5a2774aa471e2ddb2.

frb_codegen/src/utils.rs
dbsxdbsx: Fixed in https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/d72716876f438134dad99b2781564f6beb682819.

dbsxdbsx: Since https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/d72716876f438134dad99b2781564f6beb682819,  this function is no longer used, it is deleted in https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/499e17d475739320b5aff952cd6f3acb3fc6ed5e.

frb_example/pure_dart/rust/c_output_path_bak/c_output.h
dbsxdbsx: Fixed in https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/cbd8da41d3f47156452430b5a2774aa471e2ddb2.

frb_codegen/src/generator/c/mod.rs
fzyzcjy: (same as below - will revisit later)

frb_codegen/src/lib.rs
fzyzcjy: I will revisit this a bit later when having a bit of time. This may need a bit of thinking

dbsxdbsx:
I don't understand the [issue](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/4164089778/jobs/7205310340) from CI.

frb_codegen/src/lib.rs
fzyzcjy: Revisiting this problem. The new code

```
pub fn frb_codegen(
    configs: &[config::Opts],
    index: usize,
    all_symbols: &[String],
) -> anyhow::Result<()> {
    let config = configs
        .iter()
        .find(|each| each.block_index == BlockIndex(index))
        .unwrap();
```

looks equivalent to the old code https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_codegen/src/lib.rs

```
pub fn frb_codegen(config: &config::Opts, all_symbols: &[String]) -> anyhow::Result<()> {
```

especially given that, the `index` is never used in other places.

Thus, is it possible to keep the old function signature?

In addition, if we change signature it is a breaking change for everyone

fzyzcjy: Alternatively what about this:

```
// an extra function, whoever wants multi-file uses this instead
pub fn frb_codegen_multi(configs: &[config::Opts], all_symbols) {
  for index in 0..something {
    frb_codegen(configs.iter().find(|each| each.block_index == BlockIndex(index)).unwrap(), all_symbols)?;
  }
}
```

frb_codegen/src/generator/c/mod.rs
fzyzcjy: For this particular one, the configs + index LGTM, since it means "we need to know all configs, and we want to generate a dummy for that index based on other configs"

fzyzcjy:
> I don't understand the [issue](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/4164089778/jobs/7205310340) from CI.

Well you may debug it (locally?), try to retry running it to see whether it disappears, or revert relavent changes if non critical

frb_codegen/src/lib.rs
dbsxdbsx: > frb_codegen_multi

Taking this one, I guess you mean defining `frb_codegen_multi(..)` in `lib.rs` right? In the case, it is easy to see `frb_codegen(..)` is enough for `pure_dart`,and only `frb_codegen_multi(..)` is enough for `pure_dart_multi`. 

But what the one in `frb_codegen/src/main.rs`? Since it is not know whether there is single or multi-blocks, I guess using the  `frb_codegen_multi(..)`  one?

In addition, a severe problem at present, since the `dummy` issue needs to be dealt with in the 1st config block, how do I catch the global info from `frb_codegen(...)` following the style?

dbsxdbsx: >In addition, if we change signature it is a breaking change for everyone

By the way, why would you think this signature change is a greate change? As far as I know, there are only 3 places using the signature `frb_codegen(..)`, and inside the new defined `frb_codegen(..)`, after :
```rust
  let config = configs
        .iter()
        .find(|each| each.block_index == BlockIndex(index))
        .unwrap();
```
Everything would be the same, except that you can get global info by `configs` instead of the current `config` when dealing each  block.

fzyzcjy: > But what about the one in frb_codegen/src/main.rs? Since it is not known whether there is single or multi-blocks, I guess using the frb_codegen_multi(..) one?

Seems yes

> In addition, a severe problem at present, since the dummy https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024#discussion_r1109133848 needs to be dealt within the 1st config block, how do I catch the global info from frb_codegen(...) following the style?

Hmm that's a problem

> By the way, why would you think this signature change is a greate change? As far as I know, there are only 3 places using the signature frb_codegen(..)

Because a lot of users of this library may be using this API - it is a *public* API.

dbsxdbsx: > (Maybe this style is enough if all shared stuff could be done in another shared config, but not at present.)

Oh, I am wrong. Even there is another auto-generated shared config intance, it is still essential to catch global information when dealing each regular block. For example, the path for shared block, the names of shared structs....

fzyzcjy: I am busy with my code so let me write down a summarization to reduce brain context switch:

Problem: Suppose we want to generate api_1, api_2, api_3 three parts. Then, how to design the API?
Requirements: (1) Neat API (2) Best to be backward-compatible.

dbsxdbsx: >Because a lot of users of this library may be using this API - it is a public API.
oh, my god, I didn't see this. I thought the signature change this time would be as smooth as I did when introducing multi-blocks.

By the way, without changing signature, I've though introducing global constant variable. But it seems to be not feasible here, because the global instance(say `CONFIGS`) would be used in 
`lib.rs`, but it is not feasible before `frb_codegen` is called.

fzyzcjy: What about this:

```rust
/// Need to add a line of comment, telling readers `all_configs` is used because need global information
pub fn frb_codegen_part_of_multi(
    config: &config::Opts,
    // add `all_` so it is less confusing. together with comments should be clearer to readers
    all_configs: &[config::Opts],
) {
 // the real main code
}

// this API is (1) a convenient shortcut for non-multi users (2) make it backward compatible
pub fn frb_codegen(config) {
  // just delegate to frb_codegen_part_of_multi
  frb_codegen_part_of_multi(config, vec![config])
}
```

dbsxdbsx: Now, I am looking forward to the feature of default/varadic paramter, If rust has this feature, there would be no public API that needs to be changed.

And for now, is it feasible to let users to use a macro say `frb_codegen!(...)`, so that it would be easy to use either `frb_codegen!(a,b)`,  `frb_codegen!(a,b,c)` and so on.  With this macro, it would be easily to do some break change inside without changing the signature.

fzyzcjy: I guess https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024#discussion_r1109183538 will not introduce any breaking change, and at the same time, allows to get global information (all_configs) easily

dbsxdbsx: > I guess [#1024 (comment)](https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024#discussion_r1109183538) will not introduce any breaking change, and at the same time, allows to get global information (all_configs) easily

If I understand correctly, you are suggesting to refractor inside `frb_codegen(...)`?   If this is the case, where does the `all_configs` come from?

dbsxdbsx: Aside from using macro, there is another way without changing the signature. That is, new a field in `[Opt](config::Opts)` to contain essential global info, but this way just makes code too coupled.

fzyzcjy: Suppose old code on master branch is:

```
pub fn frb_codegen(config) {
 [the-old-200-lines-of-code]
}
```

or, similarly, the code at this time in this PR is:

```
pub fn frb_codegen(configs, index) {
 [the-old-200-lines-of-code]
}
```

the new code is:

```
/// Need to add a line of comment, telling readers `all_configs` is used because need global information
pub fn frb_codegen_part_of_multi(
    config: &config::Opts,
    // add `all_` so it is less confusing. together with comments should be clearer to readers
    all_configs: &[config::Opts],
) {
 [the-old-200-lines-of-code] // JUST SIMPLY MOVE IT TO HERE
}

// this API is (1) a convenient shortcut for non-multi users (2) make it backward compatible
pub fn frb_codegen(config) {
  // just delegate to frb_codegen_part_of_multi
  frb_codegen_part_of_multi(config, vec![config])
}
```

dbsxdbsx: I see. So keeping the same code in `build.rs` for `pure_dart` and `main.rs` for `frb_code_gen`, modify `build.rs`  for `pure_dart_multi` with `frb_codegen_part_of_multi(...)` defined in `lib.rs` of `frb_code_gen`.

dbsxdbsx:
> > I don't understand the [issue](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/4164089778/jobs/7205310340) from CI.
> 
> Well you may debug it (locally?), try to retry running it to see whether it disappears, or revert relavent changes if non critical

I  can't debug it locally, because I don't have an APPLE machine. And unfortunately, the issue still occurs.

fzyzcjy:
Hmm which code causes this, and shall we revert it?

dbsxdbsx:
> Hmm which code causes this, and shall we revert it?
A side question, there is not CI test for `pure_dart_multi` any longer. Are they deleted on purpose?

fzyzcjy:
There should definitely be tests about that! It may be dropped accidentically when I was refactoring https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/justfile. Feel free to fix it (e.g. in another PR)

dbsxdbsx:
> There should definitely be tests about that! It may be dropped accidentically when I was refactoring https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/justfile. Feel free to fix it (e.g. in another PR)

I mean the the example in CI test file. Meanwhile, I am out of time these 2~3 days.

fzyzcjy:
Take your time!

> I mean the the example in CI test file

Could you please point to that line? Not very sure what is example in CI. The (new) CI file indeed executes the `justfile`

dbsxdbsx:
> justfile



> Take your time!
> 
> > I mean the the example in CI test file
> 
> Could you please point to that line? Not very sure what is example in CI. The (new) CI file indeed executes the `justfile`

I mean that I only see `pure_dart`  example in the CI folder,but not `pure_dart_multi`:
image

In addition, I also can't find some command that executes the `justifle`, could you point that out?

fzyzcjy:
Oh the ci folder calls: https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/justfile

dbsxdbsx:
> justfile

which line? I didn't find it.
image

fzyzcjy:
https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/justfile

many places e.g.

image

dbsxdbsx:
> https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/justfile
> 
> many places e.g.
> 
> image

No, I mean which line of which file in the CI folder calls the `justfile`?

fzyzcjy:
Each `just sometihng` in ci.yaml will call the `something` in `/justfile` file. 

image

frb_codegen/src/lib.rs
dbsxdbsx: Fixed in https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/80edccbb1a12c6c21a5de2c78be772536335f670.

frb_codegen/src/generator/c/mod.rs
dbsxdbsx: Fixed in 
 https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/80edccbb1a12c6c21a5de2c78be772536335f670.

dbsxdbsx:
@fzyzcjy, when running  `just dart_linter_pana` in win10, I got issue:
```
$     just dart_linter_pana
flutter pub global activate pana
Package pana is currently active at version 0.21.26.
Resolving dependencies...
The package pana is already activated at newest available version.
To recompile executables, first run `flutter pub global deactivate pana`.
Installed executable pana.
Activated pana 0.21.26.
cd frb_dart && pana.bat --no-warning --line-length 80 --exit-code-threshold 0
INFO       Running `D:\flutter\bin\cache\dart-sdk\bin\dart.exe --no-analytics --version`...
INFO       Running `flutter.bat --suppress-analytics --no-version-check --version --machine`...
INFO       Running `git rev-parse --show-toplevel`...
INFO       Running `D:\flutter\bin\cache\dart-sdk\bin\dart.exe --no-analytics pub get --no-example`...
INFO       Running `D:\flutter\bin\cache\dart-sdk\bin\dart.exe --no-analytics pub outdated --json --up-to-date --no-dev-dependencies --no-dependency-overrides`...
INFO       Analyzing package...
INFO       Running `D:\flutter\bin\cache\dart-sdk\bin\dart.exe --no-analytics analyze --format machine bin`...
INFO       Running `D:\flutter\bin\cache\dart-sdk\bin\dart.exe --no-analytics analyze --format machine lib`...
INFO       Running `git init`...
INFO       Running `git remote add origin https://github.com/fzyzcjy/flutter_rust_bridge`...
INFO       Running `git remote show origin`...
INFO       Running `git fetch --depth 1 --no-recurse-submodules origin master`...
INFO       Running `git ls-tree -r --name-only --full-tree origin/master`...
INFO       Unable to parse `pubspec.yaml` from git repository. Invalid argument(s): Path "frb_dart/pubspec.yaml" is not normalized.
           Invalid argument(s): Path "frb_dart/pubspec.yaml" is not normalized.
           #0      GitLocalRepository._assertPathFormat (package:pana/src/repository/git_local_repository.dart:223:7)
           #1      GitLocalRepository.showStringContent (package:pana/src/repository/git_local_repository.dart:156:5)
           #2      checkRepository.matchRepoPubspecYaml (package:pana/src/repository/check_repository.dart:91:36)
           #3      checkRepository (package:pana/src/repository/check_repository.dart:159:27)
           <asynchronous suspension>
           #4      SharedAnalysisContext.verifyRepository (package:pana/src/package_context.dart:72:24)
           <asynchronous suspension>
           #5      checkPubspecUrls (package:pana/src/references/pubspec_urls.dart:60:30)
           <asynchronous suspension>
           #6      followsTemplate.checkPubspec (package:pana/src/report/template.dart:186:25)
           <asynchronous suspension>
           #7      followsTemplate (package:pana/src/report/template.dart:254:26)
           <asynchronous suspension>
           #8      createReport (package:pana/src/report/create_report.dart:42:5)
           <asynchronous suspension>
           #9      PackageAnalyzer._inspect (package:pana/src/package_analyzer.dart:315:15)
           <asynchronous suspension>
           #10     PackageAnalyzer.inspectDir.<anonymous closure> (package:pana/src/package_analyzer.dart:143:14)
           <asynchronous suspension>
           #11     withTempDir (package:pana/src/utils.dart:236:12)
           <asynchronous suspension>
           #12     main (file:///C:/Users/dbsx/AppData/Local/Pub/Cache/hosted/pub.dev/pana-0.21.26/bin/pana.dart:176:19)
           <asynchronous suspension>
INFO       Running `D:\flutter\bin\cache\dart-sdk\bin\dart.exe --no-analytics format --output=none --set-exit-if-changed --line-length 80 C:\Users\dbsx\AppData\Local\Temp\pana_73d49721\frb_dart\bin`...
INFO       Running `D:\flutter\bin\cache\dart-sdk\bin\dart.exe --no-analytics format --output=none --set-exit-if-changed --line-length 80 C:\Users\dbsx\AppData\Local\Temp\pana_73d49721\frb_dart\lib`...

## ✗ Follow Dart file conventions (20 / 30)
### [x] 0/10 points: Provide a valid `pubspec.yaml`

<details>
<summary>
Failed to verify repository URL.
</summary>

Please provide a valid [`repository`](https://dart.dev/tools/pub/pubspec#repository) URL in `pubspec.yaml`, such that:  

 * `repository` can be cloned,
 * a clone of the repository contains a `pubspec.yaml`, which:,
    * contains `name: flutter_rust_bridge`,
    * contains a `version` property, and,
    * does not contain a `publish_to` property.

Unable to parse `pubspec.yaml` from git repository. Invalid argument(s): Path "frb_dart/pubspec.yaml" is not normalized.
</details>

### [*] 5/5 points: Provide a valid `README.md`


### [*] 5/5 points: Provide a valid `CHANGELOG.md`


### [*] 10/10 points: Use an OSI-approved license

Detected license: `MIT`.

## ✓ Provide documentation (10 / 10)
### [*] 10/10 points: Package has an example


## ✓ Platform support (20 / 20)
### [*] 20/20 points: Supports 6 of 6 possible platforms (**iOS**, **Android**, **Web**, **Windows**, **MacOS**, **Linux**)

* ✓ Android
* ✓ iOS
* ✓ Windows
* ✓ Linux
* ✓ MacOS
* ✓ Web

## ✓ Pass static analysis (30 / 30)
### [*] 30/30 points: code has no errors, warnings, lints, or formatting issues


## ✓ Support up-to-date dependencies (20 / 20)
### [*] 10/10 points: All of the package dependencies are supported in the latest version

|Package|Constraint|Compatible|Latest|
|:-|:-|:-|:-|
|[`args`]|`^2.3.1`|2.4.0|2.4.0|
|[`build_cli_annotations`]|`^2.1.0`|2.1.0|2.1.0|
|[`colorize`]|`^3.0.0`|3.0.0|3.0.0|
|[`js`]|`^0.6.4`|0.6.7|0.6.7|
|[`meta`]|`^1.3.0`|1.9.0|1.9.0|
|[`path`]|`^1.8.1`|1.8.3|1.8.3|
|[`puppeteer`]|`^2.12.0`|2.22.0|2.22.0|
|[`shelf`]|`^1.3.2`|1.4.0|1.4.0|
|[`shelf_static`]|`^1.1.1`|1.1.1|1.1.1|
|[`shelf_web_socket`]|`^1.0.2`|1.0.3|1.0.3|
|[`tuple`]|`^2.0.1`|2.0.1|2.0.1|
|[`uuid`]|`^3.0.6`|3.0.7|3.0.7|
|[`web_socket_channel`]|`^2.2.0`|2.3.0|2.3.0|
|[`yaml`]|`^3.1.1`|3.1.1|3.1.1|

<details><summary>Transitive dependencies</summary>

|Package|Constraint|Compatible|Latest|
|:-|:-|:-|:-|
|[`archive`]|-|3.3.6|3.3.6|
|[`async`]|-|2.10.0|2.10.0|
|[`collection`]|-|1.17.1|1.17.1|
|[`convert`]|-|3.1.1|3.1.1|
|[`crypto`]|-|3.0.2|3.0.2|
|[`http`]|-|0.13.5|0.13.5|
|[`http_parser`]|-|4.0.2|4.0.2|
|[`logging`]|-|1.1.1|1.1.1|
|[`mime`]|-|1.0.4|1.0.4|
|[`petitparser`]|-|5.2.0|5.2.0|
|[`pointycastle`]|-|3.6.2|3.6.2|
|[`pool`]|-|1.5.1|1.5.1|
|[`source_span`]|-|1.9.1|1.9.1|
|[`stack_trace`]|-|1.11.0|1.11.0|
|[`stream_channel`]|-|2.1.1|2.1.1|
|[`string_scanner`]|-|1.2.0|1.2.0|
|[`term_glyph`]|-|1.2.1|1.2.1|
|[`typed_data`]|-|1.3.1|1.3.1|
</details>

To reproduce run `dart pub outdated --no-dev-dependencies --up-to-date --no-dependency-overrides`.

[`args`]: https://pub.dev/packages/args
[`build_cli_annotations`]: https://pub.dev/packages/build_cli_annotations
[`colorize`]: https://pub.dev/packages/colorize
[`js`]: https://pub.dev/packages/js
[`meta`]: https://pub.dev/packages/meta
[`path`]: https://pub.dev/packages/path
[`puppeteer`]: https://pub.dev/packages/puppeteer
[`shelf`]: https://pub.dev/packages/shelf
[`shelf_static`]: https://pub.dev/packages/shelf_static
[`shelf_web_socket`]: https://pub.dev/packages/shelf_web_socket
[`tuple`]: https://pub.dev/packages/tuple
[`uuid`]: https://pub.dev/packages/uuid
[`web_socket_channel`]: https://pub.dev/packages/web_socket_channel
[`yaml`]: https://pub.dev/packages/yaml
[`archive`]: https://pub.dev/packages/archive
[`async`]: https://pub.dev/packages/async
[`collection`]: https://pub.dev/packages/collection
[`convert`]: https://pub.dev/packages/convert
[`crypto`]: https://pub.dev/packages/crypto
[`http`]: https://pub.dev/packages/http
[`http_parser`]: https://pub.dev/packages/http_parser
[`logging`]: https://pub.dev/packages/logging
[`mime`]: https://pub.dev/packages/mime
[`petitparser`]: https://pub.dev/packages/petitparser
[`pointycastle`]: https://pub.dev/packages/pointycastle
[`pool`]: https://pub.dev/packages/pool
[`source_span`]: https://pub.dev/packages/source_span
[`stack_trace`]: https://pub.dev/packages/stack_trace
[`stream_channel`]: https://pub.dev/packages/stream_channel
[`string_scanner`]: https://pub.dev/packages/string_scanner
[`term_glyph`]: https://pub.dev/packages/term_glyph
[`typed_data`]: https://pub.dev/packages/typed_data


### [*] 10/10 points: Package supports latest stable Dart and Flutter SDKs


## ✓ Support sound null safety (20 / 20)
### [*] 20/20 points: Package and dependencies are fully migrated to null safety!


Points: 120/130.
error: Recipe `dart_linter_pana` failed on line 166 with exit code 127
```
What does `Path "frb_dart/pubspec.yaml" is not normalized.` mean?

dbsxdbsx:
@fzyzcjy, everything is done. To sum up:
- refine function `generator::c::generate_dummy` to solve the issue;
- refine  function `frb_codegen` and add new `frb_codegen_multi`, which is also useful for development in the future;
- after testing, still use `static` instead of `inline` for  API within  auto-generated c header file(s); 
- refine doc for using multi-blocks with iOS/MacOS.

book/src/integrate/ios_headers.md
```diff
@@ -52,3 +52,11 @@ and add one ourselves. In the **Build Settings** tab, set the
 
 Finally, use `dummy_method_to_enforce_bundling` somewhere within
 `macos/Runner/AppDelegate.swift`, as long as Xcode does not consider it dead code.
+
+# For multi-blocks
+If there are [multi-blocks](https://cjycode.com/flutter_rust_bridge/feature/multiple_files.html):
+- for iOS, just add the 1st generated block-header files in `Runner-Bridging-Header.h`.
+- for MacOS, just add the 1st generated block-header files as **Objective-C Bridging Header**.
```
fzyzcjy, line 59:
```suggestion
# For multi-blocks

If there are [multi-blocks](https://cjycode.com/flutter_rust_bridge/feature/multiple_files.html):

- For iOS, just add the 1st generated block-header files in `Runner-Bridging-Header.h`.
- For MacOS, just add the 1st generated block-header files as **Objective-C Bridging Header**.
```

frb_codegen/src/generator/c/mod.rs
fzyzcjy: Weirdly the latest code is still config + configs (not configs + index). Maybe somehow it is accidentally reverted?

frb_codegen/src/lib.rs
```diff
@@ -40,10 +40,22 @@ mod parser;
...
 mod transformer;
-mod utils;
+pub mod utils;
 use error::*;
 
 pub fn frb_codegen(config: &config::Opts, all_symbols: &[String]) -> anyhow::Result<()> {
+    frb_codegen_multi(config, &[config.clone()], 0, all_symbols)
+}
+
+/// the `all_configs` here is used only for multi-blocks
```
fzyzcjy, line 50:
```suggestion
/// the `all_configs` here is used only for multi-blocks, because the primary block needs information from all other blocks
```

frb_codegen/src/lib.rs
```diff
@@ -40,10 +40,22 @@ mod parser;
...
+pub fn frb_codegen_multi(
+    config: &config::Opts,
+    all_configs: &[config::Opts],
+    index: usize,
```
fzyzcjy, line 54: uber-nit: Given that config == all_configs[index], what about removing `config` parameter

(to be honest it has been some time after I first reviewed this PR, so my memory is not that fresh. hopefully this uber-nit is correct!)

frb_codegen/src/lib.rs
```diff
@@ -40,10 +40,22 @@ mod parser;
...
+    all_symbols: &[String],
+) -> anyhow::Result<()> {
+    assert_eq!(config.block_index, BlockIndex(index));
```
fzyzcjy, line 57: If using the new API mentioned above, we can instead assert like:

```
assert!(configs.iter().enumerate().all(|index,config| config.index == BlockIndex(index))
```

then we are sure configs' indices are 0,1,2,...

frb_codegen/src/main.rs
```diff
@@ -16,8 +16,8 @@ fn main() -> anyhow::Result<()> {
 
     // generation of rust api for ffi
     let all_symbols = get_symbols_if_no_duplicates(&configs)?;
-    for config in configs.iter() {
-        if let Err(err) = frb_codegen(config, &all_symbols) {
+    for (i, config) in configs.iter().enumerate() {
+        if let Err(err) = frb_codegen_multi(config, &configs, i, &all_symbols) {
```
fzyzcjy, line 20:
```suggestion
    for (config_index, config) in configs.iter().enumerate() {
        if let Err(err) = frb_codegen_multi(config, &configs, config_index, &all_symbols) {
```

frb_example/pure_dart/rust/build.rs
```diff
@@ -21,16 +21,21 @@ fn main() {
...
+    let all_configs = config_parse(raw_opts);
 
-    // generation of rust api for ffi
-    let all_symbols = get_symbols_if_no_duplicates(&configs).unwrap();
-    for config in configs.iter() {
-        frb_codegen(config, &all_symbols).unwrap();
-    }
+    // generation of rust api for ffi (single block)
+    let all_symbols = get_symbols_if_no_duplicates(&all_configs).unwrap();
+    frb_codegen(&all_configs[0], &all_symbols).unwrap();
```
fzyzcjy, line 40: would be great to assert there are only 1 elements in vec

frb_example/pure_dart_multi/rust/build.rs
```diff
@@ -56,9 +56,9 @@ fn main() {
     // get opts from raw opts
     let configs = config_parse(raw_opts);
 
-    // generation of rust api for ffi
+    // generation of rust api for ffi (multi-blocks)
     let all_symbols = get_symbols_if_no_duplicates(&configs).unwrap();
-    for config in configs.iter() {
-        frb_codegen(config, &all_symbols).unwrap();
+    for (i, config) in configs.iter().enumerate() {
+        frb_codegen_multi(config, &configs, i, &all_symbols).unwrap()
```
fzyzcjy, line 62:
```suggestion
    for (config_index, config) in configs.iter().enumerate() {
        frb_codegen_multi(config, &configs, config_index, &all_symbols).unwrap()
```

fzyzcjy (commented):
Good job! Ready to merge after a few nits about minor programming style (for most of them, just click "commit suggestion" on github)

frb_example/pure_dart/rust/build.rs
dbsxdbsx: Refined in https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/1217983475c2a047aa56593d4bb47da14ee354c1.

frb_codegen/src/lib.rs
dbsxdbsx: Refined in https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/1217983475c2a047aa56593d4bb47da14ee354c1.

frb_codegen/src/lib.rs
dbsxdbsx: Refined in https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/1217983475c2a047aa56593d4bb47da14ee354c1.

dbsxdbsx:
@fzyzcjy, [for this](https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024#discussion_r1118068647), since there is already a `config` in `frb_codegen_multi`, there is no need to use `index` again. But to make the parameters less confusing, the `configs` is changed to `all_configs` in  https://github.com/fzyzcjy/flutter_rust_bridge/pull/1024/commits/1217983475c2a047aa56593d4bb47da14ee354c1.

frb_codegen/src/generator/c/mod.rs
dbsxdbsx: > Weirdly the latest code is still config + configs (not configs + index). Maybe somehow it is accidentally reverted?
Since there is already a `config` in frb_codegen_multi, there is no need to use the `index` again. But to make the parameters less confusing, the configs is changed to all_configs in https://github.com/fzyzcjy/flutter_rust_bridge/commit/1217983475c2a047aa56593d4bb47da14ee354c1.

frb_codegen/src/main.rs
```diff
@@ -16,8 +16,8 @@ fn main() -> anyhow::Result<()> {
 
     // generation of rust api for ffi
     let all_symbols = get_symbols_if_no_duplicates(&configs)?;
-    for config in configs.iter() {
-        if let Err(err) = frb_codegen(config, &all_symbols) {
+    for (config_index, _) in configs.iter().enumerate() {
```
fzyzcjy, line 19:
```suggestion
    for config_index in 0..configs.len() {
```

same for others

frb_example/pure_dart_multi/rust/build.rs
```diff
@@ -56,9 +56,9 @@ fn main() {
     // get opts from raw opts
     let configs = config_parse(raw_opts);
 
-    // generation of rust api for ffi
+    // generation of rust api for ffi (multi-blocks)
     let all_symbols = get_symbols_if_no_duplicates(&configs).unwrap();
-    for config in configs.iter() {
-        frb_codegen(config, &all_symbols).unwrap();
+    for (config_index, _) in configs.iter().enumerate() {
```
fzyzcjy, line 61: same here

fzyzcjy (approved):
Ready to merge!

frb_example/pure_dart_multi/rust/build.rs
dbsxdbsx: Refined in dcfe7f3

fzyzcjy:
Good job!

frb_codegen/src/lib.rs
```diff
@@ -40,10 +40,26 @@ mod parser;
 mod source_graph;
 mod target;
 mod transformer;
-mod utils;
+pub mod utils;
 use error::*;
 
 pub fn frb_codegen(config: &config::Opts, all_symbols: &[String]) -> anyhow::Result<()> {
+    frb_codegen_multi(&[config.clone()], 0, all_symbols)
```
Alex-A4, line 47: How this should work if `index` is always 0? 
How should we use it?

fzyzcjy:
frb_codegen is for single-block
frb_codegen_multi for multi block

@dbsxdbsx - maybe we should update the doc to tell users how to use it? I guess we should add a few `/// ...` comments to frb_codegen and frb_codegen_multi, and briefly mention that users should read corresponding doc in our mini book

fzyzcjy:
Hmm there was a reply here asking how to use it, but later it seems to be deleted by the author?

Alex-A4:
> Hmm there was a reply here asking how to use it, but later it seems to be deleted by the author?

Yeah, I found in files

Alex-A4:
Hmm, it looks like here's a strange behaviour.
Our code root: https://github.com/broxus/nekoton_bridge/tree/attempt_generate_from_multiple_files/packages/nekoton_bridge
Look at files:
[crypto/mnemonic/mnemonic_api.dart](https://github.com/broxus/nekoton_bridge/blob/attempt_generate_from_multiple_files/packages/nekoton_bridge/lib/src/nekoton/crypto/encrypted_key/encrypted_key_api.dart)
[crypto/encrypted_key/encrypted_key_api.dart](https://github.com/broxus/nekoton_bridge/blob/attempt_generate_from_multiple_files/packages/nekoton_bridge/lib/src/nekoton/crypto/encrypted_key/encrypted_key_api.dart)

Both files have `class MnemonicTypeG with _$MnemonicTypeG` but on rust side both files look into [native/src/nekoton/models_api.rs](https://github.com/broxus/nekoton_bridge/blob/attempt_generate_from_multiple_files/packages/nekoton_bridge/native/src/nekoton/models_api.rs) which in dart side doesn't contains such class.

And I still get duplicate error:
```
conflicting implementations of trait `flutter_rust_bridge::IntoDartExceptPrimitive` for type `nekoton::models_api::MnemonicTypeG`
   --> packages/nekoton_bridge/native/src/encrypted_key_generated.rs:81:1
    |
81  | impl support::IntoDartExceptPrimitive for MnemonicTypeG {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `nekoton::models_api::MnemonicTypeG`
    |
   ::: packages/nekoton_bridge/native/src/mnemonic_generated.rs:124:1
    |
124 | impl support::IntoDartExceptPrimitive for MnemonicTypeG {}
    | ------------------------------------------------------- first implementation here
```

dbsxdbsx:
> frb_codegen is for single-block frb_codegen_multi for multi block
> 
> @dbsxdbsx - maybe we should update the doc to tell users how to use it? I guess we should add a few `/// ...` comments to frb_codegen and frb_codegen_multi, and briefly mention that users should read corresponding doc in our mini book

I think the doc with  `/// ...` is enough to explain `frb_codegen_multi ` and `frb_codegen`. I guess the signatures of the 2 functions are easy to read, right? 

What about doing a little modification for the function docs like this:

```rust
/// when the API is only defined in 1 rust file(block), take this one for generation, where `config`
/// is the instance containing all information to the API file(block), and `all_symbols` contains
/// all unique APIs defined in the file mentioned above.
/// If APIs are defined in more than 1 rust file(block), use `frb_codegen_multi` instead.
pub fn frb_codegen(config: &config::Opts, all_symbols: &[String]) -> anyhow::Result<()> {
...
}

/// the function is used only for cases with multi-blocks when there are
/// more than 1 API block. And because the current block to deal with needs information
/// from all other blocks， so `all_configs` is used here,
/// with `index` refers to the place of the current block to deal with.
/// For details on how to take advantage of multi-blocks, please refers to this article: https://cjycode.com/flutter_rust_bridge/feature/multiple_files.html
pub fn frb_codegen_multi(
    all_configs: &[config::Opts],
    index: usize,
    all_symbols: &[String],
) -> anyhow::Result<()> { 
...
}
```

dbsxdbsx:
>Both files have class MnemonicTypeG with _$MnemonicTypeG but on rust side both files look into [native/src/nekoton/models_api.rs](https://github.com/broxus/nekoton_bridge/blob/attempt_generate_from_multiple_files/packages/nekoton_bridge/native/src/nekoton/models_api.rs) which in dart side doesn't contains such class.

@Alex-A4, for your issue:
1. I wonder if this issue only occurs after updating to v1.65.1?(I notice that you used v1.64.0 before) If this is the case, it is weird, because I only update the generated c-header files, which should no effect other parts of frb.
2. Though I'm not familiar with the conflict issue with `trait`(I didn't implement generation for feature trait),  it seems that it is some conflict like [this](https://github.com/fzyzcjy/flutter_rust_bridge/issues/987). Maybe it is a good time to review your issue again after[ the shared-part has been totally decoupled](https://github.com/fzyzcjy/flutter_rust_bridge/issues/987#issuecomment-1407940083).

fzyzcjy:
> I think the doc with /// ... is enough to explain frb_codegen_multi  and frb_codegen. 

I think so

PR #1026, opened 2023-02-05
Provide 'features' to optimize the number of pool workers

temeddix:
Until now, it was impossible to change the number of threads executed in the threadpool(or workerpool in WASM mode). The number of 4 was fixed. In real-world development, there are cases where you need to change the thread count.

This PR removes this limitation by providing 'features'. If you think that this number of 4 is inappropriate for your project, you can choose a different option. By specifying features in `Cargo.toml`, you can optimize the number of threads or workers in the pool. Just include `features` key when adding flutter_rust_bridge as a dependency.

```toml
// Cargo.toml

[dependencies]
flutter_rust_bridge = { workspace = true, features = ["worker-max"] }
```

Currently available options related to worker pool are:

- `worker-single`: Uses 1 worker in the pool.
- `worker-max`: Uses all available logical cores.

## Checklist

- [x] CI is passing.

## Remark for PR creator


temeddix:
But I'm not sure what `The code generator is run and the code is formatted (e.g. via just refresh_all).` means. There is no `pubspec.yaml` file at the project root. How should I run the code generator?

temeddix:
I will be happy to work on any modifications if I get any response. I am also okay with others making changes in this PR.

fzyzcjy:
Great job! I will review it after CI gets green :)

frb_rust/src/macros.rs
```diff
@@ -72,30 +72,35 @@ macro_rules! transfer {
...
-/// flutter_rust_bridge::spawn!(|port2: MessagePort| {
+/// flutter_rust_bridge::spawn!(4, |port2: MessagePort| {
 ///     port2.post_message(&JsValue::from(true)).unwrap();
 /// });
 /// # } #[cfg(not(target_family = "wasm"))] fn main() {}
 /// ```
 #[macro_export]
 macro_rules! spawn {
-    ($($tt:tt)*) => {{
+    ($wc:expr, $($tt:tt)*) => {{
```
fzyzcjy, line 82: uber-nit: would be great to avoid abbreviation `wc` (I guess it is worker_count?)

frb_rust/src/macros.rs
```diff
@@ -72,30 +72,35 @@ macro_rules! transfer {
...
         let worker = $crate::transfer!($($tt)*);
         #[cfg(not(target_family = "wasm"))]
         {
-            $crate::thread::THREAD_POOL.lock().execute(worker)
+            use $crate::thread;
+            thread::create_thread_pool($wc);
```
fzyzcjy, line 87: The problem of the implementation of the PR is that, it seems the `worker_count` only works for the *first* call of `spawn!`, since `create_thread_pool` only creates the pool in the first call. This will make users very confused IMHO :(

fzyzcjy:
Also cc @Desdaemon who did some of the thread pool (e.g. the web part) :)

temeddix:
I'm trying to fix bugs on WASM worker pool, but since I am not used to WASM I have trouble fixing it. Also, I am not sure how I can change configuraition to wasm to enable `#[cfg(wasm)]` macro. I'll try to handle CI problems as much as possible.

frb_rust/src/macros.rs
temeddix: Yes, it means exactly worker_count. I will fix using abbreviation :)

fzyzcjy:
For wasm cc @Desdaemon 

Well I realize the semantics have a little problem (see my comments above). Thus, no need to worry about the CI, before discussing and fixing semantics :)

frb_rust/src/macros.rs
temeddix: Oh, I thought that calling spawn! was reposible for creating the pool itself. I now see that it is for passing functions into the existing thread pool. Thanks for telling me:)

Desdaemon:
> Also, I am not sure how I can change configuraition to wasm to enable #[cfg(wasm)] macro.

It was only defined within frb_rust (see build.rs), so outside of it you'll need to use the standard `#[cfg(target_family = "wasm")]`.

temeddix:
I changed the way things work. When you use `FLUTTER_RUST_BRIDGE_HANDLER` in `api.rs`, flutter_rust_bridge doesn't create the pool for you. In this case, you manually create your own.

temeddix:
I know there are so many questions, but can this be solved on PR's side?

image

fzyzcjy:
That windows error looks like an infra bug. You may need to file a bug at that github action https://github.com/actions-rs/toolchain

frb_rust/src/thread.rs
```diff
@@ -1,34 +1,56 @@
...
+            None => {
+                *mutex_guard = Some(ThreadPool::with_name(
+                    "frb_workerpool".into(),
+                    workers_count as usize,
+                ));
+            }
+            Some(_) => {
+                error!("Thread pool already exists (non-WASM)");
+            }
+        };
```
fzyzcjy, line 22: Looks much better now!

Is it possible that, we somehow eliminate the whole *global variable* (THREAD_POOL_HOLDER), and put it as a field of the Handler struct? By doing so, the code style and maintainability will be higher.

temeddix: Thanks! I will be working on that In a day or two.

fzyzcjy: You are welcome and take your time!

temeddix: IMHO it seemed best to have the threadpool or the workerpool as a field of `ThreadPoolExecutor`. I'm not sure how I should make `spawn!` macros in that case, but I'll give it a try.

temeddix: I also wanted to ask if `Mutex<ThreadPool>` or `thread_local!{WorkerPool}` is necessary. I thought that if the Rust function(async) is called from the main Dart thread, we don't need to use `Mutex` or `thread_local!` because the caller is always on the main thread. Is this approach correct?

fzyzcjy: > I'm not sure how I should make spawn! macros work in that case, but I'll give it a try.

cc @Desdaemon - what do you think? IIRC the `spawn!` macro is introduced by your PR

> I also wanted to ask if Mutex<ThreadPool> or thread_local!{WorkerPool} is necessary

Feel free to try to remove Mutex etc! If Rust compiler is happy and there is no unsafe code, we are safe to go :)

Desdaemon: Since `spawn!` depends on a thread pool visible from global scope, it does make it harder to call it without a handle on the thread pool. What we could do is assume that `FLUTTER_RUST_BRIDGE_HANDLER` is in scope (i.e. defined by user or imported from bridge file) as well as being the right type, so that compile errors are emitted when the thread pool cannot be accessed from the handler. This is a breaking change, and should be communicated as such to the consumers.

fzyzcjy: @Desdaemon Or, maybe we should add an argument to `spawn!`, i.e. explicitly pass that handler/thread-pool to it. Because global variables are known to be against the best programming styles.

temeddix: Idea of passing threadpool as an argument looks cleaner, 👌. I'm not working on it yet so design changes can take place.

Desdaemon: That's fine, the only reason I didn't mention passing the thread pool is it would end up looking like `spawn!(HANDLER.thread_pool, || { .. })` which seems counter-intuitive and bit too tightly coupled IMO.

Least to say, this is a shared problem when the larger ecosystem of `std::thread::spawn` not being replaceable.

temeddix: Given that macros are just converted into plain code at compile time instead of working like a function, maybe I should just find a way to access global `FLUTTER_RUST_BRIDGE_HANDLER` in `api.rs` or `bridge_generated.rs` within `spawn!`

temeddix: After diving in to the code generation mechanism, I kind of became skeptical about this idea of `threadpool per handler`.

- If `ThreadPool` becomes a field somewhere inside `SimpleHandler`, in theory user can create multiple threadpools if they make multiple handlers. In this situation, `spawn!` cannot determine which threadpool it should use(unless passing threadpool as an argument, which is not intuitive).
- If `ThreadPool` remains to be a global variable in `thread.rs` in `frb_rust` crate folder, we can assure that there is only and only one threadpool dedicated for flutter_rust_bridge function calls.

For example, if the user has multiple custom library crates each of which having `FLUTTER_RUST_BRIDGE_HANDLER` in `bridge_generated.rs`, there would be multiple threadpools because there are multiple handlers. This can be excessive, since threadpool is already a collection of multiple threads.

I am also skeptical about my approach because if the user has multiple custom library crates, it is unclear in which library should the user use the function `create_pool(N)`.

IMHO current system is brilliant and keeping it would be better. Although global variable is not the best programming style, it can still serve as a single source of data.

I just hope this 'system' can be customized, such as changing the thread count. There could be a number of ways...
- Defining FRB's own configuration file(not mandatory), located in user's project root or `./native`, where the user can customize the number of threads in the pool, log threshold level, etc.
- Environment variables.
- Reading `build.rustflags` object in `.cargo/config.toml` file.

fzyzcjy: Hmm that's a problem.

Alternatively, we may refer to how other crates solve the problem. For example, `rayon` is a parallel computing library and has a global thread pool - similar to our case! It solves it by: https://github.com/rayon-rs/rayon/blob/master/FAQ.md#how-many-threads-will-rayon-spawn

> If you want to alter the number of threads spawned, you can set the environmental variable RAYON_NUM_THREADS to the desired number of threads or use the [ThreadPoolBuilder::build_global function](https://docs.rs/rayon/*/rayon/struct.ThreadPoolBuilder.html#method.build_global) method.

where the [build_global](https://docs.rs/rayon/1.6.1/rayon/struct.ThreadPoolBuilder.html#method.build_global) mentions it is allowed to be called exactly once

temeddix:
I thought that for now, there is not a big demand for specifying the exact number of workers. Therefore I thought this 'features' will be helpful. Now the user can use 'features' to optimize the pool :)

temeddix:
I only made some slight modifications in the code and I'm confident that I didn't break or violate formatting rules. However, CI is not passing due to various infra bugs... Please fix these github workflows

frb_rust/src/thread.rs
```diff
@@ -1,4 +1,32 @@
-pub const WORKERS_COUNT: usize = 4;
+#[cfg(not(wasm))]
+fn get_thread_count() -> usize {
+    #[allow(unused_mut, unused_assignments)]
+    let mut thread_count: usize = 4;
+    #[cfg(feature = "thread-count-single")]
+    {
+        worker_count = 1;
+    }
+    #[cfg(feature = "thread-count-max")]
```
fzyzcjy, line 9: uber-nit: would be great to emit compile-time error if user wrongly specifies both features

frb_rust/src/thread.rs
```diff
@@ -1,4 +1,32 @@
...
+    #[cfg(feature = "worker-count-single")]
+    {
+        worker_count = 1;
+    }
+    #[cfg(feature = "worker-count-max")]
+    {
+        worker_count = 16;
+    }
+    worker_count
+}
```
fzyzcjy, line 29: uber-nit: looks like we can merge the two into one function, and only use `cfg(wasm)` for a few lines, since they only have minor differences

fzyzcjy:
The CI seem to be having formatting problems. Maybe `just refresh_all`?

frb_rust/src/thread.rs
temeddix: Okay :) That wouldn't be hard

frb_rust/src/thread.rs
temeddix: Good point, I'll be working on that in a day

fzyzcjy: Take your time!

frb_rust/src/thread.rs
```diff
@@ -1,4 +1,32 @@
...
+fn get_worker_count() -> usize {
+    #[allow(unused_mut, unused_assignments)]
+    let mut worker_count: usize = 4;
+    #[cfg(feature = "worker-count-single")]
+    {
+        worker_count = 1;
+    }
+    #[cfg(feature = "worker-count-max")]
+    {
+        worker_count = 16;
```
fzyzcjy, line 26: btw, why 16 instead of something like "number of cores"? (I guess because there is no such API in the web?)

However, I guess few users have 16 cpu cores on their computers. Thus is it possible we find an API similar to std::thread::available_parallelism

temeddix: I just found out that workerpool relies on web workers, which actually ARE true multithreading. I should change the code. Also, web workers make use of logical cores, using the same `std::thread::available_parallelism` would be valid on WASM worker pool I think.

Reference about web workers: https://stackoverflow.com/questions/11871452/can-web-workers-utilize-100-of-a-multi-core-cpu

temeddix:
> The CI seem to be having formatting problems. Maybe `just refresh_all`?

I don't think I understood well. Is that a git console command? or a button somewhere in GitHub?

fzyzcjy:
Run `just refresh_all` on your local folder. It is a `justfile` and executes some scripts

book/src/feature/worker_pool.md
```diff
@@ -0,0 +1,21 @@
+# Worker pool
```
fzyzcjy, line 1: TODO(fzyzcjy): Refine doc after PR merges

(you can ignore this comment, just for reminder of myself)

temeddix:
Apart from the PR itself, passing the CI seems to be too hard. There are many phenomenons that are out of my scope.

1. I didn't make any change to Dart files. Howver CI was telling me that `CI / Dart linter using analyze and format (pull_request)` was not passing. Therefore I had to run `just refresh_all`, which made a huge amount of modifications to Dart files.2.    
    image
1. `CI / Flutter (Windows) integration test / flutter_windows_test (pull_request)` is occasionally failing. I understood that there is another Rust toolchain available, but I don't understand if that means I'm allowed to change workflow files in `.github` folder. Even if I'm allowed, it's not a point of this PR.
    image
1. And even though the CI is not passing.

In short, I don't know why I should be modifying these unrelated files. I only made some slight modifications to Rust files, but it was not passing the CI. Because of that I ran just commands to modify so many unrelevant files.

fzyzcjy:
Yes I agree the current CI is weird, possibly because Flutter has recently released a new version? Then maybe we should create another PR to firstly fix the CI.

temeddix:
For now, I reverted back all the unnecessary changes related to Dart files which were made by using `just` command. I only left the core changes for this PR, please let me know if there's something needed to be done on PR's side.

fzyzcjy:
Sure! I will merge as soon as CI is green - since you know, anything with red CI should not be fixed. Thus, feel free to fix the CI (probably in another PR)!

temeddix:
> Sure! I will merge as soon as CI is green - since you know, anything with red CI should not be fixed. Thus, feel free to fix the CI (probably in another PR)!

@fzyzcjy Now the CI is green :)

fzyzcjy:
🎉

Issue #941, opened 2022-12-30
[Bug] just gen generated using old flutter_rust_bridge version


kingfisherphuoc:
### Describe the bug

I used to use `flutter_rust_bridge` version 1.49.0. Now I want to update it to the version `flutter_rust_bridge: ^1.59.0`. After upgrading with clean && pub get again, it always used version 1.49.0.
Here my step:
1. clone the template repo
2. it could build an run normally
3. `pub clean && pub get` and `just clean && just gen`
4. it gave me those files:
```
// AUTO GENERATED FILE, DO NOT EDIT.
// Generated by `flutter_rust_bridge`@ 1.49.0.
```
<img width="1474" alt="Screenshot 2022-12-30 at 12 14 31" src="https://user-images.githubusercontent.com/962484/210037046-f0a5f573-7ae7-413c-b0b3-f96bd0954c17.png">

Any suggestion?

### Codegen logs with `RUST_LOG=debug` environment variable

```shell
no issue with code gen
```

fzyzcjy:
try to update versions in cargo.toml

kingfisherphuoc:
I did try to update cargo as well, but it's no use. 
<img width="1349" alt="Screenshot 2022-12-30 at 13 44 44" src="https://user-images.githubusercontent.com/962484/210042340-93af4b52-65f3-49b4-b920-31a354cd1bd2.png">
<img width="1370" alt="Screenshot 2022-12-30 at 13 44 32" src="https://user-images.githubusercontent.com/962484/210042329-a7dae938-a019-481d-8efe-e807217d22de.png">

fzyzcjy:
maybe your (globally installed) flutter_rust_bridge_codegen binary is not updated?

kingfisherphuoc:
thanks. It did the trick. However, I got another error when I run `just gen` command:
```
(base) kingfisher-phuoc@Doans-MacBook-Pro flutter_rust_bridge_template % just gen
flutter pub get
Running "flutter pub get" in flutter_rust_bridge_template...      1,360ms
flutter_rust_bridge_codegen --rust-input native/src/api.rs --dart-output lib/bridge_generated.dart --c-output ios/Runner/bridge_generated.h --c-output macos/Runner/bridge_generated.h --dart-decl-output lib/bridge_definitions.dart
error: --c-output's inputs should match --rust-input's length
```
Do you have any idea?

fzyzcjy:
[...]

PR #1050, opened 2023-02-10
docs: fix broken library initialization script

GregoryConrad:
There were a few issues with the old library initialization script:

1. Generated C output could only go to one file, but multiple were specified. This was broken sometime after the docs were made.
2. The Dart and Rust FRB versions might mismatch. This commit pins an exact version of FRB in Dart/Rust so no issues will occur.

### DIFF
--- a/book/src/library/dart_only.md
+++ b/book/src/library/dart_only.md
@@ -13,6 +13,12 @@ After the script runs, change the `flutter_rust_bridge` dependency in `/packages
 ```yaml
   flutter_rust_bridge: "1.62.1"
 ```
+Note: If you so choose, you can update the `flutter_rust_bridge` versions in
+`/packages/library_name/native/Cargo.toml` and `/packages/library_name/pubspec.yaml`
+to the latest version available, but newer versions are not guaranteed to work
+with this section of the guide due to a lack of CI testing.
+Version 1.62.1 is known to work with this guide as-is.
+CI testing is planned once the Native Assets feature is released.
 
 Finally, change the variables at the top of the script to fit your needs.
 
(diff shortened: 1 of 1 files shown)


book/src/library/dart_only.md
```diff
@@ -9,6 +9,11 @@ The following script assumes a bash shell, which you should make sure to use to
 Also, the script generates some ffi convenience files in your Dart `src/` which you should
 check out after the script completes.
 
+After the script runs, change the `flutter_rust_bridge` dependency in `/packages/library_name/pubspec.yaml` to the following:
+```yaml
+  flutter_rust_bridge: "1.62.1"
```
fzyzcjy, line 14: uber-nit: what about telling the reader that "please replace the version to whatever the latest version is"

fzyzcjy (approved):
LGTM :)

book/src/library/dart_only.md
GregoryConrad: I've had quite a few breaking changes in the last 10 or so FRB versions in my own package, so I am a bit hesitant to recommend users to use the latest version of FRB in the docs directly (this wouldn't be a problem with proper CI testing, but as I mentioned in #941, that might want to wait). I know that 1.62.1 works properly, which is why I put that explicitly in the docs.

I think if users want to update the FRB version after they finish the docs, that would be OK, but it could break their project. I just didn't want to put it in that part of the docs since users might wonder why the package the docs helped them create doesn't work.

fzyzcjy: Looks reasonable! So maybe add a line saying "1.62.1 is tested to work; you may use newer versions but this part does not have proper testing yet so may breaking change"

GregoryConrad: Should be fixed in [406ac38](https://github.com/fzyzcjy/flutter_rust_bridge/pull/1050/commits/406ac38541f7a41a50e0365ac7f5c692779980aa)

fzyzcjy:
Well done :)

PR #1056, opened 2023-02-11
Fix the problem with using `worker-max` feature in WASM

temeddix:
## Changes

This PR fixes the problem with using `worker-max` feature from #1026 in WASM configuration. Now JavaScript object `navigator.hardwareConcurrency` is used for getting the number of logical in WASM config because `std::thread` module is not available in WASM. Now this feature would be much more stable than before.

The reason #1026 could pass the CI was that `worker-max` testing was not done in WASM mode. Therefore, I made `pure_dart` example(with WASM CI test) use this feature instead of `with_flutter` example.

Also, this adds `.vscode` folder to .`.gitignore`. People using VSCode need this because configuring VSCode Rust-Analyzer to check in WASM mode creates `.vscode` folder. `.vscode` folder is also used for many other custom settings of VSCode.

## Checklist

- [x] CI is passing.

## Remark for PR creator

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -289,8 +289,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - features: [ ]
-          - features: [ wasm-start ]
+          - features: []
+          - features: [wasm-start]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


fzyzcjy:
Good fix! Will merge after CI green

temeddix:
@fzyzcjy  Now the problems are all solved. I checked WASM config with `worker-max` working on my PC too.

PR #1079, opened 2023-02-22
Fix `Optional<DateTime>`

alexthe2:
## Changes

Close #1077 

## Checklist

- [x] An issue to be fixed by this PR is listed above.
- [x] New tests are added to ensure new features are working. End-to-end tests are usually in the `./frb_example/pure_dart` example, more specifically, `rust/src/api.rs` and `dart/lib/main.dart`.
- [x] The code generator is run and the code is formatted (via `just precommit`).
- [x] If this PR adds/changes features, documentations (in the `./book` folder) are updated.
- [x] CI is passing.

## Remark for PR creator

- Justfile is a task runner for the command line interface that allows you to run shell commands from a file named `justfile` in your project directory. You can use Justfile after [installing it](https://github.com/casey/just). Note that commands written in `justfile` of this repository are expected to be run in `bash`, not `cmd` or `powershell`. Running `just ...` commands in `cmd` or `powershell` will produce errors as the syntax is not compatible. On Windows, you can use `git bash` if you have Git installed.
- If fzyzcjy does not reply for a few days, maybe he just did not see it, so please ping him.

PR #1084, opened 2023-02-23
Oragnize README layout

temeddix:
## Changes

This fixes two main issues of the current README layout.

Currently, there is a line break on the package description. Now the description is separated to improve readability.
image
image

Also, right now the contributor table's column width is inconsistent and verbose. With this PR it's neatly organized as equally-sized circles. Clicking on the badge will redirect to `Insights` GitHub tab, where contriubtions of each person is shown as a detailed graph.
image
image

You can see how it looks on the forked repo: https://github.com/temeddix/flutter_rust_bridge

IMHO I really do think that improving the first layout of the project will provide a sense of robustness and stability to new people. First impressions are important, raising the possibility of new users getting involved.

## Checklist

- [x] CI is passing.

## Remark for PR creator


README.md
```diff
@@ -1,26 +1,32 @@
...
-* **Cross-platform**: Android, iOS, Windows, Linux, MacOS, and Web.
-* **Easy to code-review & convince yourself**: This package simply simulates how humans  write boilerplate code. If you want to convince yourself (or your team) that it is safe, there is not much code to look at. No magic at all! ([More about](https://fzyzcjy.github.io/flutter_rust_bridge/safety.html) safety concerns.)
-* **Fast**: It is only a thin (though feature-rich) wrapper, without overhead such as protobuf serialization, thus performant. (More [benchmarks](https://github.com/fzyzcjy/flutter_rust_bridge/issues/318#issuecomment-1034536815) later) <small>(Throw away components like thread-pool to make it even faster)</small>
-* **Pure-Dart compatible:** Despite the name, this package is 100% compatible with [pure](https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_example/pure_dart/README.md) Dart.
+- **Memory-safe**: Never need to think about malloc/free.
```
fzyzcjy, line 22: uber-nit: maybe we can keep the unrelated formatting unchanged ;)

README.md
```diff
@@ -32,141 +38,63 @@ Here is my another open-source library :) https://github.com/fzyzcjy/flutter_smo
...
-<!-- markdownlint-restore -->
-<!-- prettier-ignore-end -->
-
-<!-- ALL-CONTRIBUTORS-LIST:END -->
+Thanks goes to these wonderful people!
+
+[![All Contributors](https://img.shields.io/github/all-contributors/fzyzcjy/flutter_rust_bridge?color=ee8449)](#contributors-)
+
+<a href="https://github.com/fzyzcjy/flutter_rust_bridge/graphs/contributors">
+  <img src="https://contrib.rocks/image?repo=fzyzcjy/flutter_rust_bridge" />
```
fzyzcjy, line 46: nit: So... how to show who makes contribution to *code* and who for *doc* etc? The original all-contributors does this by emojis.

README.md
```diff
@@ -1,26 +1,32 @@
...
+    <b><a href="https://github.com/fzyzcjy/flutter_rust_bridge">flutter_rust_bridge</a></b>
+</h1>
 
-[![Rust Package](https://img.shields.io/crates/v/flutter_rust_bridge.svg)](https://crates.io/crates/flutter_rust_bridge)
-[![Flutter Package](https://img.shields.io/pub/v/flutter_rust_bridge.svg)](https://pub.dev/packages/flutter_rust_bridge)
-[![Stars](https://img.shields.io/github/stars/fzyzcjy/flutter_rust_bridge)](https://github.com/fzyzcjy/flutter_rust_bridge)
-[![CI](https://github.com/fzyzcjy/flutter_rust_bridge/actions/workflows/ci.yaml/badge.svg)](https://github.com/fzyzcjy/flutter_rust_bridge/actions/workflows/ci.yaml)
-[![Example](https://github.com/fzyzcjy/flutter_rust_bridge/actions/workflows/post_release.yaml/badge.svg)](https://github.com/fzyzcjy/flutter_rust_bridge/actions/workflows/post_release.yaml)
-[![Codacy Badge](https://api.codacy.com/project/badge/Grade/6afbdad19e7245adbf9e9771777be3d7)](https://app.codacy.com/gh/fzyzcjy/flutter_rust_bridge?utm_source=github.com&utm_medium=referral&utm_content=fzyzcjy/flutter_rust_bridge&utm_campaign=Badge_Grade_Settings)
+<h3 align="center">High-level memory-safe binding generator for Flutter/Dart <-> Rust</h3>
```
fzyzcjy, line 5: Hmm I wonder does this work for `pub.dev` and `crates.io`? because the README will be sync into those two as well

README.md
fzyzcjy, line 46: Another issue is that the ordering cannot be customized, thus we lack some flexibilities IMHO. For example, we know @Desdaemon has done a lot of great job in this repo, so seems that we should put him to the second place. (I have not order all other contributors yet, but the all-contributors approach will have this flexibility)

fzyzcjy (commented):
Looks interesting! May need to have several more checks (see comments). Given limited time, we may firstly focus on more critical PRs and issues :)

README.md
temeddix: Oh, this was from VSCode's auto formatting by [Prettier](https://prettier.io/). What do you think about using the new format? If not, I do can revert it by hand.

README.md
temeddix: With this approach, the order of the badges follows the number of commits. The reason @Desdaemon is not in the second is because there were other contributors using small-sized, large number of commits in the PR. If the commits were squashed per PR, @Desdaemon actually would be the second I believe.

IMO the order would not that important, because the first purpose of README is mainly introducing the package, rather than all the contributors. GitHub's `Insights` tab is exactly for that :)

image

temeddix: I think those icons can be placed below at the list:
image

If you agree with this idea, I will go ahead and work on it.

README.md
temeddix: The `h1` and `h3` tag should work properly in Markdown files for both pub.dev and crates.io. When we publish the package, the README file will be rendered as HTML, which will just leave all HTML tags in place. Even complex tags like `table` were working well.

image
image

README.md
temeddix: I appreciate your giving credits to main contributors. So, this would be a matter of choice:
- between having control over the order of contributions table even though it is misaligned
- Use orgnized contribution table, using the `Insignts` tab for detailed contribution view

fzyzcjy:
I will come back with this later when having time. By the way, seems the emojis in all-contributor spec (e.g. "code", "doc") that this PR is going to remove is also important: For example, https://github.com/fzyzcjy/flutter_rust_bridge/pull/1079#issuecomment-1441712774. I strongly believe that we should clearly tell every user who made what contribution!

PR #1099, opened 2023-03-03
added doc on unit testing and stack traces

alexthe2:
## Changes

Added a doc section on how to unit test from flutter with the bridge

## Checklist
- [x] CI is passing.

## Remark for PR creator

### DIFF
--- a/book/src/feature/logging.md
+++ b/book/src/feature/logging.md
@@ -1,7 +1,7 @@
 # Stack Traces
 To pass Rust stack traces to flutter, you need to set `RUST_BACKTRACE` in the running application. For that simply add `env::set_var("RUST_BACKTRACE", "1");` before initialising the bridge.
 
-Note: `--dart-define` doesn't work, you **must** use `env::set_var`.
+Note: The `--dart-define` will not work, you **must** use `env::set_var`, because the former does not set the "environment variable" in the common sense, but instead a special thing only visible to Dart.
 
 # Logging
 
--- a/book/src/feature/logging.md
+++ b/book/src/feature/logging.md
@@ -1,8 +1,3 @@
-# Stack Traces
-To pass Rust stack traces to flutter, you need to set `RUST_BACKTRACE` in the running application. For that simply add `env::set_var("RUST_BACKTRACE", "1");` before initialising the bridge.
-
-Note: The `--dart-define` will not work, you **must** use `env::set_var`, because the former does not set the "environment variable" in the common sense, but instead a special thing only visible to Dart.
-
 # Logging
 
 Since I have seen some questions asking how logging can be implemented with a Flutter + Rust application, here are some examples.
--- a/book/src/feature/logging.md
+++ b/book/src/feature/logging.md
@@ -1,3 +1,8 @@
+# Stack Traces
+To pass Rust stack traces to flutter, you need to set `RUST_BACKTRACE` in the running application. For that simply add `env::set_var("RUST_BACKTRACE", "1");` before initialising the bridge.
+
+Note: `--dart-define` doesn't work, you **must** use `env::set_var`.
+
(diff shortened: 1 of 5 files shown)


fzyzcjy:
IIRC, we need to ensure that dylib (or whatever) is built for the host platform. For example, 

* to unit test on macos, need to build for macos (<--- not ios)
* to use it normally on a iphone, need to build for ios

alexthe2:
oh yeah, you’re right. Is `cargo build` automatically buildiing for macOS?

fzyzcjy:
IIRC yes

alexthe2:
like this?

alexthe2:
also added the stacktrace section, imo the best location is under logging, as it’s kind of logging, I can also move it into a seperate section

fzyzcjy:
Looks good :) I will come back and check later

book/src/feature/logging.md
```diff
@@ -1,3 +1,8 @@
+# Stack Traces
```
fzyzcjy, line 1: IMHO it would be great to:

1. move stack traces into separate file, say `feature/stack_trace.md`
2. Add a sentence in this file, saying "if you want to have stack trace, see that file" so users can link to there

book/src/feature/logging.md
```diff
@@ -1,3 +1,8 @@
...
+To pass Rust stack traces to flutter, you need to set `RUST_BACKTRACE` in the running application. For that simply add `env::set_var("RUST_BACKTRACE", "1");` before initialising the bridge.
+
+Note: `--dart-define` doesn't work, you **must** use `env::set_var`.
```
fzyzcjy, line 4:
```suggestion
Note: The `--dart-define` will not work, you **must** use `env::set_var`, because the former does not set the "environment variable" in the common sense, but instead a special thing only visible to Dart.
```

book/src/unit_tests_dart.md
```diff
@@ -0,0 +1,59 @@
+# Unit tests in dart
+
+To run `flutter` or `dart test` with the bridge you need to load the library on your own development machine (Windows/MacOs/Linux/CI). For that use `loadDylib`, for example:
```
fzyzcjy, line 3:
```suggestion
To run `flutter` or `dart test` with the bridge you need to load the library on your own development machine (Windows/MacOS/Linux/CI). For that use `loadDylib`, for example:

```

book/src/unit_tests_dart.md
```diff
@@ -0,0 +1,59 @@
...
+```dart
+BridgeImpl initializeExternalLibrary(String path) =>
+    BridgeImpl(
+      loadDylib(path),
+    );
```
fzyzcjy, line 8:
```suggestion
BridgeImpl initializeExternalLibrary(String path) => BridgeImpl(loadDylib(path));
```

fzyzcjy (approved):
Nice! Only a few nits and ready to merge

book/src/feature/logging.md
alexthe2: I was thinking about that, but the section is so tiny, that I thought it better fits into `Logging`, but maybe a section `Debugging` would make more sense.

book/src/unit_tests_dart.md
alexthe2: good catch, I can fix this in my code as well ;)

book/src/feature/logging.md
fzyzcjy: IMHO tiny is not a big issue. Instead, it is important for readers to quickly find them - otherwise it is equivalent to not write down it at all ;)

alexthe2: ok done

fzyzcjy:
Nice!

PR #1106, opened 2023-03-05
Add tests to gen, build, run pure_dart(_multi)

coder0xff:
## Changes

For #1105, this PR adds `#[test]`s to `frb_codegen/main.rs` to generate, build, and run pure_dart and pure_dart_multi.

It may be desirable to remove `frb_example/pure_dart/rust/build.rs` and `frb_example/pure_dart_multi/rust/build.rs` entirely, to prevent divergence, leverage Rust's test capabilities, and to avoid confusion caused by #1105 in the first place, but I have left them there to keep changes to a minimum.

I also experimented with making `pure_dart` and `pure_dart_multi` independent, not part of the workspace, but I ran into problems with Rust failing to resolve `IntoDart` implementations for Vecs of `chrono` types.

## Checklist

- [x] An issue to be fixed by this PR is listed above.
- [x] New tests are added to ensure new features are working. End-to-end tests are usually in the `./frb_example/pure_dart` example, more specifically, `rust/src/api.rs` and `dart/lib/main.dart`.
- [x] The code generator is run and the code is formatted (via `just precommit`).
- [x] If this PR adds/changes features, documentations (in the `./book` folder) are updated.
- [ ] CI is passing.

## Remark for PR creator

- Justfile is a task runner for the command line interface that allows you to run shell commands from a file named `justfile` in your project directory. You can use Justfile after [installing it](https://github.com/casey/just). Note that commands written in `justfile` of this repository are expected to be run in `bash`, not `cmd` or `powershell`. Running `just ...` commands in `cmd` or `powershell` will produce errors as the syntax is not compatible. On Windows, you can use `git bash` if you have Git installed.
- If fzyzcjy does not reply for a few days, maybe he just did not see it, so please ping him.

PR #1119, opened 2023-03-12
General CLI improvements: dumping, upgrade clap, yaml config, serialize IR, bump versions, etc

Desdaemon:
## Changes

- feat: `--dump` option to inspect config and IR
- feat: upgrade to clap v4, read config from .yml file
  - reads from `.frb.yml`, `.frb.yaml` and `.frb.json`
- fix: compile CLI without `chrono` feature
- feat: implement Serialize for IR structs
  - guarded behind `serde` feature
- fix(ci): bump MSRV to satisfy cargo-ndk
- feat: reanchor paths when parsed from yaml

## Checklist

- [x] CI is passing.

## Remark for PR creator


book/src/command_line.md
```diff
@@ -12,4 +12,15 @@ $ flutter_rust_bridge_codegen --help
...
\ No newline at end of file
+```
+
+## Configuration files
+
+You can also run `flutter_rust_bridge_codegen` with no arguments, provided one of these files exist (in order of priority):
+
+- `.frb.yml`
+- `.frb.yaml`
+- `.frb.json`
```
fzyzcjy, line 23: uber-nit: shall we use `flutter_rust_bridge` instead of `frb`, because when a reader reads "frb" in a repo, he may only google "frb" or "frb github" and never finds flutter_rust_bridge

frb_example/pure_dart/rust/.frb.yml
```diff
@@ -0,0 +1,10 @@
+rust_input: [src/api.rs]
```
fzyzcjy, line 1: nit: Given this frb.yml, maybe we should modify frb_codegen/src/main.rs's unit tests (introduced at https://github.com/fzyzcjy/flutter_rust_bridge/pull/1106) to use this config, thus remove duplication

frb_codegen/src/ir/ident.rs
```diff
@@ -1,9 +1,21 @@
...
 }
+}
+
+#[cfg(feature = "serde")]
+impl serde::Serialize for IrIdent {
+    fn serialize<S>(&self, s: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        s.serialize_newtype_struct("IrIdent", &self.raw)
```
fzyzcjy, line 16: uber-nit: why do we manually write down serialization? Is it possible we auto generate it (possibly with some annotations)

frb_codegen/src/utils.rs
```diff
@@ -210,3 +210,43 @@ impl PathExt for std::path::Path {
         }
     }
 }
+
+/// For structs that only has an `inner` serializable attribute that
+/// would be better (de)serialized as a newtype.
+#[macro_export]
+macro_rules! derive_serde_as_newtype {
```
fzyzcjy, line 217: uber-nit:

```suggestion
macro_rules! derive_serde_inner_as_newtype {
```

frb_codegen/src/main.rs
```diff
@@ -1,21 +1,46 @@
...
+            for location in CONFIG_LOCATIONS {
+                let Ok(file) = File::open(location) else {continue};
+                break 'from_file serde_yaml::from_reader(file)
+                    .map_err(|err| anyhow::anyhow!("Could not parse {location}: {err}"))?;
+            }
+            _ = err.print();
+            eprintln!("Hint: To call flutter_rust_bridge_codegen with no arguments, fill in .frb.yml with your config.");
+            std::process::exit(1)
+        }
+    };
```
fzyzcjy, line 30: uber-nit: maybe extract to a function, thus easier to read (eg no "break from_file"), also makes intent a little bit cleaner

frb_codegen/src/parser/mod.rs
```diff
@@ -631,6 +633,29 @@ impl Parse for DefaultValues {
...
+            }
+            Self::Float(value) => {
+                s.serialize_newtype_variant(NAME, 3, "Float", &value.base10_parse::<f64>().unwrap())
+            }
+            Self::Vec(value) => {
+                s.serialize_newtype_variant(NAME, 4, "Vec", &value.into_iter().collect::<Vec<_>>())
+            }
+        }
+    }
+}
```
fzyzcjy, line 657: uber-nit: hmm is it possible we do not write down this many boilerplate, but instead let serde generate for us?

fzyzcjy (commented):
Good job! Only some nits and ready to merge

frb_codegen/src/parser/mod.rs
Desdaemon: No problem, I wrote this so that I wouldn't have to convert it thru a serializable shell struct first.

frb_example/pure_dart/rust/.frb.yml
Desdaemon: Running into issues with changing working directories in one test messing up another. Perhaps a `--config-file` options is in order?

fzyzcjy: > Running into issues with changing working directories in one test messing up another.

Hmm not sure the exact issue. One way may be, shall we let frb think the base directory is the dir with .frb.yml?

Desdaemon: Better idea, since we aren't using positional arguments for the CLI, we could have it accept a single config file or a directory containing one, and we add the same entry point for the library as well.

Desdaemon:
Preferably I'd have liked to have this usage line:

```
flutter_rust_bridge_codegen [CONFIG_FILE]
flutter_rust_bridge_codegen --rust-input <RUST_INPUT> --dart_input <DART_INPUT> [OPTIONS]
```

But I'm not very good at clap-foo, it's not a dealbreaker but if anyone knows how to accomplish this it would be wonderful.

Desdaemon:
After CI goes green, I think it's a good stopping point to merge.

fzyzcjy (approved):

fzyzcjy:
Good job!

Issue #1103, opened 2023-03-05
[Bug] raw strings don’t get renamed before being transferred to dart (`r#`)


alexthe2:
### Describe the bug

When a variable is called for example `type` in rust, it must be preceeded with a `r#` (tonic does this automatically), however `r#type` is not a valid variable name in dart.

### Codegen logs with `RUST_LOG=debug` environment variable

```shell
-
```

alexthe2:
It should be an easy fix, I can check it out later, just wanted to create the bug report

fzyzcjy:
Totally agree this is a bug if that happens. Looking forward to your PR!

alexthe2:
Well guess what, if you put it into a function, everything breaks and you can’t rebuild it anymore

alexthe2:
even after just clean the generation still breaks

alexthe2:
It seems that the `syn` module failes to parse them for functions

fzyzcjy:
@alexthe2 hmm... maybe we should file an issue at `syn`, e.g. with reproducible samples?

alexthe2:
I filed one without reproducible samples and got shut down 🤣, I can file one at a later point this week

fzyzcjy:
Sure, take your time! IMHO making a reproducible sample is really important, otherwise it is time-consuming to even just figure out what's the exact problem

alexthe2:
of course, I just hoped I’d get away without reading the documentation and/or understanding it

fzyzcjy:
I guess maybe you can just copy-paste syn's demo, and add a `r#` that you see buggy into it, and see what's going on

alexthe2:
I tried, and it works for the module, so that means that issue is in the bridge, I’ll try to figure it out at one point

fzyzcjy:
I see. Take your time and looking forward to the fix!

PR #1139, opened 2023-03-21
Added support for raw strings

alexthe2:
## Changes

#1103 

## Checklist

- [x] CI is passing.

## Remark for PR creator


fzyzcjy (commented):
Good feature! Only nits :)

frb_codegen/src/generator/dart/ty_struct.rs
```diff
@@ -85,7 +86,7 @@ impl TypeDartGeneratorTrait for TypeStructRefGenerator<'_> {
         if has_methods {
             inner.insert(0, "bridge: this,".to_string());
         }
-        let inner = inner.join("\n");
+        let inner = rename_string_literal(inner.join("\n").as_str());
```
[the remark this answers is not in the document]
alexthe2: I fixed that, and even implemented it for functions, however the sourcegrapher parser failes to parse those functions so I left the test commented out and added a short explanation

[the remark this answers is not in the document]
alexthe2: but now it renames between rust and dart

alexthe2:
I think now it should be ready to review @fzyzcjy

alexthe2:
Nvm I guess

alexthe2:
whut the test passes locally

alexthe2:
Now it passes, seems that the web_test might be non deterministic (had that locally a few features ago)

frb_codegen/src/ir/ident.rs
```diff
@@ -35,4 +35,10 @@ impl IrIdent {
     pub fn dart_style(&self) -> String {
         self.raw.to_case(Case::Camel)
     }
+
+    pub fn remove_raw_prefix(&mut self) {
+        if self.raw.starts_with("r#") {
+            self.raw = self.raw[2..].to_string();
+        }
```
fzyzcjy, line 43: sth like

```suggestion
        self.raw = self.raw.strip_prefix("r#");
```

https://doc.rust-lang.org/std/string/struct.String.html#method.strip_prefix

(and tweak a bit about ownership)

frb_codegen/src/lib.rs
```diff
@@ -66,11 +66,13 @@ pub fn frb_codegen_multi(
     let raw_ir_file = config.get_ir_file()?;
 
     info!("Phase: Transform IR");
-    let ir_file = transformer::transform(raw_ir_file);
+    let mut ir_file = transformer::transform(raw_ir_file);
 
     info!("Phase: Generate Rust code");
     let generated_rust = generate_rust_code(config, &ir_file)?;
 
+    ir_file.remove_raw_prefix();
```
fzyzcjy, line 74: Ha you are pointing to an interesting place: `transformer::transform`. So, shall we do the remove-prefix thing inside the `transform`? Because a transform does nothing but transforming sth in the ir, thus our case perfectly fit it

fzyzcjy (approved):

fzyzcjy:
Good job! Now ready to review after nit changes

frb_codegen/src/lib.rs
alexthe2: I’d agree with you here 100%, but in rust it should stay `r#`, so unless I add this as a transformer function, idk sth. like `transformer::convert_for_dart(ir_file)`, I cannot do it before

alexthe2: should I add it for the one function?

fzyzcjy: Ah I see. Then the IR seen by rust and dart differs...

Then is it possible we do it somewhere else. For example, in IrStruct, when we are reading a field name, we give two function like `rust_field_name() { self.naive_fieldname }` and `fn dart_field_name() {self.naivefieldname.strip_prefix()}`.  You know, just like how today we read rust type names vs dart type names vs other things.

alexthe2: puuuuh okay, but then most functions right now are generalised and then they’ll also need two individual implementations, don’t they?

Am not 100% sure, haven’t given it a proper look yet

fzyzcjy: No? Only the function that reads *field name of a struct* so I guess should change <20 lines. If too many, just tell me!

alexthe2: Ok will check it out

alexthe2: I might have worked up a bit of magic or I just realised how damn stupid I was before, you choose

frb_codegen/src/ir/ident.rs
```diff
@@ -33,6 +33,10 @@ impl IrIdent {
     }
 
     pub fn dart_style(&self) -> String {
-        self.raw.to_case(Case::Camel)
+        self.raw
+            .strip_prefix("r#")
+            .unwrap_or(self.raw.as_str())
+            .to_string()
+            .to_case(Case::Camel)
```
fzyzcjy, line 40: Great, that's it!

fzyzcjy:
Will merge after CI passes :)

PR #1144, opened 2023-03-24
added more unit tests

alexthe2:
## Changes

## Checklist

- [ ] The code generator is run and the code is formatted (via `just precommit`).
- [ ] CI is passing.

## Remark for PR creator

### DIFF
--- a/frb_example/pure_dart/dart/lib/main.dart
+++ b/frb_example/pure_dart/dart/lib/main.dart
@@ -628,6 +628,15 @@ void main(List<String> args) async {
   //   expect(output.type, "not a type ;')");
   // });
 
+  test('test dart test more than jus ton eraw string struct', () async {
+    final output = await api.testMoreThanJustOneRawStringStruct();
+    expect(output, isA<MoreThanJustOneRawStringStruct>());
+    expect(output.regular, "regular");
+    expect(output.type, "type");
+    expect(output.async, true);
+    expect(output.another, "another");
+  });
+
   group('Platform-specific support', () {
     test('Int64List', () {
       final list = Int64List.fromList([-1, -2, -3, -4, -5]);
--- a/frb_example/pure_dart/dart/lib/bridge_definitions.dart
+++ b/frb_example/pure_dart/dart/lib/bridge_definitions.dart
@@ -604,6 +604,10 @@ abstract class FlutterRustBridgeExampleSingleBlockTest {
 
   FlutterRustBridgeTaskConstMeta get kTestRawStringItemStructConstMeta;
 
+  Future<MoreThanJustOneRawStringStruct> testMoreThanJustOneRawStringStruct({dynamic hint});
+
+  FlutterRustBridgeTaskConstMeta get kTestMoreThanJustOneRawStringStructConstMeta;
+
   Future<List<Weekdays>> listOfPrimitiveEnums({required List<Weekdays> weekdays, dynamic hint});
 
   FlutterRustBridgeTaskConstMeta get kListOfPrimitiveEnumsConstMeta;
@@ -1210,6 +1214,20 @@ class MessageId {
 
   const MessageId({
     required this.field0,
+  });
+}
+
+class MoreThanJustOneRawStringStruct {
+  final String regular;
+  final String type;
+  final bool async;
+  final String another;
+
+  const MoreThanJustOneRawStringStruct({
+    required this.regular,
+    required this.type,
+    required this.async,
+    required this.another,
   });
 }
 
--- a/frb_example/pure_dart/dart/lib/bridge_generated.dart
+++ b/frb_example/pure_dart/dart/lib/bridge_generated.dart
@@ -2282,6 +2282,22 @@ class FlutterRustBridgeExampleSingleBlockTestImpl implements FlutterRustBridgeEx
         argNames: [],
       );
 
+  Future<MoreThanJustOneRawStringStruct> testMoreThanJustOneRawStringStruct({dynamic hint}) {
+    return _platform.executeNormal(FlutterRustBridgeTask(
+      callFfi: (port_) => _platform.inner.wire_test_more_than_just_one_raw_string_struct(port_),
+      parseSuccessData: _wire2api_more_than_just_one_raw_string_struct,
+      constMeta: kTestMoreThanJustOneRawStringStructConstMeta,
+      argValues: [],
+      hint: hint,
+    ));
+  }
+
+  FlutterRustBridgeTaskConstMeta get kTestMoreThanJustOneRawStringStructConstMeta =>
+      const FlutterRustBridgeTaskConstMeta(
+        debugName: "test_more_than_just_one_raw_string_struct",
+        argNames: [],
+      );
+
   Future<List<Weekdays>> listOfPrimitiveEnums({required List<Weekdays> weekdays, dynamic hint}) {
     var arg0 = _platform.api2wire_list_weekdays(weekdays);
     return _platform.executeNormal(FlutterRustBridgeTask(
@@ -3105,6 +3121,17 @@ class FlutterRustBridgeExampleSingleBlockTestImpl implements FlutterRustBridgeEx
     if (arr.length != 1) throw Exception('unexpected arr length: expect 1 but see ${arr.length}');
     return MessageId(
       field0: _wire2api_u8_array_32(arr[0]),
+    );
+  }
+
+  MoreThanJustOneRawStringStruct _wire2api_more_than_just_one_raw_string_struct(dynamic raw) {
+    final arr = raw as List<dynamic>;
+    if (arr.length != 4) throw Exception('unexpected arr length: expect 4 but see ${arr.length}');
+    return MoreThanJustOneRawStringStruct(
+      regular: _wire2api_String(arr[0]),
+      type: _wire2api_String(arr[1]),
+      async: _wire2api_bool(arr[2]),
+      another: _wire2api_String(arr[3]),
     );
   }
 
--- a/frb_example/pure_dart/dart/lib/bridge_generated.io.dart
+++ b/frb_example/pure_dart/dart/lib/bridge_generated.io.dart
@@ -3369,6 +3369,19 @@ class FlutterRustBridgeExampleSingleBlockTestWire implements FlutterRustBridgeWi
       _lookup<ffi.NativeFunction<ffi.Void Function(ffi.Int64)>>('wire_test_raw_string_item_struct');
   late final _wire_test_raw_string_item_struct = _wire_test_raw_string_item_structPtr.asFunction<void Function(int)>();
 
+  void wire_test_more_than_just_one_raw_string_struct(
+    int port_,
+  ) {
+    return _wire_test_more_than_just_one_raw_string_struct(
+      port_,
+    );
+  }
+
+  late final _wire_test_more_than_just_one_raw_string_structPtr =
+      _lookup<ffi.NativeFunction<ffi.Void Function(ffi.Int64)>>('wire_test_more_than_just_one_raw_string_struct');
+  late final _wire_test_more_than_just_one_raw_string_struct =
+      _wire_test_more_than_just_one_raw_string_structPtr.asFunction<void Function(int)>();
+
   void wire_list_of_primitive_enums(
     int port_,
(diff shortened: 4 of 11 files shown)


alexthe2:
I got some issues in one of my projects, that I can’t reproduce in the bridge, but maybe one of these tests will fail

alexthe2:
If they don’t fail I just added more tests, but I’ll try a bit more, so don’t merge it yet

alexthe2:
found the broken part, mirrors. They seem to break everything

alexthe2:
If you don’t mind @fzyzcjy , can you explain me quickly how mirrors work exactly / is there some docs about them

fzyzcjy:
https://cjycode.com/flutter_rust_bridge/feature/lang_external.html

Also cc @Unoqwy who implemented the good mirror feature

alexthe2:
Finally got the tests to reproduce exactly my issue, will try to fix it now

alexthe2:
So the problem is that when an Enum is mirrored, its data types are stayed as regular, not the mirrored structs

alexthe2:
I got a first draft that seems to work, but still have 2 edge cases I need to figure out, list of enums and enums in pai.rs

alexthe2:
Ok so in the current master version list of mirrored enums don’t work either, meaning that my changes right now have at least fixed enums that contain mirrored structs, I still have to do some refactoring but overall I’m finished with this part, list of enums will be a seperate pull request, let’s approach this one at a time

fzyzcjy:
Looks great! I will probably review today :)

alexthe2:
> Looks great! I will probably review today :)

I need to sleep 😴 , and some of the code requires a bit of rethinking, so when u review ignore this function:
```rust
fn extract_type(fields: &Fields) -> Option<String> {
        let f = match fields {
            Fields::Named(_) => None,
            Fields::Unnamed(u) => Some(u.to_token_stream()),
            Fields::Unit => None,
        }
        .map(|s| {
            s.to_string()
                .trim_start_matches('(')
                .trim_end_matches(')')
                .to_string()
        });

        if f.is_some()
            && vec!["String", "bool", "i32", "i64", "f32", "f64"]
                .contains(&f.as_ref().unwrap().as_str())
        {
            return None;
        }

        f
    }
```

I need to reorder it

alexthe2:
Nvm, I guess. It was a very short refactor. If you can review it, it would be great. I can then change it later today (after I finally sleep).

If there are no nits and all tests pass(which I don't believe), then for the changelog would be:
- Bumped up syn, to latest v1 version
- Fixed enum containing structs
- Added a bunch of unit tests to make sure it works for the future (because it worked before, so some feature/refactor must have broken it)

fzyzcjy:
Sure, take your time! Btw for the changelog - could you please modify the PR title, such that it is easier to read for future readers :)

frb_example/pure_dart/dart/lib/main.dart
```diff
@@ -628,6 +628,44 @@ void main(List<String> args) async {
   //   expect(output.type, "not a type ;')");
   // });
 
+  test('test dart test more than jus ton eraw string struct', () async {
```
fzyzcjy, line 631:
```suggestion
  test('test dart test more than just one raw string struct', () async {
```

frb_codegen/src/ir/field.rs
```diff
@@ -7,6 +7,7 @@ crate::ir! {
 pub struct IrField {
     pub ty: IrType,
     pub name: IrIdent,
+    pub wrapper_type: Option<String>,
```
fzyzcjy, line 10: nit: May I know what is "wrapper_type" vs "ty", or can we infer the wrapper_type from the ty?

If we need to keep this field, would be great to add a bit of comments, explaining to readers this difference

fzyzcjy (commented):
Nice fix! Not checked into detailed logic yet, want to know a bit about "wrapper_type" first

frb_codegen/src/ir/field.rs
alexthe2: depends, It could maybe be a simple bool, I can give it a try

alexthe2:
how does that look?, not happy with the hardcoding though

fzyzcjy:
Quite busy now, I will check it soon (probably within a few hours) :)

alexthe2:
> Quite busy now, I will check it soon (probably within a few hours) :)

don’t worry, not pressing (maybe a bit because everything is broken on my side in the project)

fzyzcjy:
> maybe a bit because everything is broken on my side in the project

Ah then I should do it quicker!

frb_codegen/src/generator/rust/ty_enum.rs
```diff
@@ -232,7 +232,16 @@ impl TypeRustGeneratorTrait for TypeEnumRefGenerator<'_> {
                                     self.context.ir_file,
                                     self.context.config,
                                 );
-                                gen.convert_to_dart(field.name.rust_style().to_owned())
+                                let struct_name = field.ty.mirrored_nested();
+                                if field.mirrored_enum && struct_name.is_some() {
+                                    format!(
+                                        "mirror_{}({}).into_dart()",
+                                        struct_name.unwrap(),
+                                        field.name.rust_style().to_owned()
```
fzyzcjy, line 240: Disclaimer: Busy with my code so the suggestions may be inperfect!

Hmm if I understand correctly, this fix is like: The `IrField.ty`'s type name is the original struct name, instead of the mirrored struct name, but we want the latter. Thus, here we explicitly hack it to add `mirror_` prefix.

However, this looks like a general problem: What if other places also wants the mirrored name instead of the old name? But the current solution adds `mirrored_enum` flag *inside* IrField.

Ideally, maybe we can have somewhere *general* (instead of tied to IrField) showing whether a type is a mirrored enum. For example, can we ad a field in `IrType`?

Then we can do sth like

```
impl IrType {
  fn the_name_with_possibly_mirror() { ... }
}
```

and use it here happily.

Or, if we are sure every place wants the mirror instead of original type, then just make `the_name_with_possibly_mirror` the default choice, e.g. name it as `name()`

alexthe2: Hmm, ok I’ll look into that

alexthe2: The issue is solely for enums, so while I get the idea, I don’t think it’s useful to change the entire structure. I added a bunch more tests to make sure

alexthe2: what do you think of this way? I added the settings field, cause it feels a bit more *clean* then starting to bloat the IrField

fzyzcjy: Hmm I am still a bit confused.

The current code is: For each *field*, if the *type* of the field is indeed a mirror, do extra log to code generation (addd "mirror_").

However, we know it is *type* that can be a mirror. So what I proposes is: Is it possible we add `is_mirror` or something like that onto the *type*, instead of adding it as a flag for the *field*?

alexthe2:
Ohh, I get what you mean. I can change that very quickly

frb_codegen/src/generator/rust/ty_enum.rs
alexthe2: Nvm, this is impossible because type is an enum ;)

alexthe2: Technically it is possible, but that would be a lot of code, for an edge case. Because then it would have to become a type trait…, and be implemented for all of them. and then again this is only an issue for enums but the code generation would change a lot

alexthe2: should I change it, or are we sticking to this approach?

fzyzcjy: > that would be a lot of code, for an edge case

IMHO it should be only a few dozen lines of code.

alexthe2: Yes and no, the trait implementation would have to be for all types right? + The current ty approach would have to change to be changed at a later point (that's rather fine)

fzyzcjy: Let me use a few minutes and see whether I can make a patch demonstrating my idea

fzyzcjy:
Goal: If a type is a mirror, the name of it should be `mirror_ThatType` instead of `ThatType`.

Code prototype:

[TODO]

frb_codegen/src/generator/rust/ty_enum.rs
alexthe2: Ofc, I just took another look myself but it would mean that:
1. New function in IrType, that essentially does the same mirrored_nested (easily doable)
 a. Added field in StructRef, that is then used by the function
2. The current logic does more or less sth in this way:
```rust
let gen = TypeRustGenerator::new(
    field.ty.clone(),
    self.context.ir_file,
    self.context.config,
);

gen.convert_to_dart(
    field.try_name_mirror(field.name.rust_style().to_owned()),
)
```
we would have to rewrite the logic both in `ty_enum` and `ty_struct`, and they right now use `field.name` everywhere. Then essentially type needs to know its field or the field function passes itself along to the type? 

IMHO this backlink sounds rather ugly, but maybe I’m missing sth in the code

fzyzcjy: I see the problem: The bug is caused by missing a newtype wrapper, so the current way may be OK, let me check it.

frb_codegen/src/ir/field.rs
```diff
@@ -64,4 +72,15 @@ impl IrField {
 
         format!("{enum_name}.{variant_name}")
     }
+
+    /// Takes the provided name and surrounds it with the `mirror_` prefix if the field is in a
+    /// mirrored enum.
+    pub fn try_name_mirror(&self, name: String) -> String {
+        let mirrored_name = self.ty.mirrored_nested();
+        if self.settings.is_in_mirrored_enum && mirrored_name.is_some() {
```
fzyzcjy, line 80: Wait a minute, maybe we already find the answer: Is it possible we simply remove `is_in_mirrored_enum` flag, because is_in_mirrored_enum can be derived from `mirrored_name.is_some()`?

alexthe2: I hoped so as well, but I did a test and sadly there were edge cases in which it was not. I also added a few more unit tests for this

alexthe2: I will try to fix `Vec<Enums>` at one point, so maybe then I’ll find a better way for it

alexthe2:
> (notes for myself)
> 
> Goal: When generating into-dart logic, if the type of a field is a mirror, we should add `mirror_SomeType(...)` wrapper around it.
> 
> Current code: Add "is-mirror" information into the IrField struct.
> 
> Suggestion: Derive "is-mirror" information from the IrType.

small extension to this: This became a new bug, I don’t exactly know what version caused it, my guess is either 1.70, 1.71 or 1.72, because it worked before

fzyzcjy:
I see... Then maybe hack like this temporarily :(

alexthe2:
> I see... Then maybe hack like this temporarily :( Since I do not have more time for this edge case

any nits or can we merge?

frb_codegen/src/ir/field.rs
```diff
@@ -64,4 +72,15 @@ impl IrField {
 
         format!("{enum_name}.{variant_name}")
     }
+
+    /// Takes the provided name and surrounds it with the `mirror_` prefix if the field is in a
+    /// mirrored enum.
```
fzyzcjy, line 77:
```suggestion
    /// Takes the provided name and surrounds it with the `mirror_` prefix if the field is in a
    /// mirrored enum.
    /// This may be improved in the future: https://github.com/fzyzcjy/flutter_rust_bridge/pull/1144#issuecomment-1486569869
```

fzyzcjy (approved):

fzyzcjy:
No nits, good job!

alexthe2:
Cool, the release will be out once the release pipelines pass, won't it?

fzyzcjy:
Executing release pipeline on my machine :)

image

fzyzcjy:
Ping me if it does not appear after an hour since this one usually takes only a dozen (?) minutes

alexthe2:
Oh ok, thank you

fzyzcjy:
Indeed done now!

You are welcome :)

image

PR #1149, opened 2023-03-30
fix: Handle fallible case for TypeGeneralListGenerator.

### LINKED ISSUE #1109: [Bug] Mirrored Data Type cannot return Result<Vec<Struct>>
### Describe the bug

When a datatype is mirrored, returning `anyhow::Result<Vec<Struct>>` leads to an unwrap error in the generated code

```shell
-
```

### To Reproduce

_No response_

### Expected behavior

_No response_

### Generated binding code

_No response_

### OS

_No response_

1.69.0

### Flutter info

_No response_

### Version of `clang++`

_No response_

### Version of `ffigen`

_No response_

### Additional context

_No response_

trobanga:
## Changes

Fixes #1109

## Checklist

- [x] CI is passing.

## Remark for PR creator

### DIFF
--- a/frb_codegen/src/generator/rust/ty_general_list.rs
+++ b/frb_codegen/src/generator/rust/ty_general_list.rs
@@ -48,20 +48,18 @@ impl TypeRustGeneratorTrait for TypeGeneralListGenerator<'_> {
         inner
             .wrapper_struct()
             .map(|wrapper| {
-                if wired_fallible_func {
-                    format!(
-                        "{}.map(|s| s.into_iter().map(|v| {}({})).collect::<Vec<_>>())",
-                        obj,
-                        wrapper,
-                        inner.self_access("v".to_owned())
-                    )
-                } else {
+                let infallible_part = |name| {
                     format!(
                         "{}.into_iter().map(|v| {}({})).collect::<Vec<_>>()",
-                        obj,
+                        name,
                         wrapper,
                         inner.self_access("v".to_owned())
                     )
+                };
+                if wired_fallible_func {
+                    format!("{}.map(|s| {})", obj, infallible_part("s"))
+                } else {
+                    infallible_part(&obj)
                 }
             })
             .unwrap_or(obj)
--- a/frb_codegen/src/generator/rust/ty_general_list.rs
+++ b/frb_codegen/src/generator/rust/ty_general_list.rs
@@ -38,21 +38,31 @@ impl TypeRustGeneratorTrait for TypeGeneralListGenerator<'_> {
         ])
     }
 
-    fn wrap_obj(&self, obj: String, _wired_fallible_func: bool) -> String {
+    fn wrap_obj(&self, obj: String, wired_fallible_func: bool) -> String {
         let inner = TypeRustGenerator::new(
             *self.ir.inner.clone(),
             self.context.ir_file,
             self.context.config,
         );
+
         inner
             .wrapper_struct()
             .map(|wrapper| {
-                format!(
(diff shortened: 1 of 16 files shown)


fzyzcjy:
Good job! In order to ensure we have fixed it and never introduce the bug again in the future, it would be great to add a test.

> End-to-end tests are usually in the ./frb_example/pure_dart example, more specifically, rust/src/api.rs and dart/lib/main.dart.

trobanga:
I've added a test but I have problems with generating the code. Somehow some types e.g. bool are replaced with NativeFunction<...>. In the past I could solve this by using cd / && flutter_rust_bridge_codgen and adding some llvm compiler opts: `--llvm-compiler-opts=-I/usr/lib/clang/15.0.7/include`

trobanga:
In `just generate_bridge` I tried to replace {{cargo_run_codegen}} with cd / && flutter_rust_bridge_codgen but it didn't work. I'm using Arch Linux btw.

trobanga:
Ok, I've found the documentation about setting CPATH and it seems to work now.

frb_example/pure_dart/rust/.flutter_rust_bridge.yml
```diff
@@ -7,4 +7,6 @@ dart_enums_style: true
 c_output: [c_output_path/c_output.h]
 extra_c_output_path:
   - c_output_path_extra/
+llvm-compiler-opts:
+  [-I/usr/lib/clang/15.0.7/include]
```
fzyzcjy, line 11: uber-nit: Hmm I am not very sure - shall we hardcode `15.0.7` here? Since other users when using this example may not have this version

frb_codegen/src/generator/rust/ty_general_list.rs
```diff
@@ -38,21 +38,31 @@ impl TypeRustGeneratorTrait for TypeGeneralListGenerator<'_> {
...
+                        inner.self_access("v".to_owned())
+                    )
+                } else {
+                    format!(
+                        "{}.into_iter().map(|v| {}({})).collect::<Vec<_>>()",
+                        obj,
+                        wrapper,
+                        inner.self_access("v".to_owned())
+                    )
+                }
```
fzyzcjy, line 65: uber-nit: maybe we can extract the common code part of the two branches

fzyzcjy (approved):
Nice, ready to merge after nits!

frb_example/pure_dart/rust/.flutter_rust_bridge.yml
trobanga: Sry, I simply forgot to remove it...with setting CPATH it isn't necessary anymore.

frb_codegen/src/generator/rust/ty_general_list.rs
trobanga: I don't really see a way to do this easily, while remain readability. So I would keep it like this.

fzyzcjy: maybe sth like

```rust
fn f(name: &str) { format!("{}.into_iter().map(|v| {}({})).collect::<Vec<_>>()", name, wrapper, ...) }
if wired_fallible_func {
  format!("{}.map({})", obj, f("s"))
} else {
  f(obj)
}

trobanga: Ok, this should do it.

dbsxdbsx:
@fzyzcjy, did you just forget to review? I saw this pr didn't update after the last change 5 days ago.😀

fzyzcjy:
Definitely! Thanks for pinging me, I will review probably within a few hours.

PR #1145, opened 2023-03-25
Introduce auto-generated shared API block(file) for case multi-blocks

dbsxdbsx:
## Changes
This pr is decided to introduce a new mechanism to handle flaws when using frb with multi-blocks.

## idea
The idea is simple---moving all stuff(the shared user-defined structs, methods of the structs, or some other shared API like `free_WireSyncReturn`) shared(duplicated/conflicted before) to another auto-generated API block.

## Checklist
- [x] make sure this approach is workable(test if it is compilable)--- Yes, it is workable/runnable.
- [x] fix bug #987 --- no redundant type would be produced in regular API-generated file(s) since these shared types are now moved to another shared file(both in Rust and Dart).
- [x] fix bug #1111 --- this is solved apparently. For details, please refer  [here](https://github.com/fzyzcjy/flutter_rust_bridge/pull/1145#issuecomment-1661685437).
- [x] fix bug #956 --- methods for a simple `struct`  is solved, no matter where they are defined(API blocks or customized rust module files). But methods related to an `Enum` is not solved(rust-gen code is ok, dart-gen code is NOT ok, and the corresponding dart test code is added but commented out). Leave it to future pr since this pr is already too heavy.--- Still some issue left.
- [x] add other type tests, types like `list`(`vec`), `enum`, `option`, and `SyncReturn` types, and their combination are tested,(other complex types like `stream`, and `opaque type` are not tested on purpose---leave them in future pr)...
- [x] delete debug code
- [x] refine test cases for multi-blocks --- the rust/dart test cases are totally rewritten from scratch with lots of useful comments for readability.
- [x]  CI is passing --- make sure  everything is ok with(out) various flag, like  `wasm`, `dart_decl_output`(MacOs with multi-blocks is not tested partly due to #1225)
- [x] refine documents

Issue #947, opened 2023-01-02
Add a "fat" option for codegen


GregoryConrad:
# The problem
The issue presented here is still only theoretical with no actual issues caused by it (AFAIK), but I imagine it could present substantial issues in the near future.

At the moment, projects/libraries using FRB need to use the same FRB version amongst Dart and Rust, or else the support libraries might not match up properly. While this makes sense, this can present issues.

Take the following configuration:
```yaml
# A's pubspec.yaml
dependencies:
  b: ^1.0.0
  c: ^1.0.0

# B's pubspec.yaml
dependencies:
  flutter_rust_bridge: "1.58.0"

# C's pubspec.yaml
dependencies:
  flutter_rust_bridge: "1.59.0"
```

Note how the Dart/Rust libraries (B/C) need an exact version of FRB; this is so the support library versions will match up correctly. Otherwise, the project may fail to work correctly, or not even compile.

While using an exact version is the correct fix from the library's perspective, it will create a problem for the end user that relies upon 2+ FRB libraries. Obviously, 1.58.0 and 1.59.0 are not compatible, and Dart will fail to resolve dependencies, or perhaps may force one library to be constrained to an old version (if it can even find one).

Thus, I am proposing a new feature: a "fat" option in FRB itself.

# Proposed solution
Add a new option, I will call it "fat", to codegen that builds "fat" Dart files. These "fat" Dart files bundle the necessary FRB support code alongside the generated FFI interop code in the same Dart files. This would remove the Dart FRB package dependency, fixing the problem altogether. In the future, once/if this solution is implemented, it also might be a good idea to discontinue the Dart package altogether and simply force the use of this "fat" option.

fzyzcjy:
Good idea! Just one nit: We should not lose the advantage of a normal *package* (the frb_dart for example) - For example, frb_dart as a package can be linted, easily unit tested, etc. Thus, maybe we can still make frb_dart a package (just like currently), and *copy* the content of the package into the correct location during codegen (when enabling) fat.

GregoryConrad:
Was looking into this a bit more (briefly), because this issue should be addressed sooner rather than later with the amount of people that have been following the "make a package/library" part of the guide. If they publish to pub.dev, chances are some user somewhere will run into this issue eventually.

[...]

PR #1164, opened 2023-04-10
Make generated C files unique to prevent collisions

### LINKED ISSUE #1163: Support `dummy_method_to_enforce_bundling` taking class name with only 1 rust input.

Currently, if there is only 1 rust input provided, the generated `--c-output`'s `dummy_method_to_enforce_bundling` will not have a class name. Recently, a user of my project ran into a problem with this. They also use FRB in their project. However, the CI fails to compile on macOS for them because there are multiple definitions of `dummy_method_to_enforce_bundling`.

If `class-name` is provided and there is only 1 rust input, the dummy method should include the class name. I am willing to make a PR to add this functionality.

**For example:**

`flutter_rust_bridge_codegen --rust-input path/to/code.rs`
```
dummy_method_to_enforce_bundling
```
`flutter_rust_bridge_codegen --rust-input path/to/code.rs --class-name MyApi`
```
dummy_method_to_enforce_bundling_MyApi
```

One way to fix this issue is for me to create a dummy rust input. This allows the class name to be set. However because of this, many useless files are generated as well.

**Additional context**

Issue submitted to my project: https://github.com/erikas-taroza/simple_audio/issues/22

erikas-taroza:
## Changes
Fixes #1163

If `--class-name` is passed with only 1 Rust input, the class name is included in the dummy method.
**Example:**
`flutter_rust_bridge_codegen --rust-input path/to/code.rs --class-name MyApi`
```
dummy_method_to_enforce_bundling_MyApi
```

If there are 2 class names, then the default/fallback is used for the combined dummy method.
**Example:**
`flutter_rust_bridge_codegen --rust-input path/to/one.rs path/to/two.rs --class-name MyApiOne MyApiTwo`
```
dummy_method_to_enforce_bundling_CargoPackageNameInPascal
```

For single Rust inputs, the change is not breaking. However for the 2+ Rust inputs, the change is breaking.

I'm thinking about adding a new arg (ex. `--base-class-name`) that would prepend the real class name.
**Example:**
`flutter_rust_bridge_codegen --rust-input path/to/one.rs path/to/two.rs --class-name MyApiOne MyApiTwo --base-class-name BaseClassName`

```
dummy_method_to_enforce_bundling_BaseClassName_MyApiOne
```
```
dummy_method_to_enforce_bundling_BaseClassName
```

And if the user doesn't provide a base class name, `_BaseClassName` will not be added.

Let me know what you think.

## Checklist

- [x] CI is passing.

## Remark for PR creator


erikas-taroza:
I added the `--base-class-name` arg to fix the breaking issue I described above.

Here are the expected results:

No class names:
```
dummy_method_to_enforce_bundling
```

Base class name, no class names:
```
dummy_method_to_enforce_bundling_BaseClassName
```

No base class name, x class names:
```
dummy_method_to_enforce_bundling_ClassName
```

Base class name, x class names:
```
dummy_method_to_enforce_bundling_BaseClassNameClassName
```

fzyzcjy:
Good job! I will probably review within a day :)

frb_codegen/src/generator/c/mod.rs
```diff
@@ -40,17 +41,26 @@ pub fn generate_dummy(
...
-                get_dummy_func("", &func_names)
+                get_dummy_func(base_class_name, &func_names)
             )
         } else {
             basic_dummy_func
         }
     } else {
-        get_dummy_func("", func_names)
+        let default_class_name = fallback_class_name(&config.rust_crate_dir).unwrap_or_default();
+        let class_name = match config.class_name == default_class_name {
```
fzyzcjy, line 57: uber-nit: `if (...) { ... } else { ... }`

fzyzcjy (commented):
not checked details yet, firstly a uber-nit :)

fzyzcjy:
So, before checking the detailed implementation, we may need to firstly consider the API design.

1. As is mentioned above, it seems that this introduces some breaking change. Is this avoidable? We know breaking changes are bad ;)
2. It feels a little bit counter-intuitive to see both "class name" and "base class name", and a user may need a bit of time to understand what are the two things. So is it possible we somehow come up with a nicer user-facing API

Also cc @dbsxdbsx who implemented this helpful multi-file feature

erikas-taroza:
> So, before checking the detailed implementation, we may need to firstly consider the API design.
> 
> 1. As is mentioned above, it seems that this introduces some breaking change. Is this avoidable? We know breaking changes are bad ;)
> 2. It feels a little bit counter-intuitive to see both "class name" and "base class name", and a user may need a bit of time to understand what are the two things. So is it possible we somehow come up with a nicer user-facing API
> 
> Also cc @dbsxdbsx who implemented this helpful multi-file feature

1. If we use just `--class-name` the change will be breaking in multi-file scenarios. There is no way to tell whether or not the user wants to make their dummy method unique. In a single file example, the change is not breaking because providing a class name is optional. If they happen to provide it, then the class name is added to the single dummy method.
2. I agree that it may be confusing. I think a more specific arg (ex. `--dummy-method-suffix`) would be better. Let me know what you think.

fzyzcjy:
> a more specific arg (ex. --dummy-method-suffix) would be better.

That sounds more specific to me.

Btw I am not sure whether this change will cause problems to some other proposals: Mainly https://github.com/fzyzcjy/flutter_rust_bridge/issues/947#issuecomment-1454807065 by @GregoryConrad (maybe in another issue but I cannot find it). In short, in the long term we are going to ensure symbols are not duplicated by adding postfix/prefix to *each and every* of them, not only the dummy_method_to_enforce_bundling one. Maybe this PR can address this problem as well uniformly?

Also cc @GregoryConrad

erikas-taroza:
> Maybe this PR can address this problem as well uniformly?

Sure!

In that case, should this be default behavior instead of needing an arg passed in? That way the user doesn't have to configure anything to be able to use multiple FRB libraries.

GregoryConrad:
@fzyzcjy I think you are looking for this: https://github.com/dart-lang/sdk/issues/50565#issuecomment-1455695630

Ideally, it shouldn't just be the dummy method getting a prefix/suffix, but rather every function as @fzyzcjy mentioned above.

erikas-taroza:
> @fzyzcjy I think you are looking for this: [dart-lang/sdk#50565 (comment)](https://github.com/dart-lang/sdk/issues/50565#issuecomment-1455695630)
> 
> Ideally, it shouldn't just be the dummy method getting a prefix/suffix, but rather every function as @fzyzcjy mentioned above. Every library should have a unique prefix/suffix, and perhaps FRB version should be added in as well.
> 
> @erikas-taroza Thanks for working on this! It's much needed.

I agree. I was asking because the dummy method, when changed, will need to be updated in the user's Swift/Obj-C code, making it break. I think that the uniqueness should be an OOTB feature.

I'll start working on this, converting back to draft.

GregoryConrad:
FWIW, I think the prefix/suffix should be opt-in in the FRB code gen config struct, so it won’t break existing code, but is there for library authors (and others) who need it. I think breaking the swift/objc code, if the feature is opt-in, should be a non-issue.

dbsxdbsx:
@fzyzcjy, thanks for ccing me. Frankly, I am out of bandwidth to get into this issue and pr, since I am working on the[ shared block issue](https://github.com/fzyzcjy/flutter_rust_bridge/pull/1145), which may erase the suffix with block_index for shared API, if it succeeds.

But generally, I hope no new flag is added unless it is essential. Better to make changes as transparent to the user as possible.

erikas-taroza:
I do not have access to a Mac at the moment so I cannot test if the changes solve the problem. I asked the person that submitted an issue on my repo but they haven't responded yet.

If anyone can test, please create a new Flutter project with FRB (you can use the [template](https://github.com/Desdaemon/flutter_rust_bridge_template)). Then, set up [my plugin](https://github.com/erikas-taroza/simple_audio/tree/test-frb-prefix) using the `test-frb-prefix` branch. Add this to your `pubspec.yaml` and follow the README steps:
```yaml
simple_audio:
    git:
      url: https://github.com/erikas-taroza/simple_audio
      branch: test-frb-prefix
```

That branch contains code that was generated by the changes in this PR. If everything works, it should be able to run on Mac. If it doesn't, this error should pop up:
```
'dummy_method_to_enforce_bundling' has different definitions in different modules; definition in module 'simple_audio.bridge_generated' first difference is function body
```

Thank you for your help.

erikas-taroza:
@fzyzcjy Can this be reopened? The user of my plugin that initially ran into this issue confirmed that the PR works on their apple device.

fzyzcjy:
Sure!

dbsxdbsx:
I just wonder if the change made by this PR is still meaningful, if [mine ](https://github.com/fzyzcjy/flutter_rust_bridge/pull/1145) is merged. Anyway, @fzyzcjy , it is up to you.

fzyzcjy:
Good question! I am busy and have not have time to review your PR (apologies!), but briefly looking at them this looks like a question. @erikas-taroza what do you think?

erikas-taroza:
> I just wonder whether the change made by this PR is still meaningful, if [mine ](https://github.com/fzyzcjy/flutter_rust_bridge/pull/1145) is merged. Anyway, @fzyzcjy , it is up to you.

If your PR solves the original issue, then I think it is fine to use your solution instead of mine. The original issue was when a user of my plugin that uses FRB tried to use their own Rust code with FRB. It caused a naming collision on macOS/iOS because of duplicate symbols.

PR #1169, opened 2023-04-11
test: fix rust_build_and_test error in macos

huang12zheng:
## Changes

_Please list issues fixed by this PR here, using format "Fixes #the-issue-number"._

## Checklist

- [ ] An issue to be fixed by this PR is listed above.
- [ ] New tests are added to ensure new features are working. End-to-end tests are usually in the `./frb_example/pure_dart` example, more specifically, `rust/src/api.rs` and `dart/lib/main.dart`.
- [x] The code generator is run and the code is formatted (via `just precommit`).
- [x] If this PR adds/changes features, documentations (in the `./book` folder) are updated.
- [x] CI is passing.

## Remark for PR creator

- Justfile is a task runner for the command line interface that allows you to run shell commands from a file named `justfile` in your project directory. You can use Justfile after [installing it](https://github.com/casey/just). Note that commands written in `justfile` of this repository are expected to be run in `bash`, not `cmd` or `powershell`. Running `just ...` commands in `cmd` or `powershell` will produce errors as the syntax is not compatible. On Windows, you can use `git bash` if you have Git installed.
- If fzyzcjy does not reply for a few days, maybe he just did not see it, so please ping him.

1. I get a error about `store_dart_post_cobject` if I run `rust_build_and_test` in macos.
so I have this pr.
2. the libname about `with_flutter` and `pure_dart` is the same, it maybe a bad way.
3. Add something about store_dart_post_cobject in other issus
4. `loadDylib` is divided into `loadLibForFlutter`, `loadLibForDart`

PR #1170, opened 2023-04-11
Remove bridge in dart model with option, ref to #1169

huang12zheng:
## Changes

## Checklist

- [x] CI is passing.

## Remark for PR creator

This pr is based on #1169
pr about #1116


huang12zheng:
Actually, I'd like you to take a look at 101b171 and see if there are any changes to be made.

fzyzcjy:
Hmm I see some common code between this and #1169, so let me first check that one

huang12zheng:
I close it due to that I want to test locally
what's more, I don't know how to no run ci : )

huang12zheng:
Ping @fzyzcjy

frb_codegen/src/generator/dart/mod.rs
```diff
@@ -63,7 +63,10 @@ pub fn generate(ir_file: &IrFile, config: &Opts, wasm_funcs: &[IrFuncDisplay]) -
         ..
     } = &spec;
     let needs_freezed = spec.needs_freezed;
-    let common_header = generate_common_header();
+    let mut common_header = generate_common_header();
+    if config.no_use_bridge_in_method {
+        common_header.import += "import 'ffi.io.dart' if (dart.library.html) 'ffi.web.dart';";
+    }
```
fzyzcjy, line 69: uber-nit: shall we put this `if` into `generate_common_header`?

frb_codegen/src/config/opts.rs
```diff
@@ -27,6 +28,7 @@ pub struct Opts {
     pub skip_deps_check: bool,
     pub wasm_enabled: bool,
     pub inline_rust: bool,
+    pub no_use_bridge_in_method: bool,
```
fzyzcjy, line 31: uber-nit: shall we name it `use_bridge_in_method` here, since it is usually to avoid such negations. The `no_...` in `RawOpts` is acceptable since we do want that flag for end users.

frb_example/with_flutter/macos/Podfile
```diff
@@ -1,4 +1,4 @@
-platform :osx, '10.11'
+platform :osx, '10.14'
```
fzyzcjy, line 1: hmm if the old version is good, maybe we should use the old one for more compatibility?

frb_example/with_flutter/rust/src/api.rs
```diff
@@ -103,3 +107,28 @@ pub fn off_topic_deliberately_panic() -> i32 {
     std::env::set_var("RUST_BACKTRACE", "1"); // optional, just to see more info...
     panic!("deliberately panic!")
 }
+
+pub struct SumWith {
+    pub x: u32,
+}
+
+impl SumWith {
+    pub fn sum(&self, y: u32) -> u32 {
```
fzyzcjy, line 116: uber-nit: would be great to mention that, we deliberately copy SumWith and other code into here because we want to test the flag

frb_example/with_flutter/lib/bridge_generated.dart
```diff
@@ -8,6 +8,7 @@ import 'dart:async';
 import 'package:meta/meta.dart';
 import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
 import 'package:uuid/uuid.dart';
+import 'ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
```
fzyzcjy, line 11: Main issue: I am not sure, but seems that the `late final api = ...` can happen *anywhere* in the user code, instead of hardcoded into this file.

Thus, maybe we can do something like this:

1. Add a command line argument like `--extra-header-dart`. The implementation is simple - just copy whatever user provides into the top of these generated dart files.
2. `--extra-header-dart "import 'ffi.io.dart' if (dart.library.html) 'ffi.web.dart';"` for this specific frb example.
3. No more need to hardcode importing which file.

fzyzcjy (commented):
Good job! Mainly nits :)

frb_codegen/src/config/opts.rs
huang12zheng: I am follow code `no_build_runner: config.no_build_runner`.
Although my point of view is that it's not a good idea to change internally

frb_example/with_flutter/lib/bridge_generated.dart
huang12zheng: I think convention is better than configuration.
If it must be, it could do another pr.

frb_codegen/src/generator/dart/mod.rs
huang12zheng: This is bad because there is no config information.
generate_common_header actually gets an initial value, so there should be nothing wrong with that.
Initial value, and then processing.
If you add config, it doesn't mean and initial value.

But I can change it if you want-_-||

frb_example/with_flutter/lib/bridge_generated.dart
fzyzcjy: Hmm, But this forces everyone to use ffi.io.dart and ffi.web.dart to store the api instance, which I am worried whether it is too much constraint. Take Spring Boot for example - it prefers conventions over configuration, but at the same time, it does allow configuration, so anyone who needs customization can do it.

frb_codegen/src/generator/dart/mod.rs
fzyzcjy: Btw, with the suggestion below, maybe just do `let common_header = generate_common_header() + config.extra_headers;`

fzyzcjy:
Good job!

PR #1202, opened 2023-05-07
fix enum as struct field; fix errors in strict dynamic error

alanlzhang:
added `impl Default` for enum in rust

and `as int` for enum in dart

### DIFF
--- a/frb_codegen/src/main.rs
+++ b/frb_codegen/src/main.rs
@@ -49,9 +49,9 @@ mod tests {
         static ref LOGGER: () = init_logger(".", true).unwrap();
     }
 
-    #[cfg(windows)]
-    const DART: &str = "dart.bat";
-    #[cfg(not(windows))]
+    // #[cfg(windows)]
+    // const DART: &str = "dart.bat";
+    // #[cfg(not(windows))]
     const DART: &str = "dart";
 
     // VS Code runs in frb_codegen with "Run test" and flutter_rust_bridge with "Debug test" >_>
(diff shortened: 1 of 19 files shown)


fzyzcjy:
Good job! I will review after CI goes green :)

alanlzhang:
Finally CI goes green...

frb_codegen/src/main.rs
```diff
@@ -49,6 +49,11 @@ mod tests {
         static ref LOGGER: () = init_logger(".", true).unwrap();
     }
 
+    // #[cfg(windows)]
+    // const DART: &str = "dart.bat";
+    // #[cfg(not(windows))]
+    const DART: &str = "dart";
```
fzyzcjy, line 55: nit: hmm so it is the same as the old code?

fzyzcjy (commented):
(just briefly gone through it, will continue reviewing probably within a day)

fzyzcjy (approved):
Good fix! (will merge after all conversations are resolved)

frb_codegen/src/main.rs
alanlzhang: On my machine, I have to use `dart.bat` to run tests successfully. 

But on CI windows machine, `dart.bat` causes `not found` error.

So I have to comment out `dart.bat` to pass the CI.

Maybe flutter version matters.

fzyzcjy: Hmm which flutter are you using?

alanlzhang: `flutter doctor` :

image

PR #1206, opened 2023-05-08
fix: Bridge access for Rust opaques

Desdaemon:
## Changes

Fix code generated for Rust opaques when `--no-use-bridge-in-method` is active (referencing a non-existent `bridge`)

## Checklist

- [ ] CI is passing.

## Remark for PR creator

### DIFF
--- a/frb_example/pure_dart/dart/lib/ffi.io.dart
+++ b/frb_example/pure_dart/dart/lib/ffi.io.dart
@@ -1,8 +1,12 @@
 import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
 
+import 'bridge_definitions.dart';
 import 'bridge_generated.dart';
 
-FlutterRustBridgeExampleSingleBlockTestImpl initializeExternalLibrary(String path) =>
-    FlutterRustBridgeExampleSingleBlockTestImpl(
-      loadLibForDart(path),
-    );
+FlutterRustBridgeExampleSingleBlockTestImpl initializeExternalLibrary(
+  String path,
+) =>
+    _api = FlutterRustBridgeExampleSingleBlockTestImpl(loadLibForDart(path));
+
+FlutterRustBridgeExampleSingleBlockTest? _api;
+FlutterRustBridgeExampleSingleBlockTest get api => _api ?? (throw Exception('API not yet initialized'));
--- a/frb_example/pure_dart/dart/lib/ffi.web.dart
+++ b/frb_example/pure_dart/dart/lib/ffi.web.dart
@@ -1,3 +1,4 @@
+import 'bridge_definitions.dart';
 import 'bridge_generated.web.dart';
 export 'bridge_generated.web.dart';
 import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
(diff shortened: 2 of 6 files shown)


fzyzcjy (commented):
Good fix! However, I am a bit confused why there is no change at bridge_generated.rs etc. Maybe we need to add a test, otherwise this behavior is not tested and can be regressed easily in the future

fzyzcjy:
Good job! Quite busy now so may check as soon as possible :)

frb_example/pure_dart/rust/.flutter_rust_bridge.yml
```diff
@@ -8,3 +8,4 @@ c_output: [c_output_path/c_output.h]
 extra_c_output_path:
   - c_output_path_extra/
 # dump: []
+no_use_bridge_in_method: true
```
fzyzcjy, line 11: Not very sure, shall we enable this flag in this example? I guess yes, because this makes the API simpler to use (no need to carry a `bridge` instance here and there).

frb_example/pure_dart/dart/lib/ffi.web.dart
```diff
@@ -1,10 +1,18 @@
...
+    _api = FlutterRustBridgeExampleSingleBlockTestImpl.wasm(
       WasmModule.initialize(kind: const Modules.noModules(root: root)),
     );
+
+FlutterRustBridgeExampleSingleBlockTest? _api;
+FlutterRustBridgeExampleSingleBlockTest get api => _api ?? _bail('API not yet initialized');
+
+Never _bail(String msg) {
+  throw Exception(msg);
+}
```
fzyzcjy, line 18:
```suggestion
FlutterRustBridgeExampleSingleBlockTest get api => _api ?? (throw Exception('API not yet initialized'));
```

frb_example/pure_dart/dart/lib/ffi.io.dart
```diff
@@ -1,8 +1,16 @@
...
+  String path,
+) =>
+    _api = FlutterRustBridgeExampleSingleBlockTestImpl(loadLibForDart(path));
+
+FlutterRustBridgeExampleSingleBlockTest? _api;
+FlutterRustBridgeExampleSingleBlockTest get api => _api ?? _bail('API not yet initialized');
+
+Never _bail(String msg) {
+  throw Exception(msg);
+}
```
fzyzcjy, line 16: same?

fzyzcjy (approved):
Ready to merge :) Only nits

frb_example/pure_dart/rust/.flutter_rust_bridge.yml
Desdaemon: I'm on the fence here, personally I wouldn't enable the flag for pure_dart as it's not the default but up to you.

fzyzcjy: Yes that's a problem. 

An alternative may be, we somehow enhance our infrastructure to test more scenarios with various feature flags. IIRC our testing infra currently is quite limited about this - everyone who PR to make a change that only happens with *some* feature flag has to do hacky things like what you have done above...

Desdaemon: Fixtures may be the solution, but it would require a slightly more complicated PR. Luckily I should have more free time next month so I might look into it.

fzyzcjy: Fixtures look reasonable (feel free to sketch the plan before implementing so that we can discuss it a bit more), and looking forward to your PR next month!

As for this current PR - if we are going to add those thorough tests in the future, maybe we can keep this PR simple and follow your suggestion:

> personally I wouldn't enable the flag for pure_dart as it's not the default

fzyzcjy:
(will merge after CI goes green)

PR #1232, opened 2023-05-31
Fix compile when install the newest version

PonyPC:
cargo install flutter_rust_bridge_codegen v1.77.1
```
   Compiling flutter_rust_bridge_codegen v1.77.1
error[E0451]: field `config_path` of struct `Config` is private
   |
99 |         ..Default::default()
   |           ^^^^^^^^^^^^^^^^^^ field `config_path` is private

For more information about this error, try `rustc --explain E0451`.
error: could not compile `flutter_rust_bridge_codegen` due to previous error
warning: build failed, waiting for other jobs to finish...
error: failed to compile `flutter_rust_bridge_codegen v1.77.1`, intermediate artifacts can be found at
```

### DIFF
--- a/frb_codegen/src/commands.rs
+++ b/frb_codegen/src/commands.rs
@@ -77,25 +77,23 @@ fn cbindgen(
         "execute cbindgen rust_crate_dir={} c_output_path={}",
         rust_crate_dir, c_output_path
     );
-    let config = cbindgen::Config {
-        language: cbindgen::Language::C,
-        sys_includes: vec![
-            "stdbool.h".to_string(),
-            "stdint.h".to_string(),
-            "stdlib.h".to_string(),
-        ],
-        no_includes: true,
-        // copied from: dart-sdk/dart_api.h
-        // used to convert Dart_Handle to Object.
-        after_includes: Some("typedef struct _Dart_Handle* Dart_Handle;".to_owned()),
-        export: cbindgen::ExportConfig {
-            include: c_struct_names
-                .iter()
-                .map(|name| format!("\"{name}\""))
-                .collect(),
-            exclude: exclude_symbols,
-            ..Default::default()
-        },
+    let mut config = cbindgen::Config::default();
+    config.language = cbindgen::Language::C;
+    config.sys_includes = vec![
+        "stdbool.h".to_string(),
+        "stdint.h".to_string(),
+        "stdlib.h".to_string(),
+    ];
+    config.no_includes = true;
+    // copied from: dart-sdk/dart_api.h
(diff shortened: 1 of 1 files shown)


fzyzcjy:
LGTM after CI is green! More specifically, only need to worry about `Rust linter`, as the other two are flaky and known problems.

fzyzcjy (approved):

fzyzcjy:
Hmm, for this linter error, I guess we can suppress it there, since it is a false positive.

```
error: field assignment outside of initializer for an instance created with Default::default()
  --> frb_codegen/src/commands.rs:81:5
   |
81 |     config.language = cbindgen::Language::C;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: consider initializing the variable with `cbindgen::Config { language: cbindgen::Language::C, sys_includes: vec![
              "stdbool.h".to_string(),
              "stdint.h".to_string(),
              "stdlib.h".to_string(),
          ], no_includes: true, after_includes: Some("typedef struct _Dart_Handle* Dart_Handle;".to_owned()), export: cbindgen::ExportConfig {
              include: c_struct_names
                  .iter()
                  .map(|name| format!("\"{name}\""))
                  .collect(),
              exclude: exclude_symbols,
              ..Default::default()
          }, ..Default::default() }` and removing relevant reassignments
  --> frb_codegen/src/commands.rs:80:5
   |
80 |     let mut config = cbindgen::Config::default();
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#field_reassign_with_default
   = note: `-D clippy::field-reassign-with-default` implied by `-D warnings`
```

konsti219:
I think the better solution to this problem would be locking `cbindgen` at `0.24.5` while they work on fixing the broken crate version.

w-ensink:
Mozilla fixed it. Build it working again.

fzyzcjy:
> Mozilla fixed it. Build is working again.

So... shall I close the PR?

w-ensink:
> So... shall I close the PR?

I don't think you should merge this one, as it only makes the code more imperative and it doesn't really add anything. But it's totally up to you :-) 
The problem is solved either way.

fzyzcjy:
I checked a bit about the mozilla issue. Seems that I should keep it open and see what happens to the next breaking version of cbindgen. If that version still contains this breaking change, then we will need to eventually do this change I guess.

PR #1238, opened 2023-06-14
Translate tuples and enums

### LINKED ISSUE #1211: Use Dart 3 features for codegen
Generate appropriate code for records and sealed classes introduced in Dart 3. It can replace Freezed in cases where not all of its functionality is required, producing leaner code.

## Records (tuple)

Related: #329 #954

```rust
type A = (i32, String);
```

```dart
typedef A = (int, String);
```

## Sealed classes

```rust
pub enum Foo {
  Bar,
  Baz(String),
  Quux { foo: String }
}
```

```dart
sealed class Foo {
  const Foo();
}

class Bar extends Foo {
  const Bar();
}

class Baz extends Foo {
  final String field0;
  const Baz(this.field0);
}

class Quux extends Foo {
  final String foo;
  const Quux({required this.foo});
}
```

However, this is not a one-to-one replacement, since the current Freezed implementation renames enum variants from `Foo::Bar` to `Foo_Bar` to avoid name collisions, while idiomatic usage expects the names being unchanged, e.g. in switch expressions:

```dart
final foo = switch (Bar()) {
  Bar() => ..,
  Baz() => ..,
  Quux() => ..,
};
```

Desdaemon:
## Changes

Closes #1211
Blocked by ~~sunshine-protocol/allo-isolate#44~~

- `@freezed sealed class` syntax for pattern matching (~~gated behind `--dart3`~~ opt out with `--no-dart3`)
- Tuples to Dart records 

## Checklist

- [ ] CI is passing.

## Remark for PR creator


book/src/SUMMARY.md
fzyzcjy, line 1: nit: maybe add a line in https://cjycode.com/flutter_rust_bridge/feature/lang_simple.html

book/src/feature/lang_tuple.md
```diff
@@ -0,0 +1,14 @@
+# Tuples
+
+Introduced in Dart 3, [records](https://dart.dev/language/records) provide the equivalent of Rust's tuples.
```
fzyzcjy, line 3: Hey I realize https://github.com/fzyzcjy/flutter_rust_bridge/blob/b7ec2dabe9f6582c3d544f0e1239d3cba18b3dfb/frb_dart/pubspec.yaml#L6 we already require Dart 3 in our lib.

Thus, maybe we do not need to introduce this flag, and can make it the default?

(This can be done only if this flag will not introduce any breaking change)

frb_codegen/src/generator/dart/ty_record.rs
```diff
@@ -0,0 +1,72 @@
...
+                        value.safe_ident(),
+                        idx + 1
+                    )
+                } else {
+                    format!(
+                        "wireObj.field{idx} = api2wire_{}(apiObj.${});",
+                        value.safe_ident(),
+                        idx + 1
+                    )
+                }
```
fzyzcjy, line 47: hmm is it possible we unify the two cases? 

Or, does this pattern occur repeatedly? If so can we extract some common code?

frb_codegen/src/generator/rust/ty_record.rs
```diff
@@ -0,0 +1,83 @@
...
+                }}
+            }}
+
+            impl Default for {wire} {{
+                fn default() -> Self {{
+                    Self::new_with_null_ptr()
+                }}
+            }}"
+        )
+    }
```
fzyzcjy, line 82: nit: is it copy-paste-modified from somewhere in the codebase? If so we may be able to extract out some common code

fzyzcjy (commented):
Good job! Only nits

book/src/feature/lang_tuple.md
Desdaemon: It is a breaking change, since class modifiers like `sealed` were only introduced in Dart 3 and would cause a syntax error on previous versions. But if we're already on 3, it makes good sense to make it the default.

frb_codegen/src/generator/dart/ty_record.rs
Desdaemon: This was copied verbatim from IrStruct, so perhaps a solution for it may be desired as well.

fzyzcjy:
Feel free to ping me when CI is green!

Desdaemon:
@fzyzcjy I assume the MacOS integration test is still broken? Otherwise all seems to be fine.

fzyzcjy:
Yes I think so.

frb_codegen/src/generator/rust/ty_record.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    fn new_with_nullptr(&self, collector: &mut ExternFuncCollector) -> String {
+        TypeStructRefGenerator {
+            ir: self.ir.inner.clone(),
+            context: self.context.clone(),
+        }
+        .new_with_nullptr(collector)
+    }
+}
```
fzyzcjy, line 76: nit: simply extract `TypeStructRefGenerator { ...... }` into a function like, e.g.

```rust
impl TypeRecordGenerator {
    fn same_arg_struct_ref_generator(&self) -> TypeStructRefGenerator {
        TypeStructRefGenerator {
            ir: self.ir.inner.clone(),
            context: self.context.clone(),
        }

File: frb_codegen/src/generator/dart/ty_record.rs (hunk 2 above)
Comment (fzyzcjy) [reply frb_codegen/src/generator/dart/ty_record.rs]: 👀 Seems we forget to update this? It is still duplicated there

**Comment by fzyzcjy** [comment]
Ready to merge after a new nit + another unresolved nit above!

File: frb_codegen/src/generator/dart/ty_record.rs (hunk 2 above)
Comment (Desdaemon) [reply frb_codegen/src/generator/dart/ty_record.rs]: Record fields are accessed using the `${index + 1}` syntax compared to our homegrown `field$index`, so they aren't compatible and had to be brought over from ty_struct.

Comment (fzyzcjy) [reply frb_codegen/src/generator/dart/ty_record.rs]: Looking at https://github.com/fzyzcjy/flutter_rust_bridge/blob/b7ec2dabe9f6582c3d544f0e1239d3cba18b3dfb/frb_codegen/src/generator/dart/ty_struct.rs#L42-L56, what about this:

```rust
fn f(api_obj_field_name: &str) {
                    if field.ty.is_struct() {
                        format!(
                            "_api_fill_to_wire_{}(apiObj.{}, wireObj.{});",
                            field.ty.safe_ident(),
                            api_obj_field_name,
                            field.name.rust_style(),
                        )
                    } else {
                        format!(
                            "wireObj.{} = api2wire_{}(apiObj.{});",
                            field.name.rust_style(),
                            field.ty.safe_ident(),
                            api_obj_field_name,
                        )
                    }
}
```

where api_obj_field_name is `${idx+1}` here and `field.name.dart_style()` there

Desdaemon: I could merge that code, but personally I think that introduces a bit of unneeded coupling between the two modules and might end up being split again in the future, but up to you.

fzyzcjy: Hmm I am not sure. We already introduced coupling between Record and Struct in rust/ty_record.rs. So for consistency, I guess it is either acceptable to couple in dart/ty_record.rs, or we should also avoid that in the rust/ty_record.rs?

fzyzcjy:
Great job!

PR #1298, opened 2023-07-29
feat: reinforce error handling, reduce panics

Desdaemon:
## Changes

- Replace `anyhow::Error` with canonical error listings for library functions
  - Each module tries to manage its own error listing, rather than a single union of errors
- Allow `Err(anyhow!(..))?` as the new adhoc error syntax
  - Aim to reduce dependency on panicking as an error handling mechanism
- New flag `--keep-going` to delay error reporting to after codegen
- Use `anyhow::Context` where possible to make code concise and preserve backtraces

IR functions and methods are not included and should/may come in a separate PR.

The following public APIs have been changed:
- `ensure_tools_available` has opaque error type
- `dump::dump_multi` has opaque error type

## Checklist

- [ ] CI is passing.

## Remark for PR creator

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,6 @@ flutter_rust_bridge_macros = { path = "frb_macros", version = "1" }
 flutter_rust_bridge = { path = "frb_rust", default-features = false, version = "1" }
 flutter_rust_bridge_codegen = { path = "frb_codegen", version = "1" }
 
-[profile.release]
+[profile.release.package.flutter_rust_bridge_codegen]
 strip = "debuginfo"
-lto = "thin"
 debug = 1 # preserves some debug information
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,3 +32,4 @@ flutter_rust_bridge_codegen = { path = "frb_codegen", version = "1" }
 [profile.release]
 strip = "debuginfo"
 lto = "thin"
+debug = 1 # preserves some debug information
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ anyhow = "1.0.64"
 chrono = "0.4.23"
 lazy_static = "1.4.0"
 uuid = "1.1.2"
+thiserror = "1.0"
 
 flutter_rust_bridge_macros = { path = "frb_macros", version = "1" }
 flutter_rust_bridge = { path = "frb_rust", default-features = false, version = "1" }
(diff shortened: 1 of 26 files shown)


fzyzcjy:
Looks great - this will improve developer experience!

fzyzcjy:
(I will review probably within 24h)

frb_codegen/src/utils/dart_repository/dart_repo.rs
```diff
@@ -83,28 +83,26 @@ impl DartRepository {
...
+                DartToolchain::lock_filename()
+            )
         })?;
         let dependency = lock_file.packages.get(package);
         let version = match dependency {
             Some(dependency) => {
                 let pm = dependency.installed_in();
                 if pm.as_ref() != Some(&manager) {
-                    return Err(anyhow::Error::new(Error::InvalidDep {
+                    return Err(Error::InvalidDep {
```
fzyzcjy, line 101: nit: are we missing backtrace when migrating from anyhow error to thiserror?

seems we can add it https://stackoverflow.com/questions/61744018/how-do-i-print-a-backtrace-without-panicking-using-thiserror

I am not sure whether we should add the backtrace though - e.g. InvalidDep already tells you a lot of information, maybe they are already enough?

fzyzcjy (approved):
Good job and ready to merge after nits!

Cargo.toml
```diff
@@ -31,3 +32,4 @@ flutter_rust_bridge_codegen = { path = "frb_codegen", version = "1" }
 [profile.release]
 strip = "debuginfo"
 lto = "thin"
+debug = 1 # preserves some debug information
```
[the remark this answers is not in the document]
Desdaemon: I don't know if there is a way to set profiles for individual crates yet (Rust wouldn't allow it), if there is one I'll put it in for this PR.

frb_codegen/src/utils/dart_repository/dart_repo.rs
Desdaemon: Yeah, I've been thinking about what makes a backtrace useful in such cases. If the error is descriptive enough as to be self-diagnosable by the user then perhaps not. It is possible to add in your own backtraces though, so I can go back and add them in where needed.

fzyzcjy:
Btw, maybe we should add some doc about how errors should be handled? Or even better, if there could be some lints.

For example, when reviewing https://github.com/fzyzcjy/flutter_rust_bridge/pull/1320, I realize it is easy to introduce suboptimial ways of handling errors in new code. (btw that PR is just a draft and proof-of-concept so it is OK; but I am thinking about more general PRs in the future)

Desdaemon:
Maybe this solution is a bit extreme, but lints like [unwrap_in_result](https://rust-lang.github.io/rust-clippy/stable/index.html) can be used to discourage unannotated unwraps. I tried applying the lint to frb_codegen and there were only several cases so it could be done rather quickly, but long-term may not be sufficient.

fzyzcjy:
Looks interesting! I agree one lint rule may not be enough to catch all style problems.

Issue #1301, opened 2023-07-30
[Bug] bump proc-macro2


gutenfries:
### Describe the bug

I recently downloaded a project from git to start hacking away at it again and after some basic version upgrading, finding out what versions play well with each other, etc, this issue kept popping up.

FRB will build all the generated files without error, but when I attempt to run my code (via `flutter run -d linux`) it gives me the following error:

```console
flutter run -d linux  
Launching lib/main.dart on Linux in debug mode...
error[E0635]: unknown feature `proc_macro_span_shrink`
  --> /home/gutenfries/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proc-macro2-1.0.52/src/lib.rs:92:30
   |
92 |     feature(proc_macro_span, proc_macro_span_shrink)
   |                              ^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0635`.
error: could not compile `proc-macro2` (lib) due to previous error
error: failed to compile `corrosion-generator v0.1.0 (/home/gutenfries/Desktop/playback/client/build/linux/x64/debug/_deps/corrosion-src/generator)`, intermediate artifacts can be found at `/home/gutenfries/Desktop/playback/client/build/linux/x64/debug/_deps/corrosion-src/generator/target`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
CMake Error at /home/gutenfries/Desktop/playback/client/build/linux/x64/debug/_deps/corrosion-src/cmake/Corrosion.cmake:594 (message):
  Building CMake Generator for Corrosion - failed
Call Stack (most recent call first):
  /home/gutenfries/Desktop/playback/client/build/linux/x64/debug/_deps/corrosion-src/CMakeLists.txt:73 (include)


Building Linux application...                                           
Exception: Unable to generate build files

```

upon doing some more research (using google lol), I've found that this is an issue with `proc-macro`

Basically, the Rust team pushed some changes to prod, and the proc-macro2 folks haven't been able to make the necessary changes to their codebase at the time of the release.
(see https://github.com/rust-lang/rust/issues/113152 for more details)

This presents a breaking bug for FRB, as anyone with `>= nightly-2023-06-28` installed will run into the same (breaking) issue.

## Proposed fix(es):

the issue has recently (quite literally since I started typing this up lol) been resolved within `proc-macro2`, and versions `>=1.0.60` appropriately address this.
[...]

PR #1306, opened 2023-07-31
(chore): manually bump `proc-macro2` -> `1.0.66`

gutenfries:
# Changes:

Manually bumps `proc-macro2` to `1.0.66` to address issues with versions `<=1.0.65` concerning the nightly toolchain.

Fixes https://github.com/fzyzcjy/flutter_rust_bridge/issues/1301

## Checklist:

- [ ] Implement CI testing for changes

## Notes:

blindly upgrading cargo crates _MAY_ revert different proc-macro2 versions, causing unexpected behavior. be cautious when updating/installing cargo crates.


gutenfries:
@fzyzcjy 
CI is failing randomly, however, the same couple CI checks are also failing randomly on `main`, so I am not concerned about the changes in this PR.

fzyzcjy:
macos & web are indeed flaky now. a for ios, as long as one ios out of two passes it is ok

frb_example/pure_dart/rust/Cargo.lock
fzyzcjy, line 1: nits: What about modifying `Cargo.toml` if 1.0.55 is guaranteed to have bugs?

fzyzcjy, line 1: nit: If upgrading to 1.0.66, will our MSRV be affected? In other words, for people using older rust versions, will we have trouble? It would be great to automatically verify this. For example, we can modify the CI a little bit, using a `matrix` to run a bit of code with multiple rust versions.

fzyzcjy, line 1: nit: similarly, it would be great to reproduce the bug you are fixing in the CI, such that we are 100% sure it will not regress in the future.

fzyzcjy (commented):
Good job! Only a bit of nits

frb_example/pure_dart/rust/Cargo.lock
gutenfries: the MSRV SHOULD be unaffected.

The specific change that broke proc-macro2 was that rust depreciated an API in the nightly toolchain which the proc-macro2 did not have sufficiently ready.

If a user has an older version of rust (obv nightly) there will be no unintended compat issues. this is tested in proc-macro2, I believe.

Someone compiling with an _older_ version of rust will in a sense "have an extra API available, which they don't need", if that makes sense

gutenfries: I agree that this should be CI tested, I will work on pushing changes to this PR and let you know when they are ready for review.

gutenfries:
@fzyzcjy ready for review

frb_example/pure_dart/rust/Cargo.lock
fzyzcjy: You are right

fzyzcjy:
Well done! I will probably review within 24 hours

.github/workflows/ci.yaml
```diff
@@ -477,3 +477,49 @@ jobs:
...
+  # -------------- MSRV & specific version testing --------------
+  msrv:
+    name: MSRV & specific version testing
+    runs-on: ubuntu-latest # as of now, there is no real reason to test on multiple OSs
+    strategy:
+      matrix:
+       rust_version:
+         - nightly-2023-06-28   # pre https://github.com/rust-lang/rust/issues/113152
+         - nightly-2023-07-26   # post https://github.com/rust-lang/rust/issues/113152
+         - nightly              # latest nightly
```
fzyzcjy, line 490: it would be great to add

* latest stable
* old stable e.g. 1.64.0

we can exclude them for places where we need nighly (e.g. web?), but it is great to support stable explicitly.

fzyzcjy (approved):

.github/workflows/ci.yaml
fzyzcjy: Oops, I see you are just talking about *this* issue instead of MSRV, then yes, let me merge it

fzyzcjy:
@all-contributors please add @ gutenfries for code

fzyzcjy:
@all-contributors please add @gutenfries for code

Issue #1302, opened 2023-07-31
[Bug] flutter_rust_bridge_codegen  can't understand enums/struct defined by macros


JustSimplyKyle:
### Describe the bug

It failed to generate types defined by macros.

### Codegen logs with `RUST_LOG=debug` environment variable

```shell
2023/07/31 18:31:57 [DEBUG] configs=[Opts { rust_input_path: "/home/kyle/coding/era-connect/app/native/src/api.rs", dart_output_path: "/home/kyle/coding/era-connect/app/lib/api/gen/bridge_generated.dart", dart_decl_output_path: Some("/home/kyle/coding/era-connect/app/lib/api/gen/bridge_definitions.dart"), c_output_path: ["/home/kyle/coding/era-connect/app/macos/Runner/bridge_generated.h"], rust_crate_dir: "/home/kyle/coding/era-connect/app/native", rust_output_path: "/home/kyle/coding/era-connect/app/native/src/bridge_generated.rs", class_name: "Native", dart_format_line_length: 80, dart_enums_style: false, skip_add_mod_to_lib: false, llvm_path: ["/opt/homebrew/opt/llvm", "/usr/local/opt/llvm", "/usr/lib/llvm-9", "/usr/lib/llvm-10", "/usr/lib/llvm-11", "/usr/lib/llvm-12", "/usr/lib/llvm-13", "/usr/lib/llvm-14", "/usr/lib/", "/usr/lib64/", "C:/Program Files/llvm", "C:/msys64/mingw64"], llvm_compiler_opts: "", manifest_path: "/home/kyle/coding/era-connect/app/native/Cargo.toml", dart_root: Some("/home/kyle/coding/era-connect/app"), build_runner: true, block_index: BlockIndex(0), skip_deps_check: false, wasm_enabled: false, inline_rust: false, bridge_in_method: true, extra_headers: "", dart3: true }]
2023/07/31 18:31:57 [DEBUG] Trying to parse "/home/kyle/coding/era-connect/app/native/src/api.rs"
2023/07/31 18:31:57 [DEBUG] Trying to parse "/home/kyle/coding/era-connect/app/native/src/api/config/mod.rs"
2023/07/31 18:31:57 [DEBUG] Trying to parse "/home/kyle/coding/era-connect/app/native/src/api/config/config_loader.rs"
[...]

PR #1320, opened 2023-08-09
Understand macros

MnlPhlp:
## Changes

Some ideas/experiments for fixing #1302.
To get around the problem with expanding marker macros i just transform them to comments and back when reading. This way all the parsing/searching markers can stay the same.
The only tests that don't work yet are the non_final fields because those macros still get removed. Same for the default value macros. The same trick with commenting should work here.
We can't use the build scripts anymore for code-generation because they cant use cargo-expand, so we might need more changes in the tests. Code generation with just precommit works fine

## Checklist

- [ ] CI is passing.

## Remark for PR creator


fzyzcjy:
Good job! I will check ASAP (hopefully within a day)

MnlPhlp:
The handling of the macros is probably not ideal. I just wanted to see if the idea works before spending more time.
The macro should probably operate directly on the TokenStream instead of on strings. This was just easier to implement quickly for testing the idea.
But all existing tests work now 🎉

frb_example/pure_dart/rust/src/api.rs
```diff
@@ -1695,3 +1695,17 @@ pub fn test_tuple_2(value: Vec<(String, i32)>) {
 pub fn sync_return_mirror() -> SyncReturn<ApplicationSettings> {
     SyncReturn(external_lib::get_app_settings())
 }
+
+macro_rules! generate_struct {
```
fzyzcjy, line 1699: nit: would be great to test more, e.g. the marker macros you mentioned

(surely not now, but when we are sure this way does work! just want to add this sentence now in case I forget in the future)

justfile
```diff
@@ -44,6 +44,7 @@ dart_pub_get mode="default":
 # ============================ build & test ============================
 
 rust_build_and_test:
+    just _install_crate cargo-expand
```
fzyzcjy, line 47:
```suggestion
    just install_expand
```

frb_codegen/src/commands.rs
```diff
@@ -245,3 +250,82 @@ pub fn build_runner(dart_root: &str) -> Result {
...
 }
+
+lazy_static! {
+    static ref CACHE: Mutex<HashMap<String, String>> = Mutex::new(HashMap::new());
+}
+
+pub fn cargo_expand(path: &str, module: Option<&str>) -> String {
+    let mut cache = CACHE.lock().unwrap();
+    match cache.get(path) {
+        Some(expanded) => {
```
fzyzcjy, line 261: nit: https://stackoverflow.com/questions/28512394/how-to-lookup-from-and-insert-into-a-hashmap-efficiently

frb_codegen/src/utils/command_runner.rs
```diff
@@ -90,16 +89,10 @@ pub(crate) fn execute_command<'a>(
...
-    if log_enabled!(log::Level::Info) {
-        cmd.stdout(Stdio::inherit()).stderr(Stdio::inherit());
-    }
 
     if let Some(current_dir) = current_dir {
         cmd.current_dir(current_dir);
     }
-    if log_enabled!(log::Level::Info) {
-        cmd.stdout(Stdio::inherit()).stderr(Stdio::inherit());
-    }
```
fzyzcjy, line 102: 👀

frb_codegen/src/utils/command_runner.rs
```diff
@@ -112,14 +105,10 @@ pub(crate) fn execute_command<'a>(
 
     let stdout = String::from_utf8_lossy(&result.stdout);
     if result.status.success() {
-        debug!(
-            "command={:?} stdout={} stderr={}",
-            cmd,
-            stdout,
-            String::from_utf8_lossy(&result.stderr)
-        );
```
fzyzcjy, line 120: 👀  seems we should not remove logs, o/w hard to debug.

but for the cargo expand, we can remove logs for *it* (eg add a flag)

frb_codegen/src/utils/misc.rs
```diff
@@ -184,3 +185,18 @@ macro_rules! ir {
...
+
+pub fn read_rust_file(path: &PathBuf) -> String {
+    let module = match path.file_name().and_then(|e| e.to_str()).unwrap_or("") {
+        "mod.rs" => path
+            .parent()
+            .and_then(|p| p.file_name())
+            .and_then(|f| f.to_str()),
+        "lib.rs" => None,
+        "" => None,
+        _ => path.file_stem().and_then(|f| f.to_str()),
```
fzyzcjy, line 197: nit: what if api.rs is in a nested module? e.g. `src/hello/world/api.rs`

frb_example/pure_dart/rust/build.rs
```diff
@@ -1,24 +0,0 @@
-use lib_flutter_rust_bridge_codegen::{
```
fzyzcjy, line 1: nit: maybe we should at least remain one build.rs in our integration test system to ensure build.rs does not break (when there is no macros)

frb_macros/src/lib.rs
```diff
@@ -2,40 +2,29 @@
...
-        .collect()
-}
-
 /// Attribute to guide code generation.
 ///
 /// For what it can do, have a look at the documentation website.
 #[proc_macro_attribute]
-pub fn frb(_: TokenStream, item: TokenStream) -> TokenStream {
-    remove_marker_attr(item, "frb")
+pub fn frb(attribute: TokenStream, item: TokenStream) -> TokenStream {
```
fzyzcjy, line 9: (todo for @fzyzcjy - recheck this when the final implementation is done)

fzyzcjy (commented):
Good job! Briefly checked

fzyzcjy:
> The handling of the macros is probably not ideal. I just wanted to see if the idea works before spending more time.
The macro should probably operate directly on the TokenStream instead of on strings. This was just easier to implement quickly for testing the idea.
But all existing tests work now 🎉
At least locally. CI does not have cargo expand yet. I should probably also add a better error message that tells the user to install it in that case.

Sure! So I only briefly checked the code and skipped that part.

frb_codegen/src/commands.rs
MnlPhlp: Thanks for the hint. Makes that function much simpler.

frb_codegen/src/utils/command_runner.rs
MnlPhlp: These two blocks prevent the output to be captured in the output variable.
This means we can't use the command output if we redirect it to the console directly. It is still shown by the log messages

MnlPhlp: Without these changes the stdout variable was always empty

frb_codegen/src/utils/command_runner.rs
MnlPhlp: It is not a problem for cargo expand, it was just a lot of output. I changed it to only output in the case of an error, but we could change that back.

frb_codegen/src/utils/misc.rs
MnlPhlp: good point. I should split the path before the  "src" and create a module like hello::world::api in that case.
This also means adjustments in the extract_module function

frb_example/pure_dart/rust/build.rs
MnlPhlp: The one in pure_dart_mutli should still work i think. Not sure why i removed that.
I try to add that again

MnlPhlp: added the build script back in for pure_dart_multi

frb_example/pure_dart/rust/src/api.rs
MnlPhlp: you mean the markers (like mirror) inside macro generated Code? I can add that.

frb_codegen/src/utils/misc.rs
MnlPhlp: sub modules are correctly found now. Also added some tests

MnlPhlp:
@fzyzcjy  do you think the macros should be changed to use the TokenStream directly? If not i would leave them as they are, because i think it also is easier to understand/adjust and doesn't really matter for performance.
I fixed the other suggestions from the review, so please check again when you have time.
I am still looking into some failing ci runs but i don't really know what the problem is there.

fzyzcjy:
> @fzyzcjy do you think the macros should be changed to use the TokenStream directly? If not i would leave them as they are, because i think it also is easier to understand/adjust and doesn't really matter for performance.

I guess it would be great to use token stream, since it may be more robust?

By the way, it may be helpful to add a (unit) test case to that logic, then it is clear for everyone who reads the function to immediately understand the input/output!

frb_codegen/src/utils/command_runner.rs
fzyzcjy: I see. curious why we do this: 

        cmd.stdout(Stdio::inherit()).stderr(Stdio::inherit());

exact same code twice? Is it by mistake or deliberate

frb_codegen/src/utils/command_runner.rs
fzyzcjy: I agree. Maybe we can add a function parameter to enable/disable this logging?

frb_codegen/src/commands.rs
```diff
@@ -245,3 +250,111 @@ pub fn build_runner(dart_root: &str) -> Result {
...
+                .map(|n| n + searched.len())
+                .unwrap_or_default();
+            if start == 0 {
+                continue;
+            }
+            let end = expanded[start..]
+                .find(&format!("\n{}}}", " ".repeat(spaces)))
+                .map(|n| n + start)
+                .unwrap_or(expanded.len());
+            expanded = &expanded[start..end];
```
fzyzcjy, line 291: hmm shall we use such a mutable variable (`mut expanded: &str`)? it may be great to make it immutable when possible

frb_codegen/src/commands.rs
```diff
@@ -245,3 +250,111 @@ pub fn build_runner(dart_root: &str) -> Result {
...
+        .entry(String::from(dir))
+        .or_insert_with(|| run_cargo_expand(dir));
+    extract_module(expanded, module)
+}
+
+fn extract_module(mut expanded: &str, module: Option<String>) -> String {
+    if let Some(module) = module {
+        let mut spaces = 0;
+        for module in module.split("::") {
+            let searched = format!("mod {module} {{\n");
```
fzyzcjy, line 279: currently we use string manipulation. shall we use token stream manipulation directly? (the downstream already parses string to token stream, so maybe we can move the parsing up to here)

fzyzcjy (commented):
Looks good! Only two nits and I am ready to merge (after green CI)

Desdaemon:
Just want to note that frb_macros only removes the attributes at **compile time** and does not affect the AST input of codegen. What you could do is try to have codegen optimistically `syn::parse` the input tokens as an Item, and ignore the macro if parsing fails.

frb_example/pure_dart/rust/src/api.rs
MnlPhlp: added

frb_codegen/src/commands.rs
MnlPhlp: I don't think we can change that. In get_ir_file the source code is also used in text format.
```rust
    pub fn get_ir_file(&self) -> ParserResult<IrFile> {
        // info!("Phase: Parse source code to AST");
        let source_rust_content = read_rust_file(&PathBuf::from(&self.rust_input_path));
        let file_ast = syn::parse_file(&source_rust_content)?;

        // info!("Phase: Parse AST to IR");
        parser::parse(&source_rust_content, file_ast, &self.manifest_path)
    }

```

frb_codegen/src/commands.rs
MnlPhlp: i changed it to be immutable using fold. I am not sure if that is better. I think it's fine but it's a bit more complicated to read.

frb_codegen/src/utils/command_runner.rs
MnlPhlp: i think it's fine to just always activate. If you don't want the output you could change the logging level.

MnlPhlp:
updating the lock file worked for fixing the "Rust build and test" ci run on ubuntu. Is it ok to do that change?
Before changing it the ci run failed but it worked locally.

frb_codegen/src/utils/command_runner.rs
Desdaemon: I think twice is a mistake, but it definitely was meant for passing output and error messages to the user.

fzyzcjy:
(really busy now, will try to reply today)

fzyzcjy:
> updating the lock file worked for fixing the "Rust build and test" ci run on ubuntu. Is it ok to do that change?
Before changing it the ci run failed but it worked locally.

What command did you run to update it? If it is just a routine update, and it is compatible with old rust, then it is ok

frb_codegen/src/commands.rs
fzyzcjy: Hmm you are right. I do hope it can be ast instead of string manipulation :)

frb_codegen/src/commands.rs
fzyzcjy: Hmm, I am not sure. Originally I thought about making the param immutable and making a variable like `let mut interestd_part = expanded; ... change interested_part ...` on your old code. But anyway both approaches looks OK.

frb_codegen/src/utils/misc.rs
```diff
@@ -184,3 +185,55 @@ macro_rules! ir {
...
+fn get_dir_and_mod(path: &str) -> (String, Option<String>) {
+    #[cfg(windows)]
+    let path = &path.replace('\\', "/");
+    let src_index = path
+        .rfind("/src/")
+        .expect("src dir must exist in rust project");
+    let dir = &path[..src_index];
+    #[cfg(windows)]
+    let dir = dir.strip_prefix("//?/").unwrap_or(dir);
+    let module = &path[src_index + 5..];
```
fzyzcjy, line 204: nit: the magical "5" may be improved: what about `const NEEDLE = "/src/"`, and here only use `NEEDLE.len()`

fzyzcjy (approved):
Ready to merge after green CI :)

.github/workflows/ci.yaml
fzyzcjy, line 1: Oh one more nit I forget just now: It would be great to add doc into the book, telling users your good job! For simplicity, maybe we can mimic things like https://cjycode.com/flutter_rust_bridge/feature/worker_pool.html and add a 3.19 section about the macro problem this PR solves.

p.s. do not forget to update https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/book/src/SUMMARY.md

fzyzcjy (approved):

frb_codegen/src/utils/misc.rs
MnlPhlp: good point. Changed that

MnlPhlp:
CI is green except for that one MacOS Integration test that was already failing before

fzyzcjy:
@all-contributors please add @MnlPhlp for code,doc

fzyzcjy:
Well done!

PR #1364, opened 2023-09-17
Use delegate-attr to simplify code

upsuper:
## Changes

This PR replaces the use of `delegate` crate with `delegate-attr` crate to simplify the code while keeping the same functionality.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -491,14 +491,14 @@ dependencies = [
 ]
 
 [[package]]
-name = "delegate"
-version = "0.8.0"
+name = "delegate-attr"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


fzyzcjy (approved):
Thanks, LGTM!

Btw, curious whether this syntax works well or not in IDEs (e.g. CLion / VSCode) - will they show error that the empty body `{}` has no return values?

upsuper:
> Good job! I will wait for CI before merging (excluding those known to fail). Btw, I am curious whether this syntax works well or not in IDEs (e.g. CLion / VSCode) - will they show error that the empty body `{}` has no return values?

That's a good question. I'm using RustRover, and it doesn't seem to be a problem there at least. See the screenshot below:
Screenshot from 2023-09-17 19-10-09

I'm not very familiar with VSCode, so I have to say I'm not sure. My guess would be that if they can expand proc macro, they should be fine.

fzyzcjy:
@all-contributors please add @upsuper for code

PR #1388, opened 2023-10-09
Refactor `Vec<Option<_>>`

Desdaemon:
## Changes

(Internal) Refactors `Vec<Option<_>>` to work better with inner types, particularly delegates and primitives.

## Checklist

- [x] CI is passing.

## Remark for PR creator


fzyzcjy:
Looks great! Feel free to ping me after the PR is ready for review

Desdaemon:
@fzyzcjy All tests are passing except for some timed-outs, should be good for review now

frb_codegen/src/ir/ty_boxed.rs
```diff
@@ -72,7 +72,7 @@ impl IrTypeTrait for IrTypeBoxed {
 
     fn rust_wire_is_pointer(&self, target: Target) -> bool {
         !target.is_wasm()
-            || !self.inner.is_js_value() && !self.inner.is_array() && !self.inner.is_primitive()
+            || (!self.inner.is_js_value() && !self.inner.is_array() && !self.inner.is_primitive())
```
Desdaemon, line 75: This might be a mistake

frb_example/pure_dart/rust/c_output_path_extra/c_output.h
```diff
@@ -109,6 +109,33 @@ typedef struct wire_ExoticOptionals {
   struct wire_NewTypeInt *newtypeint;
 } wire_ExoticOptionals;
 
+typedef struct wire_list_opt_box_autoadd_i32 {
+  int32_t **ptr;
```
Desdaemon, line 113: Opportunity for optimization. When sizeof(T) is smaller than a pointer, it might make sense to remove the extra layer of indirection and track null indexes in another way (either a separate list/bitmask or niche optimization), but might not be straightforward as far as performance wins are concerned.

fzyzcjy: Looks reasonable! One way may be creating a separate issue, and gather feedback about how many people needs this performance optimization (i.e. they have bottleneck at this place), and the implement it after many people want it and it does improve speed measurably.

frb_dart/lib/src/helpers.dart
```diff
@@ -166,3 +166,12 @@ List<UuidValue> wire2apiUuids(Uint8List raw) {
...
 }
+
+List<T?> mapNonNull<T, I>(List<I?> items, T Function(I) mapper) {
+  final out = List<T?>.filled(items.length, null);
+  for (var i = 0; i < items.length; ++i) {
+    final item = items[i];
+    if (item != null) out[i] = mapper(item);
+  }
+  return out;
+}
```
fzyzcjy, line 177: Alternatively, we may write `items.map((item) => item == null ? null : mapper(item)).toList()`. I am not sure whether that one will be less performant though (i.e. can Dart compiler optimize it).

frb_codegen/src/ir/ty_optional_list.rs
```diff
@@ -0,0 +1,46 @@
+use crate::ir::*;
+use crate::target::Target;
+
+crate::ir! {
+    #[no_serde]
+    pub struct IrTypeOptionalList {
```
fzyzcjy, line 6: nit: it would be great to mention briefly why we implement this Vec-Option, instead of using the already existing Vec.

frb_codegen/src/generator/rust/ty_optional_list.rs
```diff
@@ -0,0 +1,52 @@
...
+        collector: &mut super::ExternFuncCollector,
+        block_index: crate::utils::misc::BlockIndex,
+    ) -> Acc<Option<String>> {
+        Acc {
+            io: Some(collector.generate(
+                &format!("new_{}_{}", self.ir.safe_ident(), block_index),
+                [("len: i32", "int")],
+                Some(&format!("*mut {}", self.ir.rust_wire_type(Target::Io))),
+                &format!(
+                    "let wrap = {} {{ ptr: support::new_leak_vec_ptr(core::ptr::null_mut(), len), len }};
```
fzyzcjy, line 40: nit: It seems that we are having some similarities with e.g. general_list. As usual, it it possible we somehow extract the common parts to reduce code duplication? (especially the "dangerous" part - unsafe leak ptr etc)

fzyzcjy (commented):
Well done! Only nits and I am ready to merge

frb_codegen/src/ir/ty_optional_list.rs
Desdaemon: Good call, I'll add some notes in the docs somewhere.

frb_dart/lib/src/helpers.dart
Desdaemon: This might require benchmarking to verify, but map() is known to be slow in transpiled JS (for-loops get JIT'd better) and AOT (indirect calls may not be inlined). Since this is reused many times, I don't mind being a bit more thorough with it.

frb_codegen/src/generator/rust/ty_optional_list.rs
Desdaemon: I considered reusing generate_list_allocate_func, but I found it is not very appropriate for what OptionalList is trying to do and would require adjusting the function anyways.

fzyzcjy (approved):

fzyzcjy:
(Will merge after CI is green!)

PR #1420, opened 2023-11-20
Fix `rustfmt::skip`

rogurotus:
Can't add `#[rustfmt::skip]` for `bridge_generated`
```
error[E0659]: `io` is ambiguous
    --> frb_example/pure_dart/rust/src/bridge_generated.rs:4149:9
     |
4149 | pub use io::*;
     |         ^^ ambiguous name
```


fzyzcjy:
Good job!

PR #1463, opened 2023-12-21
Fixes #1454 by setting the msrv

### LINKED ISSUE #1454: [Bug] cargo install 'flutter_rust_bridge_codegen@^2.0.0-dev.2' has compilation errors
### Describe the bug

First: @fzyzcjy Gratulations on the v2 release! Will upgrade and test these days :)

When running
`cargo install 'flutter_rust_bridge_codegen@^2.0.0-dev.2'`
the following errors appear (=> I am continuing with cargo binstall):

```
   Compiling flutter_rust_bridge_codegen v2.0.0-dev.2
error[E0446]: crate-private type `CodecModePack` in public interface
  --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/func.rs:18:5
   |
18 |     pub codec_mode_pack: CodecModePack,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak crate-private type
   |
  ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/generator/codec/structs.rs:13:1
   |
13 | pub(crate) struct CodecModePack {
   | ------------------------------- `CodecModePack` declared as crate-private

error[E0446]: crate-private type `lang::Lang` in public interface
  --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ident.rs:31:5
   |
31 |     pub fn style(&self, lang: &Lang) -> String {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak crate-private type
   |
  ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/generator/codec/sse/lang/mod.rs:11:1
   |
11 | pub(crate) enum Lang {
   | -------------------- `lang::Lang` declared as crate-private

error[E0446]: crate-private type `lang::Lang` in public interface
  --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/namespace.rs:94:5
   |
94 |     pub fn style(&self, lang: &Lang) -> String {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak crate-private type
   |
  ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/generator/codec/sse/lang/mod.rs:11:1
   |
11 | pub(crate) enum Lang {
   | -------------------- `lang::Lang` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/boxed.rs:12:5
    |
12  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
13  | |         &self,
14  | |         f: &mut F,
15  | |         ir_context: &impl IrContext,
16  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/dart_fn.rs:13:5
    |
13  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
14  | |         &self,
15  | |         f: &mut F,
16  | |         ir_context: &impl IrContext,
17  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/dart_opaque.rs:9:5
    |
9   | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
10  | |         &self,
11  | |         f: &mut F,
12  | |         ir_context: &impl IrContext,
13  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/delegate.rs:51:5
    |
51  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
52  | |         &self,
53  | |         f: &mut F,
54  | |         ir_context: &impl IrContext,
55  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/dynamic.rs:8:5
    |
8   | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
9   | |         &self,
10  | |         _f: &mut F,
11  | |         _ir_context: &impl IrContext,
12  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/enumeration.rs:48:5
    |
48  |     pub fn get<'a>(&self, file: &'a impl IrContext) -> &'a IrEnum {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 | pub(crate) trait IrContext {
    | -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/enumeration.rs:54:5
    |
54  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
55  | |         &self,
56  | |         f: &mut F,
57  | |         ir_context: &impl IrContext,
58  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/general_list.rs:10:5
    |
10  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
11  | |         &self,
12  | |         f: &mut F,
13  | |         ir_context: &impl IrContext,
14  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/optional.rs:12:5
    |
12  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
13  | |         &self,
14  | |         f: &mut F,
15  | |         ir_context: &impl IrContext,
16  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/ownership.rs:22:5
    |
22  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
23  | |         &self,
24  | |         f: &mut F,
25  | |         ir_context: &impl IrContext,
26  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/primitive.rs:25:5
    |
25  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
26  | |         &self,
27  | |         _f: &mut F,
28  | |         _ir_context: &impl IrContext,
29  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/primitive_list.rs:11:5
    |
11  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
12  | |         &self,
13  | |         f: &mut F,
14  | |         ir_context: &impl IrContext,
15  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/record.rs:14:5
    |
14  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
15  | |         &self,
16  | |         f: &mut F,
17  | |         ir_context: &impl IrContext,
18  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/rust_auto_opaque.rs:14:5
    |
14  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
15  | |         &self,
16  | |         f: &mut F,
17  | |         ir_context: &impl IrContext,
18  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/rust_opaque.rs:28:5
    |
28  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
29  | |         &self,
30  | |         f: &mut F,
31  | |         ir_context: &impl IrContext,
32  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/structure.rs:29:5
    |
29  |     pub fn get<'a>(&self, ir_context: &'a impl IrContext) -> &'a IrStruct {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 | pub(crate) trait IrContext {
    | -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/structure.rs:36:5
    |
36  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
37  | |         &self,
38  | |         f: &mut F,
39  | |         _ir_context: &impl IrContext,
40  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/unencodable.rs:26:5
    |
26  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
27  | |         &self,
28  | |         _f: &mut F,
29  | |         _ir_context: &impl IrContext,
30  | |     ) {
    | |_____^ can't leak crate-private trait
    |
   ::: /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:151:1
    |
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:52:5
    |
52  |     pub fn visit_types<F: FnMut(&IrType) -> bool>(&self, f: &mut F, ir_context: &impl IrContext) {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak crate-private trait
...
151 | pub(crate) trait IrContext {
    | -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:80:5
    |
80  | /     fn visit_children_types<F: FnMut(&IrType) -> bool>(
81  | |         &self,
82  | |         f: &mut F,
83  | |         ir_context: &impl IrContext,
84  | |     );
    | |______^ can't leak crate-private trait
...
151 |   pub(crate) trait IrContext {
    |   -------------------------- `IrContext` declared as crate-private

error[E0445]: crate-private trait `IrContext` in public interface
   --> /Users/patmuk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.0.0-dev.2/src/library/codegen/ir/ty/mod.rs:78:1
    |
78  | #[enum_dispatch]
    | ^^^^^^^^^^^^^^^^ can't leak crate-private trait
...
151 | pub(crate) trait IrContext {
    | -------------------------- `IrContext` declared as crate-private
    |
    = note: this error originates in the attribute macro `enum_dispatch` (in Nightly builds, run with -Z macro-backtrace for more info)

Some errors have detailed explanations: E0445, E0446.
For more information about an error, try `rustc --explain E0445`.
error: could not compile `flutter_rust_bridge_codegen` (lib) due to 24 previous errors
warning: build failed, waiting for other jobs to finish...
error: failed to compile `flutter_rust_bridge_codegen v2.0.0-dev.2`, intermediate artifacts can be found at `/var/folders/65/4zdp9whj28j122crpwzk7d9m0000gs/T/cargo-installM9o6jl`
```

```shell
failed before that step
```

### To Reproduce

_No response_

### Expected behavior

_No response_

### Generated binding code

_No response_

### OS

MacOS 14.2

v2.0.0-dev.2

### Flutter info

```shell
failed before that step
```

### Version of `clang++`

_No response_

### Version of `ffigen`

_No response_

### Additional context

_No response_

patmuk:
## Changes
- fixes #1454 by adding the MSRV to fab_codegen

## Checklist

- [x] New tests are added to ensure new features are working. End-to-end tests are usually in the `./frb_example/pure_dart` example, more specifically, `rust/src/api/whatever.rs` and `test/api/whatever_test.dart`. => test is in ci
- [x] CI is passing. (Operations are reproducible using corresponding `./frb_internal something` commands shown in CI.)
=> all passes, except "Benchmark" jobs. They did not pass before this changes either.
## Remark for PR creator
- The ci test installs "crate msrv", which verifies that the version set is correct. It can be used to find the correct version number as well - but that should be done locally, not in ci - as the MSRV usually doesn't change ofter. The ci fails only when the MSRV is not working anymore, by installing that version and running cargo check.
- This PR adds a MSRV to frb_codegen only, no other rust crates.
- There are two ways to specify the MSRV in the cargo.toml:
  - [platform.rust-version] (which I choose) and 
  - [platform.metadata.MSRV]

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -104,6 +104,21 @@ jobs:
       - run: ./frb_internal generate-website
 
   # ----------------------------------- codegen -----------------------------------
+  msrv:
+    name: 'Test :: FRB Codegen :: MSRV'    
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+
+    - name: Run cargo msrv
+      run: |
+        cargo install cargo-msrv
+        cargo msrv --path frb_codegen verify
 
   generate_run_frb_codegen_command_generate:
     name: 'Generate :: FRB Codegen :: Command Generate'
@@ -146,7 +161,7 @@ jobs:
           flutter-version: ${{ env.FRB_MAIN_FLUTTER_VERSION }}
           architecture: x64
       - uses: taiki-e/install-action@cargo-llvm-cov
-
+      
       # execute
       - run: ./frb_internal generate-run-frb-codegen-command-generate --set-exit-if-changed --package ${{ matrix.package }} --coverage
 
@@ -339,7 +354,6 @@ jobs:
             version: nightly
           - image: ubuntu-latest
             version: nightly-2023-12-16
-          # TODO(@fzyzcjy, @gutenfries): add MSRV here, as well as other locked versions as needed
 
     steps:
       # setup
@@ -367,8 +381,8 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           name: ${{ github.job }}--${{ matrix.info.image }}--${{ matrix.info.version }}--coverage
-          path: target/coverage
-
+          path: target/coverage  
+    
   test_dart_native:
     name: 'Test :: Dart :: Native'
     runs-on: ${{ matrix.image }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -347,8 +347,6 @@ jobs:
   test_rust:
     name: 'Test :: Rust'
     runs-on: ${{ matrix.info.image }}
-    outputs:
-        FRB_MSRV_RUST_VERSION: ${{ env.FRB_MSRV_RUST_VERSION }}
     strategy:
       fail-fast: false
       matrix:
@@ -366,9 +364,9 @@ jobs:
           - image: ubuntu-latest
             version: nightly-2023-12-16
           # tests the MSRV
-          # run on all rust packages, though only neede for frb_rust and frb_codegen
+          # run on all rust packages, though only needed for frb_rust and frb_codegen
           - image: ubuntu-latest
-          # update this, if a later MSVR is needed
+          # update this, if a later MSRV is needed
             version: 1.74.0
 
     steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -365,7 +365,7 @@ jobs:
           - image: ubuntu-latest
             version: nightly-2023-12-16
           - image: ubuntu-latest
-            version: ${{ env.FRB_MSRV_RUST_VERSION }}
+            version: $env.FRB_MSRV_RUST_VERSION
 
     steps:
       # setup
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -348,6 +348,8 @@ jobs:
   test_rust:
     name: 'Test :: Rust'
     runs-on: ${{ matrix.info.image }}
+    outputs:
+        FRB_MSRV_RUST_VERSION: ${{ env.FRB_MSRV_RUST_VERSION }}
     strategy:
       fail-fast: false
       matrix:
@@ -365,7 +367,7 @@ jobs:
           - image: ubuntu-latest
             version: nightly-2023-12-16
           - image: ubuntu-latest
-            version: $env.FRB_MSRV_RUST_VERSION
+            version: ${{ needs.test_rust.outputs.FRB_MSRV_RUST_VERSION }}
 
     steps:
       # setup
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -369,7 +369,7 @@ jobs:
           # run on all rust packages, though only neede for frb_rust and frb_codegen
           - image: ubuntu-latest
           # update this, if a later MSVR is needed
-            version: 1.70.0
+            version: 1.74.0
 
     steps:
       # setup
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,6 @@ on:
 env:
   CARGO_TERM_COLOR: always
   FRB_MAIN_RUST_VERSION: 1.74.0
-  FRB_MSRV_RUST_VERSION: 1.68.0
   FRB_MAIN_DART_VERSION: 3.2.0
   FRB_MAIN_FLUTTER_VERSION: 3.16.0
 
@@ -366,8 +365,11 @@ jobs:
             version: nightly
           - image: ubuntu-latest
             version: nightly-2023-12-16
+          # tests the MSRV
+          # run on all rust packages, though only neede for frb_rust and frb_codegen
           - image: ubuntu-latest
-            version: ${{ needs.test_rust.outputs.FRB_MSRV_RUST_VERSION }}
+          # update this, if a later MSVR is needed
+            version: 1.68.0
 
     steps:
       # setup
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -369,7 +369,7 @@ jobs:
           # run on all rust packages, though only neede for frb_rust and frb_codegen
           - image: ubuntu-latest
           # update this, if a later MSVR is needed
-            version: 1.68.0
+            version: 1.70.0
 
     steps:
       # setup
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,6 +10,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   FRB_MAIN_RUST_VERSION: 1.74.0
+  FRB_MSRV_RUST_VERSION: 1.68.0
   FRB_MAIN_DART_VERSION: 3.2.0
   FRB_MAIN_FLUTTER_VERSION: 3.16.0
 
@@ -104,21 +105,30 @@ jobs:
       - run: ./frb_internal generate-website
 
   # ----------------------------------- codegen -----------------------------------
-  msrv:
-    name: 'Test :: FRB Codegen :: MSRV'    
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v2
+  # instead of running this explicit test for frb_codegen and frb_rust only, we run the test for
+  # all rust packages in the job "test_rust". While this would not be needed for these other packages,
+  # we avoid a ci dependency to cargo-msrv.
+  # However, the code is left commented here in case this decission changes.
+  # msrv:
+  #   name: 'Test :: FRB Codegen :: MSRV'    
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #   - uses: actions/checkout@v2
 
-    - name: Install Rust
-      uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
+  #   - name: Install Rust
+  #     uses: actions-rs/toolchain@v1
+  #     with:
+  #       toolchain: stable
 
-    - name: Run cargo msrv
-      run: |
-        cargo install cargo-msrv
-        cargo msrv --path frb_codegen verify
+  #   - name: Verify MSRV for frb_codegen
+  #     run: |
+  #       cargo install cargo-msrv
+  #       cargo msrv --path frb_codegen verify
+
+  #   - name: Verify MSRV for frb_rust
+  #     run: |
+  #       cargo install cargo-msrv
+  #       cargo msrv --path frb_rust verify
 
   generate_run_frb_codegen_command_generate:
     name: 'Generate :: FRB Codegen :: Command Generate'
@@ -354,6 +364,8 @@ jobs:
             version: nightly
           - image: ubuntu-latest
             version: nightly-2023-12-16
+          - image: ubuntu-latest
+            version: ${{ env.FRB_MSRV_RUST_VERSION }}
 
     steps:
       # setup
--- a/frb_codegen/Cargo.toml
+++ b/frb_codegen/Cargo.toml
@@ -7,7 +7,7 @@ license.workspace = true
 name = "flutter_rust_bridge_codegen"
 repository.workspace = true
 version.workspace = true
-rust-version = "1.72.0"
+rust-version = "1.74.0"
 
 [lib]
 name = "lib_flutter_rust_bridge_codegen"
--- a/frb_codegen/Cargo.toml
+++ b/frb_codegen/Cargo.toml
@@ -2,12 +2,12 @@
 categories.workspace = true
 description.workspace = true
 edition.workspace = true
+rust-version = "1.73.0"
 keywords.workspace = true
 license.workspace = true
 name = "flutter_rust_bridge_codegen"
 repository.workspace = true
 version.workspace = true
-rust-version = "1.74.0"
 
 [lib]
 name = "lib_flutter_rust_bridge_codegen"
--- a/frb_codegen/Cargo.toml
+++ b/frb_codegen/Cargo.toml
@@ -7,6 +7,7 @@ license.workspace = true
 name = "flutter_rust_bridge_codegen"
 repository.workspace = true
 version.workspace = true
+rust-version = "1.72.0"
 
 [lib]
 name = "lib_flutter_rust_bridge_codegen"
--- a/frb_codegen/Cargo.toml
+++ b/frb_codegen/Cargo.toml
@@ -2,7 +2,7 @@
 categories.workspace = true
 description.workspace = true
 edition.workspace = true
-rust-version = "1.73.0"
+rust-version = "1.74.0"
 keywords.workspace = true
 license.workspace = true
 name = "flutter_rust_bridge_codegen"
--- a/frb_rust/Cargo.toml
+++ b/frb_rust/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "flutter_rust_bridge"
 edition.workspace = true
-rust-version = "1.69.0"
+rust-version = "1.70.0"
 version.workspace = true
 description.workspace = true
 license.workspace = true
--- a/frb_rust/Cargo.toml
+++ b/frb_rust/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "flutter_rust_bridge"
 edition.workspace = true
+rust-version = "1.69.0"
 version.workspace = true
 description.workspace = true
 license.workspace = true


.github/workflows/ci.yaml
```diff
@@ -104,6 +104,21 @@ jobs:
       - run: ./frb_internal generate-website
 
   # ----------------------------------- codegen -----------------------------------
+  msrv:
```
fzyzcjy, line 107: nit: just add a few lines below https://github.com/fzyzcjy/flutter_rust_bridge/blob/13c099fe196622f11bf0216affe00d086b867b7b/.github/workflows/ci.yaml#L331-L342

For example

```
          - image: ubuntu-latest
            version: 1.70 # or whatever version you like
```

patmuk: I can do that - but I don't understand completely why ...
cargo MSRV is installing the version specified in frb_codegen/cargo.toml itself, and runs cargo check afterwards. 
If we would specify a specific version in the image, it would get overwritten anyways - and that is good, as we don't need to maintain the rust version on two places (cargo.toml and ci).

Or did you mean something else?

fzyzcjy: Ok I realize we are talking two different things 🤣

* https://github.com/foresterre/cargo-msrv: Find the minimum supported Rust version (MSRV) for your project. Only need to run once locally to find it. 
* MSRV = "minimal supported rust version", e.g. https://internals.rust-lang.org/t/rust-msrv-policy-and-linux-distros/17074. If I understand correctly (correct me if I am wrong!), for example, MSRV=1.60 means "anyone with rust version >=1.60 can happily use my crate as a dependency".

Thus, Suppose we want MSRV=1.70 (surely we can aim lower). Then it seems that we can add a line as suggested above. Then, if the CI fails, it means anyone using rust 1.70 will fail to use our lib, and we should fix errors until ci is happy.

patmuk:
The [ci](https://github.com/fzyzcjy/flutter_rust_bridge/actions/workflows/ci.yaml) build has some issues I would need some help with:

1. The job [Test :: Flutter :: Native:: iOS](https://github.com/fzyzcjy/flutter_rust_bridge/pull/1463#logs) fails! 
The error is
```
[   +7 ms] executing: xcrun simctl spawn 90F52CAA-A5C8-4946-8432-DF0D4EF688AC log stream --style json --predicate eventType = logEvent AND processImagePath ENDSWITH "Runner" AND (senderImagePath ENDSWITH "/Flutter" OR senderImagePath ENDSWITH "/libswiftCore.dylib" OR processImageUUID == senderImageUUID) AND NOT(eventMessage CONTAINS ": could not find icon for representation -> com.apple.") AND NOT(eventMessage BEGINSWITH "assertion failed: ") AND NOT(eventMessage CONTAINS " libxpc.dylib ")
```
I don't understand how the changes lead to that - probably rust 1.74.0 is not the MSRV, but we need a higher number? Which are you using as the default in ci?

2. The ci runs take much longer now! 53m! Though other (successful) runs take around 40 min - maybe that is a normal variation?

3. I introduced a few warnings we should get rid of. Probably I am using a wrong settings for
``` 
    - uses: actions/checkout@v2
    - name: Install Rust
      uses: actions-rs/toolchain@v1
```
?

The warnings are:
```
[Test :: FRB Codegen :: MSRV](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/7282632799/job/19845306520)
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions-rs/toolchain@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
```
```
[Test :: FRB Codegen :: MSRV](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/7282632799/job/19845306520#step:3:30)
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```
(they are repeated a few times)

patmuk:
oh, 1. could be because of https://github.com/fzyzcjy/flutter_rust_bridge/issues/1464. Can you maybe verify that it runs locally? I haven't set up docker, etc.

fzyzcjy:
> 1. could be because of https://github.com/fzyzcjy/flutter_rust_bridge/issues/1464.

Usually, if one of the two ios test fails, it is just flaky, and ignore it. Also you can "bump" the CI by commiting some nonsense (e.g. a space change)

> The ci runs take much longer now! 53m! Though other (successful) runs take around 40 min - maybe that is a normal variation?

No worries, run more as you progress, and we will see whether it is just a variation

> The following actions uses node12

Again no worries until the last step. The existing CI has some warning as well due to 3rd party actions

.github/workflows/ci.yaml
patmuk: Hmm ... I don't understand what adding the suggested line does. I am ok adding it - but I am either understanding it wrong, or it is not needed :)
-> This would mark the ubuntu-latest image to have rust 1.70.0 installed? Probably not really installing that (as for that the dockerfile would need to be changed), but marking it for the reader of the ci-file?

MSRV=1.60 means exactly what you said.
But cargo-msrv does both: Using bisect to find the minimal rust version, by installing multiple rust versions -> this I did locally only.  
Or verifying that a set version is (still) correct, by installing only that rust version and executing cargo check. -> (only) this is done in ci.

The line 'rust-version' in frb_codegen/cargo.toml is hindering anyone to cargo install frb_codegen with a lower rust version. Cargo-msrv does nothing here.

The test in the ci with cargo msrv only checks if changes in frb_codegen need a higher rust version.

For example: If a change would require 1.75.0, instead of the version 1.74.0 set in cargo.toml, the ci run would fail.
But a user with rust 1.74.0 could still run cargo install frb_codegen, which would fail at compiling, but not tell that the MSRV is higher.  
Only after manually updating the cargo.toml to rust-version-1.75.0 the ci would be green again, and that user would get the error that his rust version is too old for cargo install frb_codegen.

Sorry for being so verbose - I am sure you got almost all already :)
So, what is your suggested line in ci.yaml doing?

patmuk:
Ok, super - will do so!

.github/workflows/ci.yaml
fzyzcjy: > -> This would mark the ubuntu-latest image to have rust 1.70.0 installed? Probably not really installing that (as for that the dockerfile would need to be changed), but marking it for the reader of the ci-file?

It means, IMHO, add a new test (so called "matrix") with condition rust 1.70 (while existing tests have stable, nightly, etc)
For more info, maybe refer to github actions' documentations. 

> Or verifying that a set version is (still) correct, by installing only that rust version and executing cargo check. -> (only) this is done in ci.

This looks interesting! However, if I understand correctly, using this part of cargo-msrv seems to have the same effect as adding the two lines suggested above. If so, we can choose the simpler method to avoid future maintainence burden :) But cargo-msrv for local testing is quite good.

> The line 'rust-version' in frb_codegen/cargo.toml is hindering anyone to cargo install frb_codegen with a lower rust version. Cargo-msrv does nothing here.

This LGTM!

patmuk: > It means, IMHO, add a new test (so called "matrix") with condition rust 1.70 (while existing tests have stable, nightly, etc) For more info, maybe refer to github actions' documentations.
> 
Ah, yes, I see that now :)
> > Or verifying that a set version is (still) correct, by installing only that rust version and executing cargo check. -> (only) this is done in ci.
> 
> This looks interesting! However, if I understand correctly, using this part of cargo-msrv seems to have the same effect as adding the two lines suggested above. If so, we can choose the simpler method to avoid future maintainence burden :) But cargo-msrv for local testing is quite good.

Yes - if 
```
      # execute
      - run: ./frb_internal test-rust --coverage`
```
is ONLY executing tests in frb_codegen (not easy to figure out). If it runs other tests as well we would run them on that version as well.

I think your suggestion works - but maintenance would get worse (?): I you mean by "more maintenance" that we depend on cargo-msrv, then yes, your alternative would remove that dependency.

But, if I am not wrong, that more than just frb_codegen tests are run here, then:  
One could install frb_codegen with binstall - and if the generated crates have a lower MSRV, all would be ok for that user. But the ci would fail here.  

Thus, we fail in a situation where everything would be ok, and the ci takes more time by running this scenario as well. And maybe a user needs to use an older rust toolchain.

However, `test-rust` is limited to frb_codegen only, I agree with your suggestion!
If not - well, my scenario is an unlikely edge-case. So, if my assumptions are correct, but you would still prefer independence from cargo-msrv over the drawbacks mentioned, I will implement your suggestion anyways :)

fzyzcjy: > is ONLY executing tests in frb_codegen (not easy to figure out). If it runs other tests as well we would run them on that version as well.

Runs test for all rust packages inside this git repository indeed. But it looks good if we ensure everything has the same MSRV, which simplifies things.

In addition, not only frb_codegen, but also the MSRV for frb_rust (the runtime library) is quite important. We surely do not want a low frb_codegen MSRV, while super high frb_rust MSRV!

patmuk: > Runs test for all rust packages inside this git repository indeed. But it looks good if we ensure everything has the same MSRV, which simplifies things.
True - I can look into adding the rust-version to all tomls, where they are generated.
> 
> In addition, not only frb_codegen, but also the MSRV for frb_rust (the runtime library) is quite important. We surely do not want a low frb_codegen MSRV, while super high frb_rust MSRV!
But the other way around might be desired: frb_codegen needs 1.74.0. Let's assume the rest is fine with 1.60.0. A user can get around with not having 1.74.0 by `cargo binstall frb_codegen`. The rest needs to be compiled, so anything below 1.60.0 would not work, but 1.60.0 would be fine. Maybe he has a reason for not upgrading to 1.74.0 (any of his dependencies not working with that version). I just want to illustrate this point so the tradeoff is clear.

So, how I see it:
- Variant A: cargo msrv checks the MSRV of frb_codegen only
- Variant B: all rust packages in this repo are run against a specific rust version

Contra-points:
- Variant A: 
  - ci depends on cargo-msrvg
- Variant B:
  - all rust packages of this repo need have the same MSRV, which is the highest among them. (Actually: All compiled packages vs binstallable (which is frb_codegen only)
    - users are potentially forced to use a higher rust version, if the coiled packages can have a lower MSRV than the binstallable. 
  - ci takes longer, as all packages are tested in the new image
  - maintenance is higher, as the MSRV needs to be updated in two places: cargo.toml and the ci image definition

Do you see anything else or disagree with that? I think variant A is better - but I will implement whatever you think is best. After all, both variant's contra points are not very strong, so it doesn't matter much - and it is your project :)

fzyzcjy: > ci takes longer, as all packages are tested in the new image

If we are worried, maybe we can add a flag to the script to enable only one package vs all packages.
But generally speaking, from my observations, CI for *this* part is quite fast, at least much faster than ios tests etc.

So are we really having very different MSRVs here? Our code is quite standard, so I guess indeed we will have a pretty good MSRV such that there is no worry at all about "frb_codegen different MSRV than frb_rust".

patmuk: > If we are worried, maybe we can add a flag to the script to enable only one package vs all packages. But generally speaking, from my observations, CI for _this_ part is quite fast, at least much faster than ios tests etc.

Sounds good!

> So are we really having very different MSRVs here? Our code is quite standard, so I guess indeed we will have a pretty good MSRV :) such that there is no worry at all about "frb_codegen different MSRV than frb_rust".

I think at the moment we have that: frb_codegen needs 1.74.0 - that was the initial issue.
The generated code and runtime is fine with a lower MSRV (I checked for `dart_minimal`, which is fine with 1.72.1, and frb_rust, which is fine with  1.70.0.).

I am most concerned about the additional maintenance needed in the ci file (though that is minimal admittedly):  
If  the ci fails because a higher MSRV is needed, only the cargo.toml has to be updated.  
In Variant B the toml and the ci.yaml need both an update. Not a big deal - but some additional thing to keep in mind.

fzyzcjy: > Not a big deal - but some additional thing to keep in mind.

Yes I agree it is not a huge problem :)

patmuk: Sounds like you prefer Variant B :) Good, will change this PR later.

I will set the MSRV in frb_rust as well.  
Setting it in the examples requires a bit more effort to make that properly - as this should be done through code generation.  
But I think this is not needed - as a user needs to have the MSRV for frb_rust anyways, and I assume there is no way that the created/integrated project would have something that requires a higher MSRV.
So, I will skip that, ok?

fzyzcjy: No need for examples! Since they are just examples :)

patmuk: It turns out that the changes are not that trivial ... 

1. setting the MSRV in ci.yaml as a variable is challenging - not sure if it does the trick in the end (see the diff).
2. the CI fails, as there are two identical jobs now (`Test :: Rust :: ubuntu-latest`). Logs like the version provided is not picked up.
3. The ci should fail, as I intentionally set the MSRV under the one specified - but it doesn't. Probably the version is indeed not picked up correctly.

I don't really want to hardcode the version without a variable, so one understands this is for MSRV. Will do so now, just to see if that is the only issue. But if this takes some more digging into to fix I prefer to save some time and revert back to Variant A.

patmuk: It works now :) Please look into the code. The MSRV is hardcoded, but I commented it so that it should be clear what to do if it fails in the future.
I left the implementation for Variant A inside as well - but commented.
I know, the code is in git as well - but I bet that just in a few weeks it would be hard to digg for that.

I ran some test ci runs with intentionally wrong version numbers - all works as expected.

Please let me know if I can do anything else regarding this PR :)

fzyzcjy:
Oh please ping me whenever I do not reply within a day! I did not see your reply above until now (when I open this page to refer to another issue).

.github/workflows/ci.yaml
```diff
@@ -323,6 +347,8 @@ jobs:
   test_rust:
     name: 'Test :: Rust'
     runs-on: ${{ matrix.info.image }}
+    outputs:
+        FRB_MSRV_RUST_VERSION: ${{ env.FRB_MSRV_RUST_VERSION }}
```
fzyzcjy, line 351: (this seems to be redundant and needs to be removed)

.github/workflows/ci.yaml
```diff
@@ -339,7 +365,11 @@ jobs:
             version: nightly
           - image: ubuntu-latest
             version: nightly-2023-12-16
-          # TODO(@fzyzcjy, @gutenfries): add MSRV here, as well as other locked versions as needed
+          # tests the MSRV
+          # run on all rust packages, though only neede for frb_rust and frb_codegen
+          - image: ubuntu-latest
+          # update this, if a later MSVR is needed
```
fzyzcjy, line 371:
```suggestion
          # update this, if a later MSRV is needed
```

.github/workflows/ci.yaml
```diff
@@ -339,7 +365,11 @@ jobs:
             version: nightly
           - image: ubuntu-latest
             version: nightly-2023-12-16
-          # TODO(@fzyzcjy, @gutenfries): add MSRV here, as well as other locked versions as needed
+          # tests the MSRV
+          # run on all rust packages, though only neede for frb_rust and frb_codegen
```
fzyzcjy, line 369:
```suggestion
          # run on all rust packages, though only needed for frb_rust and frb_codegen
```

frb_codegen/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 categories.workspace = true
 description.workspace = true
 edition.workspace = true
+rust-version = "1.74.0"
```
fzyzcjy, line 5: It may be great if we can lower the MSRV by fixing some trivial compile errors ;)

fzyzcjy (commented):
Good job!

patmuk:
> Oh please ping me whenever I do not reply within a day! I did not see your reply above until now (when I open this page to refer to another issue).

Surely! No worries, I thought you might have a Christmas break (if you celebrate it).

.github/workflows/ci.yaml
patmuk: Oh yes! This is a left-over of trying to define the variable on top of the CI file - removed it now.

frb_codegen/Cargo.toml
patmuk: Hmmm - I personally don't mind, as I like to use the latest version.  
But I suggest opening another issue for that. In #1454 you can get an impression of what kind of changes would have to be made (that was with rust 1.71., the compiler complained about `pup (crate)` mostly). 
I assume that would require a redesign of the modularization.  
IMHO: Not worth it, would not do that unless somebody complains.

I got to the MSRV of 1.74.0 by using `cargo msrv`, which performs a bisect search. So it won't compile to anything lower without changes to the code.
I would leave it as it, and only of somebody raises an issue needing a lower version work on specifically supporting from that version on.

fzyzcjy:
> Surely! No worries, I thought you might have a Christmas break (if you celebrate it).

No, as you can see from my github profile, I am on github 365 days per year ;)

frb_codegen/Cargo.toml
fzyzcjy: > I suggest opening another issue for that

Sure! Then let's keep this PR simple and atomic.

fzyzcjy:
@all-contributors please add @patmuk for code

PR #1491, opened 2023-12-25
Fix grammar in advantages.svg

adsouza:
The correct noun form for powerful is actually just power, which is the root. Similarly, the noun form of simple is simplicity.

## Changes

There's no PR for this because it's merely a minor edit to the text of an image.

### DIFF
--- a/website/misc/advantages.svg
+++ b/website/misc/advantages.svg
@@ -7,7 +7,7 @@
    viewBox="0 0 100 60.999999"
    version="1.1"
    id="svg5"
-   inkscape:version="1.3.2 (091e20e, 2023-11-25)"
+   inkscape:version="1.2.1 (9c6d41e4, 2022-07-14)"
    sodipodi:docname="advantages.svg"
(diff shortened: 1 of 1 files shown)


fzyzcjy:
Thank you for the change! Totally agree it should be simplicity, not simpleness.

As for powerfulness, I am not very sure: `power` (https://dictionary.cambridge.org/dictionary/english/power) seems to commonly refer to "ability to control people and events" or "the amount of political control a person or group has in a country", while I definitely do not want to control anyone! So I choose the less-common powerfulness.

adsouza:
I don't want to argue over this but I'll leave you with one suggestion: take a look at the results of https://www.google.com/search?q=%22expressive+power%22 & see if that changes your mind at all.

fzyzcjy:
Thank you! I know power has that meaning (at least dict says so) :) I am just worried if I use "power" without things like "expressive"... Anyway I will also consult my friend focusing on this area as well.

fzyzcjy:
@all-contributors please add @adsouza for doc

PR #1500, opened 2023-12-26
refactor: remove double-allocation for enum CST decode

Desdaemon:
## Changes

(CST) Remove `*mut Enum_Variant` and inflate functions in favor of `MaybeUninit`. This reduces 1-2 allocations per enum instantiation and should reduce binary size somewhat.

## Checklist

## Remark for PR creator

### DIFF
--- a/frb_codegen/src/library/codegen/generator/wire/rust/spec_generator/codec/cst/decoder/ty/enumeration.rs
+++ b/frb_codegen/src/library/codegen/generator/wire/rust/spec_generator/codec/cst/decoder/ty/enumeration.rs
@@ -1,15 +1,17 @@
 use crate::codegen::generator::acc::Acc;
 use crate::codegen::generator::misc::target::{Target, TargetOrCommon};
-use crate::codegen::generator::wire::rust::spec_generator::extern_func::{ExternClass, ExternClassMode, ExternFunc};
+use crate::codegen::generator::wire::rust::spec_generator::extern_func::{ExternClass, ExternClassMode};
 use crate::codegen::generator::wire::rust::spec_generator::output_code::WireRustOutputCode;
 use crate::codegen::generator::wire::rust::spec_generator::codec::cst::base::*;
 use crate::codegen::generator::wire::rust::spec_generator::codec::cst::decoder::ty::WireRustCodecCstGeneratorDecoderTrait;
-use crate::codegen::ir::field::IrField;
+use crate::codegen::ir::ident::IrIdent;
 use crate::codegen::ir::ty::enumeration::{IrEnum, IrEnumMode, IrVariant, IrVariantKind};
-use crate::codegen::ir::ty::IrType;
+use crate::codegen::ir::ty::structure::IrStruct;
 use itertools::Itertools;
 use crate::codegen::generator::wire::rust::spec_generator::codec::cst::decoder::impl_new_with_nullptr::generate_impl_new_with_nullptr_code_block;
 use crate::codegen::generator::wire::rust::spec_generator::codec::cst::decoder::misc::rust_wire_type_add_prefix_or_js_value;
+
+const NIL_FIELD: &str = "nil__";
 
 impl<'a> WireRustCodecCstGeneratorDecoderTrait for EnumRefWireRustCodecCstGenerator<'a> {
     fn generate_decoder_class(&self) -> Option<WireRustOutputCode> {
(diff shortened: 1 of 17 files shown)


frb_example/flutter_via_create/rust/src/frb_generated.io.rs
```diff
@@ -68,7 +68,7 @@ pub extern "C" fn cst_new_list_prim_u_8(len: i32) -> *mut wire_cst_list_prim_u_8
 }
 
 #[repr(C)]
-#[derive(Clone)]
+#[derive(Clone, Copy)]
 pub struct wire_cst_list_prim_u_8 {
     ptr: *mut u8,
     len: i32,
```
fzyzcjy, line 74: nit: I am a bit worried - should we make this Copy? Since it contains raw pointers, and copying it may unwantedly make more references than we realize.

frb_codegen/src/library/codegen/generator/wire/rust/spec_generator/codec/cst/decoder/ty/enumeration.rs
```diff
@@ -93,21 +96,12 @@ impl<'a> WireRustCodecCstGeneratorDecoderTrait for EnumRefWireRustCodecCstGenera
...
-            .iter()
-            .filter_map(|variant| self.generate_impl_new_with_nullptr_variant(variant))
-            .collect_vec();
-
         Some(WireRustOutputCode {
             body: generate_impl_new_with_nullptr_code_block(
                 self.ir.clone(),
                 self.context,
-                "Self { tag: -1, kind: core::ptr::null_mut() }",
+                "Self { tag: -1, kind: core::mem::MaybeUninit::uninit() }",
```
fzyzcjy, line 103: nit: currently, it seems that we use MaybeUninit + union. What if we do not use MaybeUninit and only use a union with one extra case. For example:

```
kind: MaybeUninit<SthKind>

pub union SthKind {
  apple: *mut Apple,
  orange: *mut Orange,
}
```

change to

```
kind: SthKind

pub union SthKind {
  nothing: (), // <--- And we init the union to this variant
  apple: *mut Apple,
  orange: *mut Orange,
}
```

fzyzcjy:
Great optimization! Just two nits (discussions).

frb_example/flutter_via_create/rust/src/frb_generated.io.rs
Desdaemon: This was a consequence of requiring Clone on structs, which means that MaybeUninit<T> requires T to be Copy. If we aren't using Clone for anything we could remove both.

fzyzcjy: I see. Then maybe we can avoid MaybeUninit below?

frb_codegen/src/library/codegen/generator/wire/rust/spec_generator/codec/cst/decoder/ty/enumeration.rs
Desdaemon: I think that would be similar to `MaybeUninit::zeroed()` so it would just be a stylistic choice, up to you. In fact this is how MaybeUninit is defined as well.

fzyzcjy: > In fact this is how MaybeUninit is defined as well.

Totally agree - that's why I have this proposal ;)

> I think that would be similar to MaybeUninit::zeroed() so it would just be a stylistic choice, up to you. 

I personally want to avoid adding those Copy, thus removing MaybeUninit looks a bit helpful

frb_example/flutter_via_create/rust/src/frb_generated.io.rs
Desdaemon: I checked back and it's not so simple, it seems that the Clone derive requires every union field to implement Copy which is stronger than a union's usual requirement of either ManuallyDrop or Copy.

I also tried removing both derives but we do have some functions that require Clone, so we're at an impasse. I think it's actually fine to implement Copy for both, since until they have been deserialized they are still packs of pointers and numbers which are trivially copyable without loss of safety; the real unsafety of pointers is at the point of dereferencing.

fzyzcjy: So, if I understand correctly, this means that, Copy is needed even without the usage of MaybeUninit?

If so, maybe we have to use Copy.

Desdaemon: Here's the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a064bde04abb0b443100b00651d7dc78) for the issue, but yes the only way for a union to implement Clone is to also implement Copy.

frb_example/pure_dart/rust/src/frb_generated.io.rs
```diff
@@ -39040,2127 +36229,2117 @@ pub struct wire_cst_exotic_optionals_twin_sync_sse {
...
     tag: i32,
-    kind: *mut SpeedTwinRustAsyncKind,
+    kind: SpeedTwinRustAsyncKind,
 }
 #[repr(C)]
+#[derive(Clone, Copy)]
 pub union SpeedTwinRustAsyncKind {
-    Unknown: *mut wire_cst_SpeedTwinRustAsync_Unknown,
-    GPS: *mut wire_cst_SpeedTwinRustAsync_GPS,
+    GPS: core::mem::ManuallyDrop<wire_cst_SpeedTwinRustAsync_GPS>,
```
fzyzcjy, line 37890: nit: Curious why do ManuallyDrop? We do not seem to generate drop functions manually, so will we have memory leaks?

fzyzcjy (commented):
LGTM! Just one nit about ManuallyDrop

frb_example/pure_dart/rust/src/frb_generated.io.rs
Desdaemon: Actually this is (was) a requirement for union fields that I briefly mentioned earlier, that they must be either Copy or wrapped in ManuallyDrop. And they are all Copy now so we can remove this.

fzyzcjy:
(I will review after a sleep, roughly 8 hours later - with a brief look this LGTM and ready to merge :) )

fzyzcjy (approved):

Issue #1458, opened 2023-12-19
[Bug] option "add-mod-to-lib" is named "no-add-mod-to-lib" in the docs


patmuk:
### Describe the bug

The documentation [states](https://cjycode.com/flutter_rust_bridge/guides/custom/codegen) the option name wrongly as `no-add-mod-to-lib` while it is `add-mod-to-lib`.

Additionally maybe wrong configuration options should be rejected - or at least a warning should be outputted.

### Codegen logs with `RUST_LOG=debug` environment variable

```shell
not relevant
```

### Version of `flutter_rust_bridge_codegen`

v2.0.0-dev.2

fzyzcjy:
> The documentation [states](https://cjycode.com/flutter_rust_bridge/guides/custom/codegen) the option name wrongly as no-add-mod-to-lib while it is add-mod-to-lib.

This is a bug in doc: Indeed, when using yaml, **all** things are positive; while using command line arguments, sometimes it is negative (`no-sth`).

I will auto generate a default config file and append it to doc when having time. 

> Additionally maybe wrong configuration options should be rejected - or at least a warning should be outputted.

Totally agree! I will also do that.

fzyzcjy:
Done on master branch as well :) Feel free to open issues if needed!

Issue #1464, opened 2023-12-21
Fix iOS CI which once in a while fails (flaky), while it works well locally


fzyzcjy:

fzyzcjy:
Lower priority, will work on it later

PR #1509, opened 2023-12-27
fixes #1508 and partially #1458 by adding configuration to fail when a field is wrong

### LINKED ISSUE #1508: Fail if the config file has a wrong key entry
As mentioned in #1458 flutter_rust_bridge_codegen should fail, if a key is misspelled.

Currently it continues silently, taking the default value. For example, if instead "no_add_mod_to_lib: true" one enters "no_mod_to_lib: true" the command silently uses the default value instead of the specified one.

Fail in the described case, so the error is noticed and fixed.

None. Ignoring wrong fields is the default of serde, so that structural changes in the Type still work with old configuration files. In this case, however, it should be explicit which settings are used instead of silently overwriting them. Breaking outdated configuration files is good, as otherwise the behavior using a different default value, would occur.

**Additional context**

patmuk:
## Changes

Fixes #the-issue-number: #1508 and partially #1458 
(#1458 needs a correction in the documentation, which I leave up to @fzyzcjy)

## Checklist

## Remark for PR creator

### DIFF
--- a/frb_codegen/src/binary/commands_parser.rs
+++ b/frb_codegen/src/binary/commands_parser.rs
@@ -133,8 +133,6 @@ mod tests {
     fn test_compute_codegen_config_mode_config_file_faulty_file() -> anyhow::Result<()> {
         configure_opinionated_test_logging();
         set_cwd_test_fixture("binary/commands_parser/flutter_rust_bridge_yaml")?;
-        // let result = std::panic::catch_unwind(|| {
-        // expecting serde to panic due to the faulty entry
         let result = run_command_line::<(), anyhow::Error>(vec![
             "",
             "generate",
--- a/frb_codegen/src/binary/commands_parser.rs
+++ b/frb_codegen/src/binary/commands_parser.rs
@@ -116,7 +116,12 @@ mod tests {
         configure_opinionated_test_logging();
         set_cwd_test_fixture("binary/commands_parser/config_file")?;
 
-        let config = run_command_line::<Config, anyhow::Error>(vec!["", "generate", "--config-file", "hello.yaml"])?;
+        let config = run_command_line::<Config, anyhow::Error>(vec![
+            "",
+            "generate",
+            "--config-file",
+            "hello.yaml",
+        ])?;
         assert_eq!(config.rust_input, "hello.rs".to_string());
         assert!(!config.dart3.unwrap());
 
@@ -163,14 +168,8 @@ mod tests {
         ];
         let config = run_command_line::<(), anyhow::Error>(common_args.clone())
             .expect("failed to parse cli args");
-        assert_eq!(
-            config.dart3,
-            Some(true)
-        );
-        assert_eq!(
-            config.rust_input,
-            "hello.rs".to_string()
-        );
+        assert_eq!(config.dart3, Some(true));
+        assert_eq!(config.rust_input, "hello.rs".to_string());
         assert_eq!(
             run_command_line::<Config, anyhow::Error>(concat([
                 common_args.clone(),
@@ -187,5 +186,4 @@ mod tests {
         let args = if_then_some!(let Commands::Generate(args) = cli.command, args).unwrap();
         compute_codegen_config(args.primary)
     }
-
 }
--- a/frb_codegen/src/binary/commands_parser.rs
+++ b/frb_codegen/src/binary/commands_parser.rs
@@ -58,6 +58,7 @@ mod tests {
     use crate::binary::test_utils::set_cwd_test_fixture;
     use clap::Parser;
     use itertools::concat;
+    use lib_flutter_rust_bridge_codegen::codegen::Config;
     use lib_flutter_rust_bridge_codegen::utils::logs::configure_opinionated_test_logging;
     use lib_flutter_rust_bridge_codegen::{codegen, if_then_some};
     use serial_test::serial;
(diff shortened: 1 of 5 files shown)


fzyzcjy:
Good job! I personally suggest adding a simple unit test or integration test to ensure that we never regress in the future:

Just mimic:
https://github.com/fzyzcjy/flutter_rust_bridge/blob/2cf7d1774f2557876d530bfbfed1fdd01b732b94/frb_codegen/src/binary/commands_parser.rs#L68

(it uses the following dir as fixture as well)
https://github.com/fzyzcjy/flutter_rust_bridge/tree/master/frb_codegen/test_fixtures/binary/commands_parser/flutter_rust_bridge_yaml

For example, add a new test mimicking it, with a broken .yaml file, and we `assert!` it should error.

patmuk:
> Good job! I personally suggest adding a simple unit test or integration test to ensure that we never regress in the future:
> 
> Just mimic:
> 
> https://github.com/fzyzcjy/flutter_rust_bridge/blob/2cf7d1774f2557876d530bfbfed1fdd01b732b94/frb_codegen/src/binary/commands_parser.rs#L68
> 
> (it uses the following dir as fixture as well) https://github.com/fzyzcjy/flutter_rust_bridge/tree/master/frb_codegen/test_fixtures/binary/commands_parser/flutter_rust_bridge_yaml
> 
> For example, add a new test mimicking it, with a broken .yaml file, and we `assert!` it should error.

Thanks :) And thanks for adding me as a contributor!

I implemented the unit tests - for the config file, but as well for the case in which the configuration is part of pub spec.yaml.
That I did not test before (!), but it works all as expected :)

fzyzcjy:
You are welcome!

frb_codegen/src/binary/commands_parser.rs
```diff
@@ -83,26 +84,68 @@ mod tests {
...
 
         Ok(())
     }
 
+    #[test]
+    #[serial]
+    fn test_compute_codegen_config_mode_config_file_faulty_file() -> anyhow::Result<()> {
+        configure_opinionated_test_logging();
+        set_cwd_test_fixture("binary/commands_parser/flutter_rust_bridge_yaml")?;
+        // let result = std::panic::catch_unwind(|| {
```
fzyzcjy, line 136:
```suggestion
```

fzyzcjy (approved):
LGTM!

frb_codegen/src/binary/commands_parser.rs
patmuk: Oh! Thanks, overlooked that! The comment below is unnecessary as well, since I changed it to propagate Result objects instead panicking. Removed that as well.
Funny, the code coverage dropped by one line - I guess it was that one. I will wait for the CI to finish before closing this thread.

patmuk: That was it - code coverage is good again (nice side effect to spot overlooked commented code lines :))!
There is a failing dart test, but I am sure that is because of #1464.
@fzyzcjy ready to be merged :)

patmuk: Oh, it looks like I was wrong - the codecov/project job still fails :(
Indeed, there is one new code line, but new tests - though I assume codecov doesn't understand that these tests cover it.
How can I configure this? Or do we just ignore it here?

fzyzcjy:
The code coverage tool is currently sometimes weird... 

https://app.codecov.io/gh/fzyzcjy/flutter_rust_bridge/pull/1509/indirect-changes

image

patmuk:
> The code coverage tool is currently sometimes weird... I commited a ignore and let's see whether it works.
> 
> https://app.codecov.io/gh/fzyzcjy/flutter_rust_bridge/pull/1509/indirect-changes
> 
> image

@fzyzcjy 
CodeCoverage still fails ...  
Do we need to derive the Macro I introduced?

in [`frb_codegen/src/library/codegen/config/config.rs`](https://app.codecov.io/gh/fzyzcjy/flutter_rust_bridge/pull/1509/blob/frb_codegen/src/library/codegen/config/config.rs)

```
  /// Configuration for code generation
  /// Refer to `GenerateCommandArgs` for documentations
  #[derive(Debug, Serialize, Deserialize)]
1
+ #[serde(deny_unknown_fields)]
  pub struct Config {
      pub base_dir: Option<String>,
      pub rust_input: String,
```
change to
```
// frb-coverage:ignore-start 
#[derive(Debug, Serialize, Deserialize)]
#[serde(deny_unknown_fields)]
// frb-coverage:ignore-end+ 
```
?

fzyzcjy:
Hmm, it is really weird... Let's just ignore it...

https://app.codecov.io/gh/fzyzcjy/flutter_rust_bridge/pull/1509?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=fzyzcjy

image

fzyzcjy:
Just make Lint::Rust CI happy and I will merge :)

patmuk:
> Just make Lint::Rust CI happy and I will merge :)

Done :) Reverted the change, now it is fine to be merged.

This closes #1508, but #1458 as well ... I assume the slight name difference between the cli argument or configuration file ("no-") is ok, as the error messages will help enough to understand that fast.
Otherwise we can change the attributes to be the same, should not be much work.

fzyzcjy:
Good job!

> I assume the slight name difference between the cli argument or configuration file ("no-") is ok, as the error messages will help enough to understand that fast.

I will update the doc about this in the next batch (possibly in a few days)

> Otherwise we can change the attributes to be the same, should not be much work.

I deliberately make `--no` only in command line, because it seems a bit counter-best-practice to have negations except for command line.

PR #1554, opened 2023-12-31
Improve upgrade.md by adding detailed clean up steps

### LINKED ISSUE #1492: Improve V2 migration experience
Happy holidays and thanks for the timely 2.0.0-dev.4 bump!

# Problem: V2 migration is underdocumented & has unexpected quirks
I followed the "Upgrade to V2" guide and managed to upgrade to 2.0.0-dev.4. However, **the upgrade experience has a few unexpected quirks that I wish other people can avoid.** In summary, the 4 steps listed in the doc is not enough in my experience:
> 1. Follow the [quickstart](https://cjycode.com/flutter_rust_bridge/quickstart) to integrate V2 boilerplate into the existing project (in a few commands).
> 2. Move original Rust code (e.g. in api.rs) to rust/src/api/simple.rs. (You can also split into multiple files, e.g. apple.rs, orange.rs, etc.)
> 3. Run code generator (whenever your code changes).
> 4. Fix compilation errors (usually just a rename or relocation) and runtime exceptions (usually have hints).

## 1. rust_builder missing
**The `rust_builder` library is not automatically generated** by running `flutter_rust_bridge_codegen generate`. I did `flutter run` after running generate but the api failed to initialize because of a missing dylib. I was quite puzzled for a while until I realized that the binary is now generated on-the-fly rather than prebuilt.

## 2. integrate overwrites existing files silently
So a natural step was to run `flutter_rust_bridge_codegen integrate` in hope of generating the `rust_builder`. However, I was surprised that **the integrate command completely overwrote my existing `main.dart` and `Cargo.toml`**. The command also autogenerated a few files that I don't need, including `rust/src/api/simple.rs` and `integration_test/simple_test.dart`.

## 3. Rust build errors hidden at default log level
I initially did not realize that those files were overwritten and just ran `flutter run` as usual. The command then mysteriously failed with a one-liner error message: `Could not build the precompiled application for the device.` I had to do `flutter run -v`, browse through the massive trace, and finally found out that the rust library is messed up due to the overwrite. Notably, I was able to see the Rust build errors in V1 because I explicitly ran cargo build myself. However, **the cargokit build errors in V2 are not even showing up at the default log level of `flutter run`.** (I was building for a real iOS device)

## 4. Missing step for removal of old files
Another step that's not mentioned in the migration steps is the **removal of any old native binary files, old bridge_generated files, any old imports, and running `flutter clean`**. This step is trivial but can cause confusion if missed during migration. For example, I forgot to remove one of the old bridge_generated imports and directly added the new imports to a dart file. This caused conflicting imports for many api functions I'm using and took me a while to realize the cause of the issue. Also running `flutter clean` seems to be required due to some old cached files needing to be deleted.

## 5. No obvious way to configure cargokit for release mode in flutter run
And lastly, cargokit defaults to debug build for rust when running `flutter run`, which is a sensible default. However, for my use case, **I always need a release build for rust** because otherwise the api is going to be horribly slow, even for testing. This is probably a bad implementation on my part but the release build speeds things up by 10x or more. I suspect that some other devs might also need a release build for testing because my assumption is that the business logic in rust is usually very performance-sensitive and might be changed relatively infrequently during UI testing. And as long as cargokit is smart enough to not rebuild rust every time I do `flutter run`, I can bear with the cost of a release build for rust.

# Proposed Changes

**1. rust_builder missing**
* Option 1: `flutter_rust_bridge_codegen generate` should generate `rust_builder` if it's missing.
* Option 2: Add instructions in migration to manually generate `rust_builder`

**2. integrate overwrites existing files silently**
integrate command should default to *not* overwrite existing files. Output a message listing the things that need to be changed in the existing files.

**3. Rust build errors hidden at default log level**
Show rust build errors when doing `flutter run` so the developers are informed of the issues.

**4. Missing step for removal of old files**
Add a step to migration guide detailing things to remove from V1

**5. No obvious way to configure cargokit for release mode in flutter run**
* Option 1: Add a config option to `flutter_rust_bridge_codegen` to specify cargokit build modes for flutter debug/profile/release builds.
* Option 2: Add instructions on how to configure cargokit to build in release mode even when flutter is in debug mode.

AlienKevin:
## Changes

Fixes #1492 

## Checklist

### DIFF
--- a/website/docs/guides/miscellaneous/upgrade.md
+++ b/website/docs/guides/miscellaneous/upgrade.md
@@ -4,10 +4,22 @@ This guide shows how to upgrade a project using flutter_rust_bridge V1 to V2.
 
 One way is as follows:
 
-* Follow the [quickstart](../../quickstart) to integrate V2 boilerplate into the existing project (in a few commands).
+* Follow the [quickstart](../../quickstart) to integrate V2 boilerplate into the existing project (in a few commands). Usually, you just need to run `flutter_rust_bridge_codegen integrate`.
 * Move original Rust code (e.g. in `api.rs`) to `rust/src/api/simple.rs`. (You can also split into multiple files, e.g. `apple.rs`, `orange.rs`, etc.)
-* Run code generator (whenever your code changes).
+* Run code generator and watch for code changes by running `flutter_rust_bridge_codegen generate --watch`.
 * Fix compilation errors (usually just a rename or relocation) and runtime exceptions (usually have hints).
+* Clean up old files and generated artifacts
+    * **Tear down automatic build**: If you followed fzyzcjy's [StackOverflow answer](https://stackoverflow.com/a/69515060/6798201) on automatic integration of Flutter and Rust, you also need to tear down the relevant configurations carefully. For the iOS setup, you need to reverse the setup steps you did by either removing the old Build Phase (if using method 1) or deleting the script_phase in your `Podfile` (if using method 2). For the Android setup, you need to reverse your changes to the `build.gradle` file.
+    * **Android-specific step**: Delete the `android/app/src/main/jniLibs/` folder, if this is where the old binaries are stored.
+    * **iOS-specific steps**:
+        1. Clean up `bridge_generated`:
+            * Delete the `ios/Runner/bridge_generated.h` file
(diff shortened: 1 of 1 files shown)


fzyzcjy:
Good job! I will check a bit more in detail later.

fzyzcjy (approved):
LGTM, will merge and do minor changes :)

PR #1547, opened 2023-12-30
documents customizing the rust folder location

patmuk:
## Changes
Documents customizing the rust project folder as done in #1459.

## Checklist

- [ ] An issue to be fixed by this PR is listed above.
- [ ] New tests are added to ensure new features are working. Please refer to [this page](https://cjycode.com/flutter_rust_bridge/guides/contributing/tip) to see how to add a test.
- [ ] `./frb_internal precommit --mode slow` (or `fast`) is run (it internal runs code generator, does auto formatting, etc).
- [x] If this PR adds/changes features, documentations (in the `./website` folder) are updated.
- [x] CI is passing. Please refer to [this page](https://cjycode.com/flutter_rust_bridge/guides/contributing/tip) to see how to solve a failed CI.

## Remark for PR creator
@fzyzcjy have a look and merge it at your convince.

PR #1561, opened 2023-12-31
(Partial implementation) Add placeholders for the needed replacements to the template files

### LINKED ISSUE #1459: FRB v2 Support external rust directory
Inspired by [crux](https://github.com/redbadger/crux) I want to use flutter as a UI-shell: The application core, it’s state, etc, should be in rust, where different shells are sending (mostly user triggered) events to the application core in rust, and receive effects to react to (typically what to render as a result of processing the event).  
I made a [demo project](https://github.com/patmuk/flutter-rust-bridge_crux_style/tree/flutter-rust-bridge_codegen) to implement this setup.
Next to a flutter shell you can find a cli shell.  
I might be over-engineering this, but I believe that the UI-shell should just be an add-on, so one can support multiple and/or switch in the future.

But I struggle to set this up properly. I believe that this is possible (Worked in FRB pre v1.80.0), but I don’t know if FRB v2 can be configured to do so - and if my approach is correct.
This is what I do: 
 
### Approach 1:  

Have the rust directory as a sibling of the flutter directory, with flutter-codegen.yaml configuring to use that.  

```
My app/
  app_core <- the rust dir
  shell_flutter <- result of flutter_rust_bridge_codegen create, minus /rust
  other shells
```

- currently my library is not found in flutter run, when I moved the folder and adapted the codegen config yaml. I should probably try pointing to the new location in RustLib.init(). I will get into more details on the issues along the way in follow-up comments.

### Approach 2:

I was just recently thinking that I could leave the internal rust folder, but import the app_core as a dependency, like:

```
My App/
  app_core <- the rust dir with the main app logic
  shell_flutter <- result of flutter_rust_bridge_codegen create
     rust <- including FBRs rust crate, which depends on app_core
  other shells
```

In this solution I would not fall into a trap, where app_core uses some FBR sepecific implementation detail and thus have a too strong coupling. But I don’t want to duplicate the interface code either (which is a handful of functions (process_event, …) and a few structures (Enum Event, …).
- currently I get a dependency issue in cargo build, some feature flag of anyhow is not compatible with stable.

Which sounds like the better, more future proof, approach? Keeping a “bridge” rust crate inside the flutter dir, depending on the app_core (and how to generate the needed code without repeating its implementation) or to move it out (making it only one api.rs file in app_core)?

Is there something hardcoded, or a principle problem with either approach?

patmuk:
## Changes

fixes #1459

## Checklist

## Remark for PR creator

This automatizes the needed steps described in #1547 to support any rust folder location.


patmuk:
@fzyzcjy I agree to make this automatic, as suggested in #1547. Though I won't have much time - this is wip.
Feel free to work on this as well!

fzyzcjy:
Sure! I may work on this after solving other issues. And since this is WIP, I temporarily use github's "draft" feature. Feel free to convert it back later!

patmuk:
Unfortunately I would have to get deeper into frb ... there are some more things to consider.  
I got stuck at the very beginning: Trying to get the configuration options for "rust_root" - or better, the rust_dir, which is derived from either "rust_root" or "rust_input" - so that I can get the directory, but project name as well from the cargo.toml. Did not understood how I can get the config-instance. Additionally I would need to set an env variable to the template directory - but I did not see where the ends are set so they are available at compile time.

IMHO the needed steps are:
- set an env variable pointing to the rust_builder template directory
- get the rust manifest file
- set "REPLACE_ME_RUST_MANIFEST_DIR" to it's directory
- read [lib.name] from the manifest file and set REPLACE_ME_RUST_LIB_NAME
- this is not trivial when we fbr_codegen integrate
  - though we could ask the user to specify the rust project name, assume rust_root as the default
- replace REPLACE_ME_RUST_MANIFEST_DIR and REPLACE_ME_RUST_LIB_NAME in the templates
- implement a new example, where the rust folder is parallel to the main frb folder
- implement tests

All-in-all not too less work ... but maybe the new example is not needed and other steps can be simplified as well.
There is at least one design decision: Shall we introduce a new option "rust_project_name" for the rare case that a user doesn't call his folder like the project? We could use this to set the correct name in cargo.toml for frb create, or not needing to parse an existing cargo.toml when frb integrate.

fzyzcjy:
No worries - there is no need to be this complex! Maybe I will just implement it directly.

patmuk:
> No worries - there is no need to be this complex! Maybe I will just implement it directly (and feel free to check the changed code, then you have more experience!).

That sounds great! Yes, will checkup on the implemented code!

fzyzcjy:
I am not sure, should I create a new PR, or continue on this PR? Seems that this is on [patmuk:enable_external_rust_folder](https://github.com/patmuk/flutter_rust_bridge/tree/enable_external_rust_folder) branch, not a branch in fzyzcjy/flutter_rust_bridge

patmuk:
Just use this PR.

Yes - I cannot push branches to your repo - thus I do branches in my fork. Though I assume you are having the same issue :) I could give you write access to my repo.
However, most probably I won't be able to contribute to this one, so a branch in your repo might be fine.

fzyzcjy:
Already done on https://github.com/fzyzcjy/flutter_rust_bridge/pull/1567 and waiting for CI
(I will check your another issue after a sleep!)

patmuk:
I took a look at the code: Looks very thoughtfully implemented :)
Yes, taking it from the cli args sounds like the most pragmatic solution.

Issue #1608, opened 2024-01-04
Implement code generation for NaiveDate


TrackerSB:
I'm using diesel to generate structs representing database entries. Given a PostgreSQL schema having a column of type `date` it generates a struct with a member of type `NaiveDate` for which `IntoDart` is not implemented. Trying to do so instantly runs into rusts orphan restriction.

Given that `IntoDart<NaiveDateTime>` is already implemented a solution should be  already within reach, I naively assume.

For the time being I can live with changing the column type to `timestamp`,  i.e. including a time resulting in `NaiveDateTime`.

fzyzcjy:
Looks reasonable and feel free to PR! flutter_rust_bridge uses allo-isolate's `IntoDart`, thus the procedure is to PR on https://github.com/shekohex/allo-isolate/blob/master/src/chrono.rs#L56 (just mimic existing ones), and then update flutter_rust_bridge.

TrackerSB:
--> https://github.com/shekohex/allo-isolate/pull/56

fzyzcjy:
(close since PR merged)

TrackerSB:
Thanks a lot for fixing and merging the PR!

fzyzcjy:
You are welcome!

PR #1631, opened 2024-01-08
Update dependency to allo-isolate

TrackerSB:
Include implementation of `IntoDart` for `NaiveDate` by updating `allo-isolate` (#1608).

Close https://github.com/fzyzcjy/flutter_rust_bridge/issues/1608


fzyzcjy:
> basically deletes a whole bunch of files and results in many more GitHub actions failing.

Maybe it is because it executes things like "run `flutter_rust_bridge_codegen create`". In other words, maybe need to update things in https://github.com/fzyzcjy/flutter_rust_bridge/tree/master/frb_codegen/assets/integration_template as well

fzyzcjy:
@all-contributors please add @TrackerSB for code

PR #1663, opened 2024-01-14
Support `Vec<Vec<u8>>` for CST/DCO codec (SSE codec already supports it)

fzyzcjy:
Depend on https://github.com/shekohex/allo-isolate/pull/57


fzyzcjy:
still awaiting allo-isolate to release next version

PR #1664, opened 2024-01-14
Fix module parsing when two modules have the same name

alanlzhang:
Fix for modules with same name, one in src/api/, and another in src/, use regex to ensure indent.

pure_dart directory structure is too clean, doesn't know how to write a clean test for this situation


fzyzcjy:
> pure_dart directory structure is too clean, doesn't know how to write a clean test for this situation

what about this: `src/api/deliberate_name_conflict.rs` + `src/deliberate_name_conflict.rs` (if I understand the problem correctly). And add a line comment to say "please do not modify the file name, because it is to test this PR"

fzyzcjy:
Btw if you see the "generate internal" CI fails, no worries and I will solve it (it is about auto generation scripts)

alanlzhang:
Emm, Could you show me how to fix CI/Lint::Rust::Primary error ? It seems just fail if there's code diff ?

fzyzcjy:
No worries about that as well, I will fix it later (though you can try `./frb_internal lint-rust --fix`, or manually apply whatever it complains)

alanlzhang:
codecov/project failed, don't know why since I added unit test for changes in cargo_expand.rs

fzyzcjy:
> codecov/project failed, don't know why since I added unit test for changes in cargo_expand.rs

https://app.codecov.io/gh/fzyzcjy/flutter_rust_bridge/pull/1664?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=fzyzcjy

no worries, it is generate::internal not succeed making it not covered

image


> CI / Test :: Flutter :: Native:: iOS iPhone12 Pro Max Simulator failed too, but I cannot figure how to fix it.

as long as one of two ios test passes, that's usually flaky

fzyzcjy:
Good job, the code roughly LGTM, I will check it a bit later (hopefully today)!

frb_codegen/src/library/commands/cargo_expand.rs
```diff
@@ -166,4 +167,29 @@ mod another {}";
...
     }
+
+    #[test]
+    pub fn test_extract_module_with_prefix() {
+        let src = "pub mod parent {
+    mod another {
+        // some code
+    }
+}";
+        let extracted = extract_module(src, Some(String::from("another"))).unwrap();
```
fzyzcjy, line 178: nit: it seems that we do not have `another` module (but only parent::another). I wonder whether we should fail, instead of outputing the whole text?

fzyzcjy (commented):
Good job! Only a nit, and I will take care of the internal test generating script

frb_codegen/src/library/commands/cargo_expand.rs
alanlzhang: image

I just mimic the old behavior because I don't know if other code depends on it. :)

fzyzcjy: I see. If CI is green, then I guess it is OK. Since that does not seem to be the correct behavior, I guess we can fix it.

alanlzhang: Great. Just return "" ? It looks like `extract_module` never fail.

fzyzcjy: I personally suggest making it fail (return `anyhow::Result` and use `bail!("error message")`), but panic may be OK as well

fzyzcjy:
Good job!

PR #1676, opened 2024-01-17
feat: strip attributes outside of cargo-expand

### LINKED ISSUE #1488: [Bug] Minor issues/improvements for frb attribute
### Describe the bug

Might open a PR soon after I'm done grokking the new v2 changes

- `#[frb(default)]` on parameters are not stripped, failing compilation
  - Requires detecting whether being expanded with cargo-expand or rustc (??)
  - Solution: rustc cfg flag at macro expansion only
- Token span information is not preserved, causing compiler errors to point at the attribute
- (feature) `#[frb(sync)]` on impl block to mark all functions as sync with opt-out

```shell
N/A
```

### To Reproduce

_No response_

### Expected behavior

_No response_

### Generated binding code

_No response_

### OS

_No response_

_No response_

### Flutter info

_No response_

### Version of `clang++`

_No response_

### Version of `ffigen`

_No response_

### Additional context

_No response_

### LINKED ISSUE #1720: Let Rust Analyzer understand the marker proc macros
### Describe the bug

When I open the rust folder in VSCode (the project was generated by flutter_rust_bridge), in many cases I have no autocompletion from rust analyzer. Maybe this bug is related to somewhere else, but right now its really annoying working without any autocomplete.

VSCode output displays nothing, the same behavior after restarting my rust analyzer server.

Here is my simple.rs file:

```rust
use std::sync::{Mutex, OnceLock, RwLock};

static COUNTER: OnceLock<RwLock<i32>> = OnceLock::new();

#[flutter_rust_bridge::frb(sync)] // Synchronous mode for simplicity of the demo
pub fn greet(name: String) -> String {
    format!("Hello, {name}!")
}

#[flutter_rust_bridge::frb(sync)]
pub fn get_counter() -> i32 {
    let locked = COUNTER.get().unwrap();
    *locked.read().unwrap()
}

#[flutter_rust_bridge::frb(sync)]
pub fn increment() {
    let mut counter = COUNTER.get().unwrap();
    let mut c = counter.write().unwrap();
    *c += 1;

    let one: OnceLock<i32> = OnceLock::new();
    let m = Mutex::new(0);
}

#[flutter_rust_bridge::frb(init)]
pub fn init_app() {
    // Default utilities - feel free to customize
    flutter_rust_bridge::setup_default_user_utils();
    COUNTER.set(RwLock::new(0)).unwrap();
}
```

Working with the COUNTER static no autocomplete, if I create for example a mutex inside a function, no autocomplete for that at all.

Do you have any advice?

### Steps to reproduce

Hint: A simple way to reproduce is to clone and modify the https://github.com/fzyzcjy/flutter_rust_bridge/tree/master/frb_example/dart_minimal example package according to your needs.
1. ...
2. ...
3. ...

### Logs

```shell
I have no error at all.
```

### Expected behavior

_No response_

### Generated binding code

_No response_

### OS

_No response_

_No response_

### Flutter info

_No response_

### Version of `clang++`

_No response_

### Additional context

_No response_

Desdaemon:
## Changes

- Inject a special `#[cfg]` only when running cargo-expand to preserve FRB attributes
- Preserve token span information

Closes #1488

Closed #1720 (<- edited by @fzyzcjy)

## Checklist

## Remark for PR creator


justfile
```diff
@@ -25,3 +25,7 @@ _rsync_ubuntu:
 
 _port_forward_ubuntu:
     ssh -L 8181:localhost:8181 ubuntu -N
+
+[no-cd]
+expand *args:
```
fzyzcjy, line 30: nit: what about

```suggestion
_expand *args:
```

to make it not publicly visible, since this may not be a standard procedure but seems like ad-hoc helpful tools

frb_macros/src/lib.rs
```diff
@@ -10,25 +10,48 @@ use proc_macro::*;
 // frb-coverage:ignore-start
 #[proc_macro_attribute]
 pub fn frb(attribute: TokenStream, item: TokenStream) -> TokenStream {
```
fzyzcjy, line 12: nit: would be great to have some unit tests to clearly show what this proc macro will do (show inputs, expected output, etc)

frb_macros/src/lib.rs
```diff
@@ -10,25 +10,48 @@ use proc_macro::*;
...
-        new_str.push_str(&item_str[start..end]);
-        new_str.push('\n');
-        last_end = end;
-    }
-    new_str.push_str(&item_str[last_end..]);
-    let item: TokenStream = new_str.parse().unwrap();
+    let mut attribute = format_frb_attribute(format!("#[frb({attribute})]"));
+    let item = strip_frb_attr(item);
+    attribute.extend(item);
+    attribute
```
fzyzcjy, line 16: nit:
```suggestion
    [format_frb_attribute(format!("#[frb({attribute})]")), strip_frb_attr(item)].concat()
```

Not sure, but I guess after `attribute.extend(item);`, the `attribute` variable does not mean attributes, but mean attributes *and* body?

frb_example/pure_dart/lib/src/rust/api/attribute.dart
```diff
@@ -14,7 +14,7 @@ Future<void> handleCustomizedStructTwinNormal(
     RustLib.instance.api.handleCustomizedStructTwinNormal(val: val, hint: hint);
 
 Future<UserIdTwinNormal> nextUserIdTwinNormal(
-        {required UserIdTwinNormal userId, dynamic hint}) =>
+        {UserIdTwinNormal userId = const UserIdTwinNormal(), dynamic hint}) =>
```
fzyzcjy, line 17: nit: would be great to add a test ensuring this

frb_codegen/src/library/commands/cargo_expand.rs
```diff
@@ -84,12 +87,14 @@ fn run_cargo_expand(
...
-        PathBuf::from("--theme=none"),
-        PathBuf::from("--ugly"),
-    ];
+    let args = command_args!(
+        "expand",
+        "--lib",
+        "--theme=none",
+        "--ugly",
+        "--config",
+        "build.rustflags=\"--cfg frb_expand\""
```
fzyzcjy, line 96: nit
```suggestion
        r#"build.rustflags="--cfg frb_expand""#
```

frb_codegen/src/library/commands/cargo_expand.rs
```diff
@@ -111,7 +116,12 @@ fn run_cargo_expand(
 
     let mut stdout_lines = stdout.lines();
     stdout_lines.next();
-    let ans = stdout_lines.join("\n").replace("/// frb_marker: ", "");
+    let ans = stdout_lines.join("\n");
+    static PATTERN: OnceLock<Regex> = OnceLock::new();
```
fzyzcjy, line 120: nit: what about lazy_static (slightly shorter code)

frb_codegen/src/library/commands/cargo_expand.rs
```diff
@@ -111,7 +116,12 @@ fn run_cargo_expand(
...
+    let pattern = PATTERN.get_or_init(|| {
```
fzyzcjy, line 121: hmm, not very sure, this is `cargo_expand` function which does not know what is a frb attribute, maybe we should not do replacement at this stage, but do it at other places

fzyzcjy (commented):
Looks great! Only some nits and discussions

frb_codegen/src/library/commands/cargo_expand.rs
Desdaemon: Personally I prefer this because no macros, but up to you

frb_codegen/src/library/commands/cargo_expand.rs
Desdaemon: We could refactor the post-processing out into another function, but this is the earliest we can do it and doesn't leave doc-comments hanging around for too long. And while cargo-expand doesn't know about `#[frb]`, rustc does because of the --config passed above.

frb_macros/src/lib.rs
Desdaemon: Interesting. Do you think tests in the macro crate could work?

fzyzcjy: IMHO yes, just standard rust unit tests, and consider `frb` as a normal rust function.
If not working because of `proc_macro_attribute`, maybe do something like `fn frb_inner` and test on that.

Our CI infra should already consider tests in frb_macros. To test this, we can firstly create a definitely failing test and see whether CI realize that.

frb_codegen/src/library/commands/cargo_expand.rs
fzyzcjy: Our codegen already uses many lazy_static, so I guess it is OK to be consistent

frb_codegen/src/library/commands/cargo_expand.rs
fzyzcjy: I see... Then what about this: Extract frb_expand (the "rustc does know it") *and* the post-processing, into a simple wrapper function. Say, `fn cargo_expand_2()` (bad name, just for demonstration), s.t. the inner cargo_expand dnk frb, only the wrapper knows

frb_macros/src/lib.rs
Desdaemon: I'm hitting the "procedural macro API is used outside of a procedural macro" so probably might have to refactor the logic out to a different crate over the weekend.

fzyzcjy: Hmm then maybe no need for this test, since we have abundant e2e tests.

frb_example/pure_dart/lib/src/rust/api/attribute.dart
Desdaemon: Added new fixture to test this! Also uses pretty_assertions for diffing.

Desdaemon:
I've been having difficulties running precommit as it keeps deleting everything inside flutter_via_create/integrate, wonder if it's intermittent or something more serious that needs fixing.

fzyzcjy:
> I've been having difficulties running precommit as it keeps deleting everything inside flutter_via_create/integrate, wonder if it's intermittent or something more serious that needs fixing.

Curently the `precommit slow` logic contains:

1. `precommit-generate` to run codegen for every package
2. `precommit-integrate` to run `flutter_rust_bridge_codegen create/integrate` for `flutter_via_create/integrate`, which deletes the folder and copy-paste-modify the internal templates into the folder

Thus, if the template is outdated, it will be pasted and override the newly generated thing...

This is suboptimal and maybe we can improve it, what do you think?

frb_codegen/src/library/utils/test_utils.rs
```diff
@@ -28,28 +28,43 @@ pub(crate) fn json_golden_test(
...
 
-    raw_golden_test(actual, &actual_str, matcher_path, |x| {
-        Ok(serde_json::from_str(&x)?)
-    })
+    raw_golden_test(
+        actual,
+        &actual_str,
+        matcher_path,
+        |x| Ok(serde_json::from_str(&x)?),
+        None,
```
fzyzcjy, line 36: Not very sure, so here we do not assert anything?

frb_codegen/src/library/utils/test_utils.rs
```diff
@@ -28,28 +28,43 @@ pub(crate) fn json_golden_test(
...
     actual_str: &str,
     matcher_path: &Path,
     deserializer: F,
+    asserter: Option<fn(&T, &T)>,
 ) -> anyhow::Result<()>
 where
     T: Eq + Debug,
     F: Fn(String) -> anyhow::Result<T>,
 {
+    #[cfg(test)]
```
fzyzcjy, line 66: nit: seems that we can make the whole fn or mod be `cfg(text)` since it is only used for tests

frb_codegen/src/library/codegen/generator/api_dart/spec_generator/class/field.rs
```diff
@@ -51,11 +51,19 @@ fn default_value_maybe_to_dart_style(value: &str, enable: bool) -> Cow<str> {
...
 fn default_value_to_dart_style(value: &str) -> String {
-    let mut split = value.split('.');
-    let enum_name = split.next().unwrap();
-
-    let variant_name = split.next().unwrap().to_string();
-    let variant_name = make_string_keyword_safe(variant_name.to_case(Case::Camel));
-
-    format!("{enum_name}.{variant_name}")
+    match value.split_once('.') {
+        // If the user is explicitly calling an enum variant's constructor
```
fzyzcjy, line 55: nit: would be great to add a quick unit test for this function (just inside this file and a few lines)

frb_macros/src/lib.rs
```diff
@@ -10,25 +10,50 @@ use proc_macro::*;
...
+    let mut output = format_frb_attribute(format!("#[frb({attribute})]"));
+    let item = strip_frb_attr(item);
+    output.extend(item);
+    output
+}
+
+fn strip_frb_attr(item: TokenStream) -> TokenStream {
+    item.into_iter()
+        .scan(None, |pound, tok| {
+            use TokenTree as T;
```
fzyzcjy, line 22: nit: `T` seems like generic argument, so maybe ``use TokenTree::*` or `as TT` may be better. But it depends on you and not a big problem

frb_codegen/src/library/commands/cargo_expand.rs
```diff
@@ -84,12 +93,14 @@ fn run_cargo_expand(
...
-        PathBuf::from("--theme=none"),
-        PathBuf::from("--ugly"),
-    ];
+    let args = command_args!(
+        "expand",
+        "--lib",
+        "--theme=none",
+        "--ugly",
+        "--config",
+        r#"build.rustflags="--cfg frb_expand""#
```
fzyzcjy, line 102: nit: as mentioned earlier, it would be great to do something like:
```
impl CachedCargoExpand {
  fn execute_awaring_frb() {
    unwrap_frb_attrs_in_doc(self.execute(extra_args = r#"build.rustflags="--cfg frb_expand""#))
  }
}
```

and let the inner code unaware of frb

fzyzcjy (commented):
LGTM, only a few nits :)

frb_codegen/src/library/utils/test_utils.rs
Desdaemon: I'll refactor this to make it more obvious, but a default asserter is provided for None.

Desdaemon:
> Thus, if the template is outdated, it will be pasted and override the newly generated thing...

Interesting, will check tomorrow what might make the template be outdated.

frb_codegen/src/library/utils/test_utils.rs
fzyzcjy: Get it, no worries (just a quick nit)

PR #1718, opened 2024-02-03
More friendly error message for Unit Struct

alanlzhang:
for unit struct:

```rust
pub struct App;
```

Error message before: 

```
struct with unit fields are not supported yet, what about using `struct YourStructName {}`
```

Error message after: 

```
struct with unit fields are not supported yet, what about using `struct App {}`
```

which is much more friendly for users to figure how to fix.

PS. I don't know how to write a good test for this ...


fzyzcjy (approved):
LGTM :)

(waiting for CI to be green)

fzyzcjy:
(no worries about the flaky codecov)

alanlzhang:
I think all tests (except the coverage one) are passed, ^_^

fzyzcjy:
Oh yes, github does not give me any notifications!

PR #1954, opened 2024-05-19
Support when struct impl block is outside api folder

### LINKED ISSUE #1906: An Issue with methods of Rust struct and other V2.0 questions
This is my 1st time using Frb v2.0(with [dev. 32](https://github.com/fzyzcjy/flutter_rust_bridge/releases/tag/v2.0.0-dev.32)) in my own project since Frb v1.x.
1. An Issue with methods of Rust struct:  
If the struct is not defined in the `Api`  folder, then its methods can't be generated 
--- This reminds me of the heavy time I spent on the multi-block issue PR to solve this issue(the potential recursive issue,  because one of the fields of the issued rust struct  could be another rust struct/enum, which may contain its own public methods...).
(Given the extensive time and effort I have already dedicated to this issue, I am hesitant to consider further contributions in this area, especially V2.0 is still not stable and under active development. My latest v1.x modification without this issue could be referred [here](https://github.com/dbsxdbsx/flutter_rust_bridge), though it has other bugs)

Questions on V2.0:
1. Within V2.0, for the tag `rust_input` in `flutter_rust_bridge.yaml`, how to specify more than one file respectively? (Seems no way at present)
I know `rust_input: rust/src/api/**/*.rs` is supported, but how to specify more than one file respectively.
2. Within V2.0, now all the API fns can be directly invoked without a prefix, But I still suggest generating Dart Classes accordingly as in V1.x to make it more distinguishable.
For example, in multi-block cases, with prefixes like `Api1.` `Api2.` can make it more clear in Dart side, and it can also be used to distinguish the same name dart methods---indeed, currently, it can be done by manually adding the dart class, but a little verbose, since it has to be done everytime the rust side fns is changed.
3. V2.0 introduces many macros, like `#[frb(sync)]`,`#[frb(ignore)]`, and compound usage case, I did find them in
separated guidance file, but I think it is better to systematically introduce them in a specific page file in guidance in table style.

fzyzcjy:
## Changes

Close #1906

## Checklist

## Remark for PR creator


fzyzcjy:
The semantics may be suboptimal (thus closed). If someone wants this, feel free to discuss!

Issue #2143, opened 2024-06-22
Add The Ability To Generate Plugins From The CLI Tool


mcmah309:
Currently the cli tool can only generate app projects, it should also be able to generate plugins (discussed here https://github.com/fzyzcjy/flutter_rust_bridge/discussions/2125)

PR #2144, opened 2024-06-22
Add The Ability To Generate Plugins From The CLI Tool

mcmah309:
## Changes

closes https://github.com/fzyzcjy/flutter_rust_bridge/issues/2143

Adds the `template` field to `flutter_rust_bridge_codegen create --template plugin` (similar to `flutter create --template plugin`) and `type` to `flutter_rust_bridge_codegen integrate --template plugin`. As well as opens the door for more templates going forward.

### Notes For Reviewers
There appear to be a lot of code changes, but that is mainly due to `frb_codegen/assets/integration_template` being split into `shared`, `app`, and `plugin`. And updating `frb_example/flutter_package`

## Checklist

## Remark for PR creator


mcmah309:
@fzyzcjy could I get your thoughts on this PR when you have the chance? Not sure if I covered all the related changes for a release. If all looks good, I can update the docs and change status to ready for review.

fzyzcjy:
Good job! Some nits

- [ ] Maybe we can make `integrate` and `create` use the same flag (e.g. both use `template`) to avoid confusion. (Or, do you choose `type` because there are some conventions there)
- [ ] To test it, try to mimic how `flutter_via_create` is tested. More specifically, (1) add a line probably below https://github.com/fzyzcjy/flutter_rust_bridge/blob/b68eb1db312293c7b2636c83e0187b465ce40488/tools/frb_internal/lib/src/makefile_dart/consts.dart#L31 (2) Add a line below https://github.com/fzyzcjy/flutter_rust_bridge/blob/b68eb1db312293c7b2636c83e0187b465ce40488/.github/workflows/ci.yaml#L242. The test may name `flutter_package`. The "generate_run_frb_codegen_command_integrate" test means nothing but "let's run the `codegen create/integrate` command and check the result agrees with existing code".

You can test any CI thing locally by copy-paste-run the `./frb_internal ...` command in corresponding CI section.

mcmah309:
@fzyzcjy2 Thanks for the feedback!
- Changed to `template`
- Added tests

This should be ready for review. I was unable to get some commands with `./frb_internal` working locally like `./frb_internal generate-run-frb-codegen-command-integrate --set-exit-if-changed --package frb_example--flutter_via_integrate --coverage`. Issues seem to stem from running in dev container, and also running flutter `3.22.2`. Side note: `fvm` does not currently work in dev containers (I run NixOs so dev containers are a must) [issue link](https://github.com/leoafarias/fvm/issues/587). So I had to update `flutter_package` manually which will likely fail the ci since the ci is using `3.22.0`. Rather than further banging my head against a wall to get things to work in my environment, I'd ask you or someone else to make the project file changes that satisfy the `git diff` test. Running everything by hand locally looks fine though.

You may want to separate `flutter_package` into `flutter_package_via_create` and `flutter_package_via_integrate` like you did with `flutter_via_create` and  `flutter_via_integrate`. But I do think `diff` testing may be a development bottleneck and I recommend creating projects then invoking their `integration_test` suites instead. But that is just a matter of opinion.

In the future it may be worth while to attach a dev `.container` file for development. That way local dev and ci dev can be seamless.

fzyzcjy:
Looks great! (Btw you mentioned the wrong account ;) that account is to auto publish new versions of flutter_rust_bridge).

Ok, I will run the diff generation later.

> But I do think diff testing may be a development bottleneck and I recommend creating projects then invoking their integration_test suites instead.

IMHO we do diff testing because, we want to know whether the `codegen create` command really works or not.

> Also, it may be worth while to attach a dev Containerfile to the this project for development. That way local dev and ci dev can be seamless. I know I would appreciate it.

That's interesting, and feel free to PR for this!

.fvm/fvm_config.json
fzyzcjy, line 1: nit: Hmm, shall we keep the old fvm_config? then it is compatible with older fvms.

frb_codegen/src/binary/commands.rs
```diff
@@ -194,6 +199,20 @@ pub(crate) struct IntegrateCommandArgs {
...
+    #[clap(short, long, value_enum, default_value = "app")]
+    pub template: ProjectTypeArg
+}
+
+#[derive(Debug,Clone,ValueEnum)]
+pub(crate) enum ProjectTypeArg {
+    /// (default) a Flutter application
+    App,
+    /// A shareable Flutter project containing an API in Dart code with a 
+    /// platform-specific implementation
```
fzyzcjy, line 214: nit: not really "platform-specific" indeed, since we can have same rust on different platforms (thus the text copied from official flutter may need a bit of change)

frb_codegen/src/library/commands/flutter.rs
```diff
@@ -1,15 +1,20 @@
 use crate::command_run;
 use crate::commands::command_runner::call_shell;
 use crate::library::commands::command_runner::check_exit_code;
+use crate::misc::ProjectType;
 use log::info;
 use std::path::Path;
 
 #[allow(clippy::vec_init_then_push)]
-pub fn flutter_create(name: &str, org: &Option<String>) -> anyhow::Result<()> {
+pub fn flutter_create(name: &str, org: &Option<String>, template: &ProjectType) -> anyhow::Result<()> {
```
fzyzcjy, line 9: nit: Make `ProjectType` Copy, thus no need for `&`

frb_codegen/src/library/integration/integrator.rs
```diff
@@ -1,24 +1,37 @@
...
+
+static PLUGIN_INTEGRATION_TEMPLATE_DIR: Dir<'_> =
+    include_dir!("$CARGO_MANIFEST_DIR/assets/integration_template/plugin");
 
 pub struct IntegrateConfig {
     pub enable_integration_test: bool,
     pub enable_local_dependency: bool,
     pub rust_crate_name: Option<String>,
     pub rust_crate_dir: String,
+    pub template: ProjectType,
```
fzyzcjy, line 34: nit: since we rename to template, maybe rename

* ProjectType
* ProjectTypeArg

to template as well

website/docs/manual/miscellaneous/92-archived/011-library/01-overview/01-setup.md
fzyzcjy, line 1: Hmm maybe a typo: since 07 is renamed to 011 for the whole folder

tools/frb_internal/lib/src/makefile_dart/release.dart
```diff
@@ -90,7 +90,7 @@ void _updateVersionInText() {
 
   for (final package in ['flutter_rust_bridge', 'flutter_rust_bridge_macros']) {
     simpleReplaceFile(
-      '${exec.pwd}frb_codegen/assets/integration_template/REPLACE_ME_RUST_CRATE_DIR/Cargo.lock.template',
+      '${exec.pwd}frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.lock',
```
fzyzcjy, line 93: nit: do not use Cargo.lock, but use Cargo.lock.template. IIRC if we use cargo.lock, when publishing rust package, cargo will be confused and *remove* the files from our template, then we do not have this file!

website/docs/manual/integrate/04-flutter-package.md
```diff
@@ -0,0 +1,5 @@
+# Flutter package
+
+## Quickstart
+
+To have a working quickstart run use the cli tool and run `flutter_rust_bridge_codegen create flutter_package --template plugin` or alternatively copy https://github.com/fzyzcjy/flutter_rust_bridge/tree/master/frb_example/flutter_package
```
fzyzcjy, line 5: nit: maybe 

- [ ] reverting doc changes except for the change in cargokit.md
- [ ] In `quickstart.md` the "flutter package" tab, mimic the "add to existing" or "craete new" tab to explain the one-line command

frb_codegen/src/library/misc/mod.rs
```diff
@@ -1 +1,10 @@
 pub(crate) mod consts;
+
+/// The type of the project
+pub enum ProjectType {
```
fzyzcjy, line 4: nit for @fzyzcjy (you dont need to do anything): maybe I should move it later

frb_codegen/src/library/integration/utils.rs
```diff
@@ -4,30 +4,47 @@ use log::debug;
 use std::fs;
 use std::path::{Path, PathBuf};
 
-// ref: `Dir::extract`
+use crate::utils::path_utils::path_to_string;
+
+/// Recursively copies the file system structure starting at `current_reference_dir`
+/// to `base_target_path` while filtering by `filter` and invoking `modifier` on each file.
 pub(super) fn extract_dir_and_modify(
```
fzyzcjy, line 11: hmm, it would be great to make this specific function unaware of things like rust crate dir (i.e. unaware of things in ReplaceContentConfig). Otherwise this one function is doing two things.

I guess you are doing this to avoid repeating the logic of replace rust crate dir etc - that's great! Maybe we can do it like:

* keep extract_dir_and_modify like the original one (i.e. only do one thing)
* create sth like `fn extract_dir_and_modify_that_is_aware_of_our_need_to_replace_some_variables()`, this one calls the extract_dir_and_modify, and this one is aware of the rust crate name logic etc.

frb_codegen/src/library/integration/integrator.rs
```diff
@@ -33,33 +46,78 @@ pub fn integrate(config: IntegrateConfig) -> Result<()> {
         .clone()
         .unwrap_or(format!("rust_lib_{}", dart_package_name));
 
+    info!("Extract and modify the Dart project directory");
```
fzyzcjy, line 49: for @fzyzcjy (you dont need to do anything): review this after the comment above is changed, since this one depends on that one

fzyzcjy (commented):
Well done, and mainly some nits!

website/docs/manual/miscellaneous/92-archived/011-library/01-overview/01-setup.md
mcmah309: I moved it to the `archived` folder, since it was outdated, so it become 011 (should probably have been 11 though). I moved it back based on
 > - reverting doc changes except for the change in cargokit.md

mcmah309:
> Looks great! (Btw you mentioned the wrong account ;) that account is to auto publish new versions of flutter_rust_bridge).

Oh haha my bad.

> Ok, I will run the diff generation later.

Thanks!

> IMHO we do diff testing because, we want to know whether the `codegen create` command really works or not.

Well the best way to test if it works IMHO is to actually run the code. That's what I meant by

>> I recommend creating projects then invoking their integration_test suites instead.

Like
```bash
flutter_rust_bridge_codegen create flutter_package --template plugin &&
flutter test flutter_package/example/integration_test/simple_test.dart &&
rm -r flutter_package
```
(probably best run the test for each platform).
rather than running into constant `diff` discrepancies.

> That's interesting, and feel free to PR for this!

I may do that. I could get started at least, write a ticket for what needs to be done and setup rust/flutter in a Containerfile and .devcontainer. Then someone else may be best take over from there. Not super familiar with github ci or additional dev dependency issues I ran into earlier.

This should be good to review again.

tools/frb_internal/lib/src/makefile_dart/release.dart
fzyzcjy: > I'm curious as to why publishing a rust crate would remove a Cargo.lock file in a different part of the file hierarchy that is not referenced from the root Cargo.toml.

I don't know, maybe cargo publish or crates.io has some logic?

fzyzcjy: But anyway, let's keep it the old way (Cargo.lock.template). At the beginning I used Cargo.lock and later it has the trouble I mentioned thus I changed to .template ...

fzyzcjy:
> Well the best way to test if it works IMHO is to actually run the code.

We have two things on ci indeed:

1. diff test: generate it and see it matches
2. running test: run and test the commited code 

> I may do that. I could get started at least, write a ticket for what needs to be done and setup rust/flutter in a Containerfile and .devcontainer. Then someone else may be best take over from there. Not super familiar with github ci or additional dev dependency issues I ran into earlier.

That looks pretty reasonable!

frb_codegen/src/library/integration/utils.rs
fzyzcjy: I see. For that specific question, maybe pass `compute_effective_path` as an *argument* may help.

frb_codegen/src/library/integration/integrator.rs
```diff
@@ -33,33 +43,83 @@ pub fn integrate(config: IntegrateConfig) -> Result<()> {
...
+                    reference_content,
+                    existing_content,
+                    &replacements,
+                    config.enable_local_dependency,
+                    Some(&[&format!("{}.dart", &dart_package_name)]),
+                )
+            },
+            &|path| filter_file(path, config.enable_integration_test),
+        )?,
+    }
```
fzyzcjy, line 107: nit: The two blocks looks quite duplicated. maybe we can extract the common parts. e.g. easiest way is to

```
let (dir, somename) = match ... {
    App => (APP_INTEGRATION_TEMPLATE_DIR, "main.dart"),
   Plugin => ... 
};
...

fzyzcjy (commented):
LGTM, only a nit!

tools/frb_internal/lib/src/makefile_dart/release.dart
mcmah309: Sure. I reverted the change. But going off my last concern, maybe I'm missing some logic, but won't the file get copied over as `Cargo.lock.template` when creating a project and since it ends with `.template` Cargo will ignore it and create a new lock file?

fzyzcjy (approved):
LGTM! I will keep it unmerged for a bit of time (maybe 1d?), since need to run codegen etc for this as mentioned above

fzyzcjy:
@all-contributors please add @mcmah309 for code

fzyzcjy:
Update: https://github.com/fzyzcjy/flutter_rust_bridge/pull/2156#issuecomment-2188784622

PR #2156, opened 2024-06-25
Add The Ability To Generate Plugins From The CLI Tool (continued)

fzyzcjy:
## Changes

_Please list issues fixed by this PR here, using format "Fixes #the-issue-number"._

## Checklist

- [ ] An issue to be fixed by this PR is listed above.
- [ ] New tests are added to ensure new features are working. Please refer to [this page](https://cjycode.com/flutter_rust_bridge/guides/contributing/tip) to see how to add a test.
- [ ] `./frb_internal precommit --mode slow` (or `fast`) is run (it internal runs code generator, does auto formatting, etc).
- [ ] If this PR adds/changes features, documentations (in the `./website` folder) are updated.
- [ ] CI is passing. Please refer to [this page](https://cjycode.com/flutter_rust_bridge/guides/contributing/tip) to see how to solve a failed CI.

## Remark for PR creator

- `./frb_internal --help` shows utilities for development.
- If fzyzcjy does not reply for a few days, maybe he just did not see it, so please ping him.

PR #2158, opened 2024-06-25
Feat/12413 to master

mcmah309:
## Changes

from: https://github.com/fzyzcjy/flutter_rust_bridge/pull/2156
closes https://github.com/fzyzcjy/flutter_rust_bridge/issues/2143

## Checklist

## Remark for PR creator


mcmah309:
@fzyzcjy I resolved the windows test issue

fzyzcjy:
Replying https://github.com/fzyzcjy/flutter_rust_bridge/pull/2156#issuecomment-2189869853:

I suggest to have a look at the old flutter_package. IIRC, flutter_package vs flutter_via_create is like:

* / vs /rust_builder
* /example vs /

mcmah309:
I basically just copied `flutter_package` and made the necessary fields into templates and made some code changes to support this. Is this the approach you wanted/how does it look now?

fzyzcjy:
Good job!

However, I am not sure whether this is the best approach. One way is to append text to existing files, e.g. for build.gradle we only need a few lines:

```
apply from: "../cargokit/gradle/plugin.gradle"
cargokit {
    manifestDir = "../REPLACE_ME_RUST_CRATE_DIR"
    libname = "REPLACE_ME_RUST_CRATE_NAME"
}
```

that would be especially great when we `integrate` it (currently we `create` it) into an existing package with modifications on those things. However, as long as the current code works, I am happy to merge, since this works for simpler cases, and users may modify manually for complex cases.

mcmah309:
Going to step away. Think all tests should pass and we should be good to go! :crossed_fingers:

mcmah309:
@fzyzcjy Might need another pair of eyes to figure this out. For https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/9674492124/job/26690272746?pr=2158 , It looks like `build/linux/x64/debug/bundle/lib/librust_lib_flutter_via_create.so` does not exist. But `build/linux/x64/debug/plugins/rust_lib_flutter_via_create/cargokit_build/x86_64-unknown-linux-gnu/debug/librust_lib_flutter_via_create.so` does exist. So it appears it is not getting copied over to the bundle. To me, no code changes suggest this should be happening, but I am also not familiar with how this build process works.

fzyzcjy:
> but I am also not familiar with how this build process works.

maybe check the cargokit doc/tutorial first, which discusses what is done.

For that error, have not checked into details, but first thing maybe, check those cargokit configurations, especially something like `../../../rust` etc to ensure the folder is right.

mcmah309:
I have read it, nothing stands out. I am probably missing something dumb that's why I am asking for another set of eyes when you get the chance. Looking at the generate `flutter_package` it looks exactly as it should as far as I can tell, I've read it over at least 10 times.

fzyzcjy:
Hmm, seems you mentioned librust_lib_flutter_via_create, while this is flutter_package. Maybe some name is wrong? 

Hmm but this seems good:

image

fzyzcjy:
Maybe search `frb_example_flutter_package` (or something like that), and ensure you have replaced *all* names to the correct one (rust_lib_flutter_package?). If missing one or two then it can be the cause.

mcmah309:
> Hmm, seems you mentioned librust_lib_flutter_via_create, while this is flutter_package. Maybe some name is wrong?

Yeah sorry that's why I did a strike through on my last comment. I was looking in and copied the wrong path. The `.so` is missing in both for `flutter_package`

mcmah309:
I think I figured it out.. well see

mcmah309:
Everything looks good except a few generator diff tests :roll_eyes: which are flakey since `Cargo.toml` may re-order depending on name of the crate, different failing https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/9678895889/job/26703681330?pr=2158 and here https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/9678640001/job/26702895836 . Do you mind adding the Cargo.lock file to the diff exclusion?

fzyzcjy:
Looks great!

> Do you mind if I add the Cargo.lock file to the diff exclusion?

Looking at [`a501b1d` (#2158)](https://github.com/fzyzcjy/flutter_rust_bridge/pull/2158/commits/a501b1d78cf30deeae56a5fd19a31349b9fbc437) seems we already exclude many locks, thus I guess it is OK

mcmah309:
Ready for a review and/or merge when you get the chance

fzyzcjy:
Great! Looks like one CI fails, maybe we can fix it?

mcmah309:
Is this what you are referring to? https://app.codacy.com/gh/fzyzcjy/flutter_rust_bridge/pull-requests/2158/issues ? This file was generated by `flutter create <package_name>`

mcmah309:
When I run `./frb_internal precommit --mode slow` (I was able to get my infra working and I setup a new container to use flutter `3.20.0`) It actually tells me I should make these changes in the other projects as well.
img
My guess is, since this only happens on the macos build, you are running macos locally and these files are generated when running `flutter create` otherwise I'm not sure (another reason why I am not really a fan of diff testing). I think I have taken this as far as I can on my end, I hope this is not blocker

fzyzcjy:
I see. What about applying that diff (e.g. copy-paste those diff into a patch file and git apply diff). Then CI will be green and we can merge :)

fzyzcjy (approved):

mcmah309:
Alright

PR #2259, opened 2024-08-28
Fix segfault when passing DartOpaque through ffi boundaries

alexlapa:
## Changes

So, ive encountered a segfault on Android devices after upgrading to frb v2.2.

```
F/libc    (26466): Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x7fffffffffffffff in tid 26639 (1.ui), pid 26466
(a_jason_example)
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
Build fingerprint: 'Redmi/rosemary_eea/rosemary:13/TP1A.220624.014/V14.0.11.0.TKLEUXM:user/release-keys'
Revision: '0'
ABI: 'arm64'
Timestamp: 2024-08-28 04:17:38.660867792+0300
Process uptime: 7s
ZygotePid: 848
Cmdline: com.instrumentisto.medea_jason_example
pid: 26466, tid: 26639, name: 1.ui  >>> com.instrumentisto.medea_jason_example <<<
uid: 10320
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x7fffffffffffffff
    x0  7fffffffffffffff  x1  b400007859d20b00  x2  7fffffffffffffff  x3  0000000000000014
    x4  000000760000a431  x5  0000007859ce5200  x6  00000077fa0bb400  x7  00000077fa0bb000
    x8  7fffffffffffffff  x9  fffffffffffffff0  x10 0000000000000010  x11 0000000000000008
    x12 0000007859ce3830  x13 000000000781a8e0  x14 000000000781a4e0  x15 00000077ed6de080
    x16 0000000000000000  x17 b4000077fa075198  x18 00000077eaefe000  x19 000000785b4adddc
    x20 0000007606259759  x21 b4000077fa074a00  x22 0000007600008081  x23 00000076062596d9
    x24 0000007600008081  x25 00000077ed4fd000  x26 b4000077fa074a00  x27 0000007606592f70
    x28 0000000800000076  x29 00000077ed6de098
    lr  00000077b9459ae4  sp  00000077ed6ddfb0  pc  00000077b945f6ac  pst 0000000080001000
backtrace:
      #00 pc 000000000079a6ac
      /data/app/~~ffeMhpu5eN-Cb-D2BNeVnw==/com.instrumentisto.medea_jason_example-XtxdlZivlameAOYXlt7bTQ==/base.apk
      #01 pc 0000000000794ae0
      /data/app/~~ffeMhpu5eN-Cb-D2BNeVnw==/com.instrumentisto.medea_jason_example-XtxdlZivlameAOYXlt7bTQ==/base.apk
      #02 pc 000000000078ba5c
      /data/app/~~ffeMhpu5eN-Cb-D2BNeVnw==/com.instrumentisto.medea_jason_example-XtxdlZivlameAOYXlt7bTQ==/base.apk
      #03 pc 000000000078e818
      /data/app/~~ffeMhpu5eN-Cb-D2BNeVnw==/com.instrumentisto.medea_jason_example-XtxdlZivlameAOYXlt7bTQ==/base.apk
      #04 pc 000000000078e7ec
      /data/app/~~ffeMhpu5eN-Cb-D2BNeVnw==/com.instrumentisto.medea_jason_example-XtxdlZivlameAOYXlt7bTQ==/base.apk
      (dart_opaque_rust2dart_decode+12)
      #05 pc 00000000000081d4  [anon:dart-code]
```

So whats going on is if a pointer does not fit in u32, it is being changed to u32::max right [here](https://github.com/fzyzcjy/flutter_rust_bridge/blob/c81a186cd7146b6be23799fac89b13eda616847e/frb_dart/lib/src/dart_opaque/_common.dart#L8) which i was able to confirm with additional logging:

```
decodeDartOpaque toInt call 12970367442122235088 => 9223372036854775807
```

Another thing is that for some reason most targets tend to create pointer with values that fit into u32 so it works fine on macOs on Intel silicon, Linux and android x86_64 thats probably why this problem was able to stay under the radar.

Thats also possible that the same exact problem might happen in some other places, since passing pointers as usize is a pretty common thing i believe, but with this fix all my tests pass and app is working properly. However there might be some other scenarios that im not aware of (perhaps when using other codecs?).

## Checklist

## Remark for PR creator


fzyzcjy:
Great catch!

Btw, I wonder what platform are you using? Dart `int` is 64bit (Web is 32bit; but the pointers are also 32bit on web), so in non-web platforms the https://github.com/fzyzcjy/flutter_rust_bridge/blob/c81a186cd7146b6be23799fac89b13eda616847e/frb_dart/lib/src/dart_opaque/_common.dart#L8 seems should not convert to u32::max.

(If you need I can provide some suggestions for e.g. what precommit subcommand to run to make CI happy)

alexlapa:
> (If you need I can provide some suggestions for e.g. what precommit subcommand to run to make CI happy)

Yeah, i've seen that precommit script but it wasnt working for me until i've rolled back local rust and flutter versions. 

And i would like to correct myself - `BigInt.toInt()` is not a problem and it works as expected (`0xffffffff.toInt() == -1`), its [this](https://github.com/fzyzcjy/flutter_rust_bridge/blob/bfc421994e755b9cbc24260f4ab5423afffb9130/frb_dart/lib/src/generalized_typed_data/_io.dart#L124) `toUnsigned(64)` that breaks pointers. So, i believe this fix can be reworked to continue using `usize` for pointers, so let me know if you think this is the right way.

Another thing that bothers me is that there is a pretty high possibility of a similar issue happening in some other cases, since passing pointers as `usize` is a pretty common thing, and all `usize`s are treated with `toUnsigned(64)` on the Dart side it seems. Maybe there is a same issue with the `dco` codec? 

Other than that i believe its ready for review.

alexlapa:
> To test cst & dco codec, you can (temporarily) set full_dep: true in config.

So i've run some tests and returning `DartOpaque` from Rust to Dart works fine using dco even without this PR, so its not affected.

> Btw, is it possible we somehow add some tests for this scenario?

Meh, i dont think there is an easy way to do this. 

Running test on CI on a `arm64-v8a` Android will indirectly check for this issue, but thats not really a specific test. And that might be problematic since AFAIK there are some limitations of android emulator, i believe you cant run arm Android on x86 host since 27(?) API. Running that on a MacOS with Apple silicone might do the job but i never checked that.

Another option is implementing allocator(wrapping default one?) that will always return pointers that are bigger then `i64::MAX`. But im not really sure how to do that correctly.

Another option is not using `MirTypePrimitive::Isize` or `MirTypePrimitive::Usize` for this but maybe adding `MirTypePrimitive::Ptr` so we might add some tests for this type that will just check the hardcoded values. Like return `u64::MAX` from Rust and assert that its `-1` in Dart. This way it would also make `*const ()` a correct return for the Rust exported functions, which would be pretty cool.

> Thus curious what will happen when Dart sees an address like your case which is larger than maximum of i64

So web is out of the question since its 32-bit. And io will be fine since u64 can be represented as i64. E.g: Rust will return `u64::MAX`, it will be translated to `-1` in Dart, and translated back to the correct value when its back in Rust. As i previously mentioned it is `toUnsigned(64).toInt()` that is the problem: 

```
0xffffffffffffffff.toInt() = -1
0xffffffffffffffff.b.toUnsigned(64) = 18446744073709551615
0xffffffffffffffff.b.toUnsigned(64).toInt() = 9223372036854775807
```

fzyzcjy:
> So i've run some tests and returning DartOpaque from Rust to Dart works fine using dco even without this PR, so its not affected.

Great!

> Meh, i dont think there is an easy way to do this.

I also think so... The environment is not trivial to create, and the unit tests may not cover wide enough.

> it will be translated to -1 in Dart, and translated back to the correct value when its back in Rust

Yes, I guess it would be great for the doc to mention something about that (though surely optional)

(I will review the PR in detail probably tonight)

fzyzcjy (commented):
> So, i believe this fix can be reworked to continue using usize for pointers, so let me know if you think this is the right way.

I guess that would have some benefits:

* The address is unsigned, so usize looks intuitive
* Other parts of the code often use usize (not isize) for addresses, so this aligns with those parts
* Also, PR size is smaller, making the change clearer

fzyzcjy:
Btw, feel free to ping me (or use `request a review` button) when ready!

fzyzcjy:
Great! I will probably review within 24hours

frb_codegen/src/library/codegen/ir/mir/ty/dart_opaque.rs
```diff
@@ -11,7 +11,8 @@ impl MirTypeTrait for MirTypeDartOpaque {
         f: &mut F,
         mir_context: &impl MirContext,
     ) {
-        self.get_delegate().visit_types(f, mir_context)
+        MirType::Primitive(MirTypePrimitive::Isize).visit_types(f, mir_context);
+        MirType::Primitive(MirTypePrimitive::Usize).visit_types(f, mir_context);
```
fzyzcjy, line 15: nit: what about

1. Add method `MirTypeDartOpaque.get_delegate_rust` which gives usize, `get_delegate_dart` which gives isize
2. At here (visit types), we call `self.get_delegate_rust().visit_types(); self.get_delegate_dart().visit_types();`
3. At places mentioned below, call get_delegate_dart/get_delegate_rust, instead of hardcoded isize/usize

frb_codegen/src/library/codegen/generator/codec/sse/ty/dart_opaque.rs
```diff
@@ -5,28 +5,36 @@ use crate::codegen::generator::codec::sse::ty::*;
...
-                    "BigInt.from(PlatformPointerUtil.ptrToInt(encodeDartOpaque(self, portManager.dartHandlerPort, generalizedFrbRustBinding))).toUnsigned(64)"
-                }
-                Lang::RustLang(_) => "self.encode()",
-            },
-        ))
+        Some(match lang {
+            Lang::DartLang(_) => {
+                simple_delegate_encode(
+                    lang,
+                    &MirType::Primitive(MirTypePrimitive::Isize),
```
fzyzcjy, line 12: here

frb_codegen/src/library/codegen/generator/codec/sse/ty/dart_opaque.rs
```diff
@@ -5,28 +5,36 @@ use crate::codegen::generator::codec::sse::ty::*;
...
+                    "PlatformPointerUtil.ptrToPlatformInt64(encodeDartOpaque(self, portManager.dartHandlerPort, generalizedFrbRustBinding))",
+                )
+            }
+            Lang::RustLang(_) => {
+                simple_delegate_encode(
+                    lang,
+                    &MirType::Primitive(MirTypePrimitive::Usize),
```
fzyzcjy, line 19: here

frb_codegen/src/library/codegen/generator/codec/sse/ty/dart_opaque.rs
```diff
@@ -5,28 +5,36 @@ use crate::codegen::generator::codec::sse::ty::*;
...
-                Lang::DartLang(_) => "decodeDartOpaque(inner, generalizedFrbRustBinding)",
-                Lang::RustLang(_) => {
-                    "unsafe { flutter_rust_bridge::for_generated::sse_decode_dart_opaque(inner) }"
-                }
-            },
-        ))
+        Some(match lang {
+            Lang::DartLang(_) => simple_delegate_decode(
+                lang,
+                &MirType::Primitive(MirTypePrimitive::Isize),
```
fzyzcjy, line 30: here

frb_codegen/src/library/codegen/generator/codec/sse/ty/dart_opaque.rs
```diff
@@ -5,28 +5,36 @@ use crate::codegen::generator::codec::sse::ty::*;
...
+                "decodeDartOpaque(inner, generalizedFrbRustBinding)",
+            ),
+            Lang::RustLang(_) => simple_delegate_decode(
+                lang,
+                &MirType::Primitive(MirTypePrimitive::Usize),
```
fzyzcjy, line 35: here

fzyzcjy (approved):
Great job! Only a nit (other 4 are pointers to point out specific line) and I am ready to merge

fzyzcjy:
Great job!

fzyzcjy:
@all-contributors please add @alexlapa for code

PR #2284, opened 2024-09-08
Add commandline option for passing features to cargo expand during codegen

### LINKED ISSUE #2283: Passing feature flags to flutter_rust_bridge_codegen is not possible
I have a crate that has a "flutter" feature, than when enabled adds the required dependencies and enables some flutter specific apis. This is because my crate is designed to be used both with and without flutter. Since feature flags are suppose to be additive I can't have it enabled by default. 

Have a commandline argument and/or a config option to enable features (pass --features to cargo expand) during codegen

Alternatives include making features default (and therefore subtractive) and using a dedicated 2nd crate. Because RUSTFLAGS isn't passed to cargo expand alone in the right place,

```
 RUSTFLAGS="--features flutter" flutter_rust_bridge_codegen generate
 ```
also fails with  `error: Unrecognized option: 'features'`

**Additional context**
I have a branch https://github.com/fmeef/flutter_rust_bridge/tree/cargo_features that adds this, making PR soon

### PR DESCRIPTION
**Pull request by fmeef**
This fixes #2283

### CODE REVIEW

File: frb_codegen/src/binary/commands.rs [hunk 1]
```diff
@@ -172,6 +172,10 @@ pub(crate) struct GenerateCommandArgsPrimary {
     /// Dump all internal data. Same as `--dump` with all possible choices chosen.
     #[arg(long)]
     pub dump_all: bool,
+
+    /// List of cargo feature flags to enable when generating
+    #[arg(short = 'F', long)]
```
fzyzcjy, line 177: nit: maybe we can allocate `short` flags later to avoid potential collision (the `long` may be enough for today)

fzyzcjy (commented):
(just a quick one-line nit when skimming it)

fzyzcjy:
Looks like there is a 10k line of code diff, possibly because some cargokit folders etc are missing. One quick way to fix this error is to revert unwanted changes in those folders.

Feel free to ping me if you need any suggestions to make CI pass etc, or ping me when it is ready for review!

fmeef:
Yes, I just realized that happened when running `./frb_internal precommit --mode slow`, fixing it.

fzyzcjy:
The code LGTM, maybe we can add a test for it. For example, maybe we can 

* Add a function that only compiles when a custom feature is enabled, in https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_example/pure_dart/rust/src/api/misc_no_twin_example_a.rs
* Modify https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/tools/frb_internal/lib/src/makefile_dart/generate.dart#L302 such that it provides that feature flag
* We should see the function has corresponding dart function generated
* You can test it by `./frb_internal generate-run-frb-codegen-command-generate --package frb_example/pure_dart`

fmeef:
While I have these tests added, I am trying to find a way to make linting/clippy respect the correct feature flags in a clean way. Will let you know when its fully working again

fmeef:
Appears to be fully working again with the tests

frb_codegen/src/binary/commands.rs
```diff
@@ -172,6 +172,10 @@ pub(crate) struct GenerateCommandArgsPrimary {
     /// Dump all internal data. Same as `--dump` with all possible choices chosen.
     #[arg(long)]
     pub dump_all: bool,
+
+    /// List of cargo feature flags to enable when generating
+    #[arg(long)]
+    pub features: Option<Vec<String>>,
```
fzyzcjy, line 178: nit: Shall we rename it to something like `rust_features`? For example, suppose one day in the future Dart supports features as well, then `features` may be confusing while `rust_features` is more understandable.

frb_codegen/src/library/codegen/parser/internal_config.rs
```diff
@@ -6,4 +6,5 @@ use serde::{Deserialize, Serialize};
 pub(crate) struct ParserInternalConfig {
     pub hir: ParserHirInternalConfig,
     pub mir: ParserMirInternalConfig,
+    pub features: Option<Vec<String>>,
```
fzyzcjy, line 9: nit: I guess it belongs to `ParserHirInternalConfig`

frb_codegen/src/library/codegen/parser/mod.rs
```diff
@@ -34,7 +34,7 @@ fn parse_inner(
     let dumper_mir = dumper.with_content(Mir);
 
     let pb = progress_bar_pack.parse_hir_raw.start();
-    let hir_raw = hir::raw::parse(&config.hir, dumper)?;
+    let hir_raw = hir::raw::parse(&config.hir, dumper, config.features.as_deref())?;
```
fzyzcjy, line 37: ... then we do not need to add this extra arg, since it is already in `config.hir`

frb_codegen/src/library/commands/cargo_expand/real.rs
```diff
@@ -55,16 +58,24 @@ fn run_raw(
...
     } else {
         vec![]
     };
 
+    if let Some(features) = features {
+        for feature in features {
+            args_choosing_crate.push("--features");
+            args_choosing_crate.push(feature.as_str());
+        }
+    }
```
fzyzcjy, line 77: Looks like this is not `args_choosing_crate` - it is more like `args_features`. So maybe

```rust
let args_features = features.unwrap_or_default().iter()
.flat_map(|feature| vec!["--features", feature])
.collect_vec();
```

frb_example/pure_dart/build.dart
```diff
@@ -1,3 +1,3 @@
 import 'package:flutter_rust_bridge_utils/flutter_rust_bridge_utils.dart';
 
-void main(List<String> args) async => simpleBuild(args);
+void main(List<String> args) async => simpleBuild(args, features: ['test']);
```
fzyzcjy, line 3:
```suggestion
void main(List<String> args) async => simpleBuild(
  args, 
  // This is solely used for testing
  features: ['test'],
);
```

frb_example/pure_dart/flutter_rust_bridge.yaml
```diff
@@ -19,3 +19,5 @@ enable_lifetime: true
 dump_all: true
 # only for development of flutter_rust_bridge, not for normal users (use local dependency)
 local: true
+features:
+  - test
```
fzyzcjy, line 23: nit: maybe we can rename this feature to something like `this_feature_is_for_test` or something else. because the current `test` may make readers confused - does it mean "enable this feature when we are running unit tests of this package"

frb_example/pure_dart/rust/src/api/misc_no_twin_example_a.rs
```diff
@@ -218,3 +218,15 @@ pub struct StructWithRustAutoOpaqueWithNonCloneData {
...
+#[allow(non_snake_case)] // There is a bug in spec_generator that requires camel case
+pub fn getFeatureTest() -> String {
+    "test".to_owned()
+}
+
+#[cfg(not(feature = "test"))]
+#[allow(non_snake_case)] // There is a bug in spec_generator that requires camel case
+pub fn getFeatureTest() -> String {
+    "nottest".to_owned()
+}
```
fzyzcjy, line 232: issue: hmm, if we have exactly the same function when features are enabled or disabled, then how can we test whether our code generator are using the correct feature? i.e. suppose our code generator does not handle feature flags at all, then it will still output the exact same thing!

frb_example/pure_dart/rust/src/api/misc_no_twin_example_a.rs
```diff
@@ -218,3 +218,15 @@ pub struct StructWithRustAutoOpaqueWithNonCloneData {
 impl StructWithRustAutoOpaqueWithNonCloneData {
     pub fn f(&self) {}
 }
+
+#[cfg(feature = "test")]
+#[allow(non_snake_case)] // There is a bug in spec_generator that requires camel case
```
fzyzcjy, line 223: That looks weird... could you please elaborate a bit more?

frb_example/pure_dart/test/api/feature_test.dart
```diff
@@ -0,0 +1,13 @@
+// FRB_INTERNAL_GENERATOR: {"forbiddenDuplicatorModes": ["rustAsync", "sse", "sync sse", "rustAsync sse"], "replaceCode": {"pseudo_manual/misc_no_twin_example_a_twin_sync": "misc_no_twin_example_a"}}
+
+import 'package:frb_example_pure_dart/src/rust/frb_generated.dart';
+import 'package:test/test.dart';
+import 'package:frb_example_pure_dart/src/rust/api/misc_no_twin_example_a.dart';
+
+Future<void> main({bool skipRustLibInit = false}) async {
```
fzyzcjy, line 7: nit: since you are using misc_no_twin_example_a.rs, just write test code in misc_no_twin_example_a_test.dart (instead of the new file feature_test.dart)

frb_utils/lib/src/dart_web_test_utils/run_test.dart
```diff
@@ -22,6 +22,14 @@ Future<void> executeTestWeb(TestWebConfig config) async {
   final webRoot = '$dartRoot/web';
   print('executeTestWeb: Pick dartRoot=$dartRoot');
 
+  List<String> cargoArgs = [];
+
+  if (dartRoot.endsWith('frb_example/pure_dart') ||
+      dartRoot.endsWith('frb_example/pure_dart_pde')) {
+    print("running wtih feature flag test");
```
fzyzcjy, line 29:
```suggestion
    print("running with feature flag test");
```

frb_utils/lib/src/simple_build_utils.dart
```diff
@@ -27,13 +27,19 @@ void simpleBuild(List<String> args) async {
     print(
         'frb_utils::simpleBuild SKIP BUILD since environment variable requires this');
   } else {
+    List<String> f = [];
+    for (String feat in features) {
+      f.add('--features');
+      f.add(feat);
+    }
```
fzyzcjy, line 34:
```suggestion
    final featureArgs = features.expand((x) => ['--features', x]).toList();
```

tools/frb_internal/lib/src/makefile_dart/consts.dart
```diff
@@ -36,8 +36,6 @@ const kDartExampleIntegratePackages = [
 
 const kDartExamplePackages = [
   'frb_example/dart_minimal',
-  'frb_example/pure_dart',
-  'frb_example/pure_dart_pde',
   'frb_example/dart_build_rs',
```
fzyzcjy, line 39: Hmm, maybe we can keep kDartExamplePackages unchanged

tools/frb_internal/lib/src/makefile_dart/generate.dart
```diff
@@ -371,18 +375,25 @@ Future<void> generateRunFrbCodegenCommandIntegrate(
...
+    if (cmd == "generate") {
+      String f = features.map((f) {
+        return '--features $f';
+      }).join(' ');
+      return await exec('flutter_rust_bridge_codegen $cmd $f',
+          relativePwd: relativePwd);
+    } else {
+      return await exec('flutter_rust_bridge_codegen $cmd',
+          relativePwd: relativePwd);
+    }
```
fzyzcjy, line 396: maybe we can keep this function unchanged. Instead, when users want to add features, they can provide `cmd` argument with corresponding extra string

tools/frb_internal/lib/src/makefile_dart/misc.dart
```diff
@@ -82,6 +82,12 @@ Future<void> precommitGenerate() async {
     for (final package in kDartExamplePackages)
       generateRunFrbCodegenCommandGenerate(GeneratePackageConfig(
           setExitIfChanged: false, package: package, coverage: false)),
+    for (final package in kDartExampleTestFeaturePackages)
+      generateRunFrbCodegenCommandGenerate(GeneratePackageConfig(
```
fzyzcjy, line 86: what about this: Do not add a `GeneratePackageConfig.features` flag (since when we are running this codegen test, we *always* want to have that feature enabled). Instead, inside generateRunFrbCodegenCommandGenerate, maybe we just check if package == pure_dart/pure_dart_pde, and add feature flags accordingly

tools/frb_internal/lib/src/makefile_dart/release.dart
```diff
@@ -103,6 +103,11 @@ void _updateVersionInText() {
           '$x/rust/src/**',
           '$x/../src/**',
         ]),
+    ...kDartExampleTestFeaturePackages.expand((x) => [
```
fzyzcjy, line 106: (same, maybe we can remove kDartExampleTestFeaturePackages variable and its special treatment)

fzyzcjy (commented):
Great job! Mainly some nits :)

fmeef:
Do these latest changes work so far?

fzyzcjy:
Looks great! I will review again within a day. Btw next time when you are ready, feel free to use "request a review" button or tell me.

fmeef:
Oh yes sorry will use that next time

fzyzcjy:
It's OK!

fzyzcjy:
Reviewed half, and then realize it seems that the discussions in 

image

are not solved. Feel free to ping me after these are handled as well!

tools/frb_internal/lib/src/makefile_dart/consts.dart
```diff
@@ -36,9 +36,9 @@ const kDartExampleIntegratePackages = [
 
 const kDartExamplePackages = [
   'frb_example/dart_minimal',
+  'frb_example/dart_build_rs',
```
fzyzcjy, line 39: nit: let's keep this unchanged

image

tools/frb_internal/lib/src/makefile_dart/generate.dart
```diff
@@ -303,11 +303,17 @@ Future<void> generateRunFrbCodegenCommandGenerate(
     GeneratePackageConfig config) async {
   await _wrapMaybeSetExitIfChanged(config, () async {
     await runPubGetIfNotRunYet(config.package);
+    print("generating with ${config.package}");
+    List<String> features = [];
+    if (config.package == "pure_dart/pure_dart_pde") {
```
fzyzcjy, line 308: looks like this is not very correct...

tools/frb_internal/lib/src/makefile_dart/generate.dart
fzyzcjy: (this is not handled)

tools/frb_internal/lib/src/makefile_dart/generate.dart
fmeef: I think I see what is happening, I used the same function name for both enabled and disabled feature. Since the rust lib is compiled with enabled feature it returns "test", which passes the test, but in reality the feature is disabled for codegen. Will fix this shortly

fmeef: Actually I dont think thats entirely a correct guess, tests are still passing even after removing that function. I am going to see if I can solve it by calling the code generator during the test itself?

fmeef: The issue here is that requesting a feature in `generateRunFrbCodegenCommandGenerate()` is essentially a no-op because the same feature is already requested from the flutter_rust_bridge.yaml file. 

```yaml
rust_features:
  - internal_feature_for_testing
```
removing this line, along with removing the duplicate `getFeatureTest()` function in the rust code causes tests to fail as expected. I am going to remove the redundant feature handling in generate.dart, etc and move to purely passing features by flutter_rust_bridge.yaml if that is ok

fzyzcjy:
Feel free to ping me when the things above are handled!

fmeef:
seems to be ready

frb_example/pure_dart/rust/src/api/misc_no_twin_example_a.rs
fmeef: I noticed that in frb_generated.rs it sometimes swaps between camel case and snake case in ways that cause compilation failures. I can try to reproduce it again shortly.

fmeef: https://github.com/fmeef/flutter_rust_bridge/actions/runs/10938682186
this is an example of the issues that arise with camel case. Note the only change I made is switching `getFeatureTest()` to `get_feature_test()` and reran codegen. Correct me if I am wrong, but those should be equivalent in the generated code right?

```
error[E0425]: cannot find function `getFeatureTest` in module `crate::api::misc_no_twin_example_a`
    --> src/frb_generated.rs:7105:81
     |
7105 |                         Result::<_, ()>::Ok(crate::api::misc_no_twin_example_a::getFeatureTest())?;
     |                                                                                 ^^^^^^^^^^^^^^ help: a function with a similar name exists: `get_feature_test`
     |
    ::: src/api/misc_no_twin_example_a.rs:223:1
     |
223  | pub fn get_feature_test() -> String {
     | ----------------------------------- similarly named function `get_feature_test` defined here

For more information about this error, try `rustc --explain E0425`.
```

fzyzcjy:
Great! I will review probably within a day

tools/frb_internal/lib/src/makefile_dart/consts.dart
```diff
@@ -38,7 +38,6 @@ const kDartExamplePackages = [
   'frb_example/dart_minimal',
   'frb_example/pure_dart',
   'frb_example/pure_dart_pde',
-  'frb_example/dart_build_rs',
```
fzyzcjy, line 41: Hmm why this line disappears

tools/frb_internal/lib/src/makefile_dart/misc.dart
```diff
@@ -111,3 +111,8 @@ String convertConfigPackage(String raw) {
   if (raw != ans) print('convertConfigPackage: $raw -> $ans');
   return ans;
 }
+
+bool isFeature(String package) {
```
fzyzcjy, line 115: nit: maybe we can give it a more descriptive name

nit2: what about `String getRustFeaturesOfPackage(String package)` and we return the features that is needed

tools/frb_internal/lib/src/makefile_dart/lint.dart
```diff
@@ -55,17 +56,18 @@ Future<void> lintRustFormat(LintConfig config) async {
...
     await exec(
-        'cargo clippy ${config.fix ? "--fix --allow-dirty --allow-staged" : ""} -- -D warnings',
+        'cargo clippy ${isFeature(package) ? "--features internal_feature_for_testing" : ""} ${config.fix ? "--fix --allow-dirty --allow-staged" : ""} -- -D warnings',
         relativePwd: package);
   }
 
   for (final package in kRustPackagesAllowWeb) {
     await exec(
-        'cargo clippy --target wasm32-unknown-unknown ${config.fix ? "--fix --allow-dirty --allow-staged" : ""} -- -D warnings',
+        'cargo clippy ${(package == "frb_example/pure_dart_pde/rust" || package == "frb_example/pure_dart/rust") ? "--features internal_feature_for_testing" : ""} --target wasm32-unknown-unknown ${config.fix ? "--fix --allow-dirty --allow-staged" : ""} -- -D warnings',
```
fzyzcjy, line 70: nit: use the `isFeature` function instead of hardcode

frb_example/pure_dart/rust/src/api/misc_no_twin_example_a.rs
fzyzcjy: That's weird... I have not seen anyone reporting their function has case problems...

In case it is issue with prefix `get_`, what about renaming the function (which also makes it more descriptive)

for example: `fn function_that_only_exists_under_feature_flag() { ... }`

fzyzcjy (commented):
Great! Ready to merge after nits

frb_utils/lib/src/dart_web_test_utils/run_test.dart
```diff
@@ -22,6 +22,13 @@ Future<void> executeTestWeb(TestWebConfig config) async {
   final webRoot = '$dartRoot/web';
   print('executeTestWeb: Pick dartRoot=$dartRoot');
 
+  List<String> cargoArgs = [];
+
+  if (dartRoot.endsWith('frb_example/pure_dart') ||
+      dartRoot.endsWith('frb_example/pure_dart_pde')) {
+    cargoArgs = ['--features', 'internal_feature_for_testing'];
+  }
+
```
fzyzcjy, line 31: nit: not sure why I missed this... the `executeTestWeb` seems to be general-purpose, thus it is great to avoid hardcode such things here.
For example, make it an argument in TestWebConfig

tools/frb_internal/lib/src/makefile_dart/lint.dart
```diff
@@ -54,18 +55,21 @@ Future<void> lintRustFormat(LintConfig config) async {
 
 Future<void> lintRustClippy(LintConfig config) async {
   for (final package in kRustPackages) {
+    String? feature = getRustFeaturesOfPackage(package);
```
fzyzcjy, line 58:
```suggestion
    final feature = getRustFeaturesOfPackage(package);
```

tools/frb_internal/lib/src/makefile_dart/lint.dart
```diff
@@ -54,18 +55,21 @@ Future<void> lintRustFormat(LintConfig config) async {
...
           relativePwd: package);
     }
     await exec(
-        'cargo clippy ${config.fix ? "--fix --allow-dirty --allow-staged" : ""} -- -D warnings',
+        'cargo clippy ${feature != null ? "--features $feature" : ""} ${config.fix ? "--fix --allow-dirty --allow-staged" : ""} -- -D warnings',
         relativePwd: package);
   }
 
   for (final package in kRustPackagesAllowWeb) {
+    String? feature = getRustFeaturesOfPackage(package);
```
fzyzcjy, line 70:
```suggestion
    final feature = getRustFeaturesOfPackage(package);
```

(you can use github and click the button to quickly accept this tiny change)

fzyzcjy (approved):
Great! I am ready to merge and only one missing nit (other two can be fixed by using github's button)

fzyzcjy:
@all-contributors please add @fmeef for code

frb_utils/lib/src/commands/test_web_command.dart
```diff
@@ -24,4 +24,7 @@ class TestWebConfig {
 
   @CliOption(help: 'Make the browser headless', defaultsTo: true)
   late bool headless;
+
+  @CliOption(help: 'Rust feature flags to set during build')
+  late List<String> rustfeatures;
```
fzyzcjy, line 29: nit: I would like to merge it, but this indeed should be `rustFeatures`, and the cli will automatically be `rust-features`....

fmeef (commented):
Fixed

PR #2303, opened 2024-09-16
Logging overhaul

patmuk:
## Changes

_Please list issues fixed by this PR here, using format "Fixes #the-issue-number"._

## Checklist

- [ ] An issue to be fixed by this PR is listed above.
- [ ] New tests are added to ensure new features are working. Please refer to [this page](https://cjycode.com/flutter_rust_bridge/guides/contributing/tip) to see how to add a test.
- [ ] `./frb_internal precommit --mode slow` (or `fast`) is run (it internal runs code generator, does auto formatting, etc).
- [ ] If this PR adds/changes features, documentations (in the `./website` folder) are updated.
- [ ] CI is passing. Please refer to [this page](https://cjycode.com/flutter_rust_bridge/guides/contributing/tip) to see how to solve a failed CI.

## Remark for PR creator

- `./frb_internal --help` shows utilities for development.
- If fzyzcjy does not reply for a few days, maybe he just did not see it, so please ping him.

PR #2308, opened 2024-09-20
Logging Overhaul

patmuk:
## Changes

## Checklist

## Remark for PR creator


patmuk:
@fzyzcjy I cleaned up the branch ... that is, I needed to create a new branch. As I could not swap the branch from the old PR (#2303) I created this new one.

patmuk:
I solved the dependency issue, but I have a code generation issue now:

In the macro code, in `frb_rust/src/log_2_dart.rs` I get:
```
  --> /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_rust/src/log_2_dart.rs:81:17
   |
81 |             use frb_generated::StreamSink;
   |                 ^^^^^^^^^^^^^ use of undeclared crate or module `frb_generated`
```

Clearly, in this code I have not generated code. But I need to refer to the StreamSink, which should be generated ...

Do you have an advice how to progress?
1. I could copy the generated code for the StreamSink to the macro as well - but this way we are close (back?) to the non-macro way in PR #2303 
2. I could probably inject it as a parameter of the macro ... but then the StreamSink needs to be defined in the users project - not a very clean solution
3. statically calling code generation? 

I am sure you have a better idea :) It is a bit cat-and-mouse: The logger code should be generated (actually integrated), but it needs the StreamSink to be generated to work ...

fzyzcjy:
> In the macro code

Standard Rust would need `crate::frb_generated::...`

patmuk:
> Standard Rust would need `crate::frb_generated::...` to find the `frb_geneated.rs`

The problem is, that there is no `struct StreamSink` declared yet. It is generated later.

```
error[E0432]: unresolved import `crate::frb_generated`
 --> /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_rust/src/log_2_dart.rs:4:12
  |
4 | use crate::frb_generated::StreamSink;
  |            ^^^^^^^^^^^^^ could not find `frb_generated` in the crate root
error[E0412]: cannot find type `StreamSink` in this scope
  --> /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_rust/src/log_2_dart.rs:87:30
   |
87 |                 stream_sink: StreamSink<Log2DartLogRecord>,
   |                              ^^^^^^^^^^ not found in this scope
   |
  ::: /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_rust/src/misc/user_utils.rs:10:5
   |
10 |     setup_logging!();
   |     ---------------- in this macro invocation
   |
   = note: this error originates in the macro `setup_logging` (in Nightly builds, run with -Z macro-backtrace for more info)
error[E0412]: cannot find type `StreamSink` in this scope
  --> /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_rust/src/log_2_dart.rs:92:29
   |
92 |                 log_stream: StreamSink<Log2DartLogRecord>,
   |                             ^^^^^^^^^^ not found in this scope
   |
  ::: /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_rust/src/misc/user_utils.rs:10:5
   |
10 |     setup_logging!();
   |     ---------------- in this macro invocation
   |
   = note: this error originates in the macro `setup_logging` (in Nightly builds, run with -Z macro-backtrace for more info)
Some errors have detailed explanations: E0412, E0432.
For more information about an error, try `rustc --explain E0412`.
```

patmuk:
I progressed a bit, but have the same problem.

I moved my macro `setup_logging` to `frb_rust/src/for_generated/boilerplate.rs`.
The `StreamSink` dependency resolves, but SSE isn't generated (yet):
```
error[E0599]: the method `add` exists for struct `StreamSink<Log2DartLogRecord>`, but its trait bounds were not satisfied
  --> src/frb_generated.rs:34:1
   |
34 | / flutter_rust_bridge::frb_generated_boilerplate!(
35 | |     default_stream_sink_codec = SseCodec,
36 | |     default_rust_opaque = RustOpaqueMoi,
37 | |     default_rust_auto_opaque = RustAutoOpaqueMoi,
38 | | );
   | | ^
   | | |
   | | method cannot be called on `StreamSink<Log2DartLogRecord>` due to unsatisfied trait bounds
   | |_doesn't satisfy `Log2DartLogRecord: SseEncode`
   |   method `add` not found for this struct
   |
note: trait bound `Log2DartLogRecord: SseEncode` was not satisfied
  --> src/frb_generated.rs:34:1
   |
34 | / flutter_rust_bridge::frb_generated_boilerplate!(
35 | |     default_stream_sink_codec = SseCodec,
36 | |     default_rust_opaque = RustOpaqueMoi,
37 | |     default_rust_auto_opaque = RustAutoOpaqueMoi,
38 | | );
   | | ^ unsatisfied trait bound introduced here
   | |_|
   |
note: the trait `SseEncode` must be implemented
  --> src/frb_generated.rs:34:1
   |
34 | / flutter_rust_bridge::frb_generated_boilerplate!(
35 | |     default_stream_sink_codec = SseCodec,
36 | |     default_rust_opaque = RustOpaqueMoi,
37 | |     default_rust_auto_opaque = RustAutoOpaqueMoi,
38 | | );
   | |_^
   = help: items from traits can only be used if the trait is implemented and in scope
   = note: the following trait defines an item `add`, perhaps you need to implement it:
           candidate #1: `Add`
   = note: this error originates in the macro `$crate::setup_logging` which comes from the expansion of the macro `flutter_rust_bridge::frb_generated_boilerplate` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Is there an even later place to call the macro? Somehow it looks like the codegen would have to run twice, first to expand the macros and then to generate the code for the `StreamSink<Log2DartLogRecord>`?

fzyzcjy:
> The problem is, that there is no struct StreamSink declared yet. It is generated later.

Hmm that's weird: Then how can users have something like 

```
fn f(a: StreamSink<i32>) {}
```

in their code (e.g. in frb_example/dart_minimal/simple.rs)?

If users can do that, then we may also be able to do that, since calling a macro is nothing but copy-paste some code into frb_example/dart_minimal

fzyzcjy:
> Is there an even later place to call the macro? 

In `frb_example/dart_minimal/simple.rs`. Just like we generate a `fn init_app() {}`, we should generate a `your_macro_name_for_logging!();`.

We deliberately do this, because by doing so, users can very easily modify/customize/disable the default logging solution, by e.g. changing that line to `your_macro_name_for_logging!(their_custom_args);` or remove that line. This is also why I choose to generate the init app function.

patmuk:
> > Is there an even later place to call the macro?
> 
> In `frb_example/dart_minimal/simple.rs`. Just like we generate a `fn init_app() {}`, we should generate a `your_macro_name_for_logging!();`.
> 
> We deliberately do this, because by doing so, users can very easily modify/customize/disable the default logging solution, by e.g. changing that line to `your_macro_name_for_logging!(their_custom_args);` or remove that line. This is also why I choose to generate the init app function.

I put the macro call (setup_logging!();`) into `frb_example/dart_minimal/rust/src/api/minimal.rs`, but get the same error
<details>
  <summary>Click me</summary>
  ```
error[E0599]: the method `add` exists for struct `StreamSink<Log2DartLogRecord>`, but its trait bounds were not satisfied
  --> src/api/minimal.rs:8:1
   |
8  |   #[frb(init)]
   |   ^^^^^^^^^^^^
   |   |
   |   method cannot be called on `StreamSink<Log2DartLogRecord>` due to unsatisfied trait bounds
   |   doesn't satisfy `Log2DartLogRecord: SseEncode`
   |
  ::: src/frb_generated.rs:34:1
   |
34 | / flutter_rust_bridge::frb_generated_boilerplate!(
35 | |     default_stream_sink_codec = SseCodec,
36 | |     default_rust_opaque = RustOpaqueMoi,
37 | |     default_rust_auto_opaque = RustAutoOpaqueMoi,
38 | | );
   | |_- method `add` not found for this struct
   |
note: trait bound `Log2DartLogRecord: SseEncode` was not satisfied
  --> src/frb_generated.rs:34:1
   |
34 | / flutter_rust_bridge::frb_generated_boilerplate!(
35 | |     default_stream_sink_codec = SseCodec,
36 | |     default_rust_opaque = RustOpaqueMoi,
37 | |     default_rust_auto_opaque = RustAutoOpaqueMoi,
38 | | );
   | | ^ unsatisfied trait bound introduced here
   | |_|
   |
note: the trait `SseEncode` must be implemented
  --> src/frb_generated.rs:34:1
   |
34 | / flutter_rust_bridge::frb_generated_boilerplate!(
35 | |     default_stream_sink_codec = SseCodec,
36 | |     default_rust_opaque = RustOpaqueMoi,
37 | |     default_rust_auto_opaque = RustAutoOpaqueMoi,
38 | | );
   | |_^
   = help: items from traits can only be used if the trait is implemented and in scope
   = note: the following trait defines an item `add`, perhaps you need to implement it:
           candidate #1: `Add`
   = note: this error originates in the macro `setup_logging` which comes from the expansion of the macro `flutter_rust_bridge::frb_generated_boilerplate` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0599`.
error: could not compile `frb_example_dart_minimal` (lib) due to 1 previous error
Unhandled exception:
ProcessException: Bad exit code (101). If you want to see extra information, set FRB_DART_RUN_COMMAND_STDERR=1
  Command: cargo build --release
#0      runCommand (package:flutter_rust_bridge/src/cli/run_command.dart:67:5)
<asynchronous suspension>
#1      simpleBuild (package:flutter_rust_bridge_utils/src/simple_build_utils.dart:30:5)
<asynchronous suspension>

stdout:

        
Error: Compiling native assets failed.
  ```
</details> 

I tried different places (inside `init_app`, inside `minimal_adder`, nowhere, but annotated with `#[frb(init)]`), but I am getting the same error message.

patmuk:
When I copy the rust code from the macro directly into `dart_minimal/minimal.rs` the code generator creates the SSE code and the StreamSink.

Maybe the macro get's expanded, but the code inside not parsed by the generator?

fzyzcjy:
Hmm, could you please make a minimal reproducible sample? I guess there may be some problems in some details. Anyway, other rust macros seem to work well. @MnlPhlp 's https://github.com/MnlPhlp/flutter_logger also uses macro and also works.

patmuk:
> Hmm, could you please make a minimal reproducible sample?

Here is the minimal example:
https://github.com/patmuk/flutter_rust_bridge/tree/logging_macro_minimal

I figured that removing the Log trait implementation:
` impl log::Log for Log2Dart {`
makes it work! 

So, somehow the trait gets into the way of code generation:
```
            #[frb(non_opaque)]
            pub struct Log2Dart {
                stream_sink: StreamSink<Log2DartLogRecord>,
            }
```

```
   |   method cannot be called on `StreamSink<Log2DartLogRecord>` due to unsatisfied trait bounds
   |   doesn't satisfy `Log2DartLogRecord: SseEncode`
```

I might be able to workaround by having the `StreamSink` not be a field of `Log2Dart`, which implements the log trait. But I think there is a bug here that might hit somebody in a similar constellation.

fzyzcjy:
Hmm I cannot see macro call in https://github.com/patmuk/flutter_rust_bridge/blob/logging_macro_minimal/frb_example/dart_minimal/rust/src/api/minimal.rs. Could you please point out more details?

patmuk:
I reconstructed the macro code into the `dart_minimal` project - and realized that some imports are missing ... and that I initialized it from the wrong side: Instead of from the rust code the initialization has to happen from the dart code `init_logger();` - which is actually the function you wanted to have (taking a log level, custom log function, and not logging anything if not called).

So, I am onto something :)

However, I am wondering where the dart code from [`#frb[dart_code=]`](https://cjycode.com/flutter_rust_bridge/guides/misc-features/dart-code) ends up? In frb_generated.dart or minimal.dart (or similar)? I can't find it ... but I assume that I call it on the wrong place:

in the macro definition I had 
```
#[macro_export]
macro_rules! setup_logging {
    () => {
        #[frb(dart_code = "
(... dart code ...)
")]
pub fn rust_logger_init(){
(... rust code ...)
}
```

but I think `pub fn rust_logger_init()` is problematic, somehow hindering the StreamSink sse code generation.
But I have to put something there, right? Just the code, starting with `use ...` looks not to work, as I can't find the generated dart code. I tried to put it into a `mod` but still no dart code.

fzyzcjy:
> but I think pub fn rust_logger_init() is problematic,

I cannot get it...

Anyway, for simplicity, maybe you can firstly ignore the macro problem and do everything directly. Only after we are 100% done, let's move to macro.

patmuk:
I made a [minimal example](https://github.com/patmuk/flutter_rust_bridge/tree/minimal_example_dart_code_generation) for dart_code generation - where I don't see the generated code :/ Could you please check what I am doing wrong there?

patmuk:
> > but I think pub fn rust_logger_init() is problematic,
> 
> I cannot get it...

Never mind, I will describe it better once it works. I made a minimal example of my problem with `#[frb(dart_code = )]`.

> 
> Anyway, for simplicity, maybe you can firstly ignore the macro problem and do everything directly. Only after we are 100% done, let's move to macro.

I had it already working without the macro - but with the dart code in the main.dart instead of the dart_code attribute.
But yes, I agree - I currently moved the code back to minimal.rs. 

Once I know what I did wrong I will do it step-by-step, first dart_code without macro, then moving to a macro.

patmuk:
FYI: commit https://github.com/fzyzcjy/flutter_rust_bridge/pull/2308/commits/d1ca1273747e915cafd540777da60189b88f900b is a working version, where all code is in `dart_minimal`. Now modifying from there ...

patmuk:
I added 
```rust
#[frb(dart_code = "
void sayit() => print('where will this be?');
")]
pub struct Teller {
    pub a: i32,
}
```
to `minimal.rs` (not committed) but don't see the dart_code output anywhere.

patmuk:
I just saw that the `dart_code` is removed, if the struct it is annotating is not used.

Can one only add the `#[frb(dart_clone= )]` attribute to structs? Not to mods?

patmuk:
> Anyway, for simplicity, maybe you can firstly ignore the macro problem and do everything directly. Only after we are 100% done, let's move to macro.

Here is the minimal example:
https://github.com/patmuk/flutter_rust_bridge/tree/logging_macro_minimal

The code is in frb_examples/dart_minimal/rust/src/api/minimal.rs.

I reverted back to a state without macros.
Here I try to use #[frb(dart_code= )] to inline the dart code - but no code is added in the generated output.

I made a [minimal example](https://github.com/patmuk/flutter_rust_bridge/tree/minimal_example_dart_code_generation) of my problem with #[frb(dart_code = )].

I realized that the dart_code is removed, if the struct it is annotating is not used.

Can one only add the #[frb(dart_clone= )] attribute to structs? Not to mods?

patmuk:
Update before you answer: I restructured the code, so the `#[frb(dart_code = )]` attribute is annotated at a struct.

So, if this limitation:
- the token it is attached needs to be called by dart code
- the token can only be a struct or function, not a mod
is true, I will update the documentation of it :)

fzyzcjy:
If so, I suspect dart_code may currently only be put inside class or something like that... Thus, we may enhance frb a little bit to let it generate dart_code for functions etc. (which is not too hard, since dart_code is nothing but copy-pasting)

patmuk:
> If so, I suspect dart_code may currently only be put inside class or something like that... Thus, we may enhance frb a little bit to let it generate dart_code for functions etc. (which is not too hard, since dart_code is nothing but copy-pasting)

Yes, that is actually the case :) I rewrote the documentation: https://github.com/fzyzcjy/flutter_rust_bridge/pull/2313

For me that limitation is ok at the moment. For the general case it is a bit weird that one needs to wrap the dart_code in a struct, and additionally that this struct needs to have another rust fn implemented - or be a parameter of another fn.

Would be good to change that :)

patmuk:
@fzyzcjy I finally made it :)

Please have a look - I suggest to first start with the documentation in `website/docs/guides/how-to/logging.md`.
Let us discuss if you find something not ergonomic or would prefer it in a different way. I can change or explain why I need to do it this way (and learn from you if that can be done differently :) ).

Here are some topics to address:
1. I did not implement any tests. `pure_dart` looked quite complex. Maybe you can do a start here? However, I am using the logging in `dart_minimal`, but with the default configuration.
2. I added `#[allow(dead_code)]` in `frb_codegen/src/library/codegen/generator/misc/structs_macro.rs`. I did this because I got may warnings of not using `mir` and `custom` when calling the code generation. We can revert this, if this warnings just appeared locally to me!
3. Similarly, I added 
```
lints.rust]
unexpected_cfgs = { level = "allow", check-cfg = ['cfg(wasm)'] }
```
to `frb_rust/Cargo.toml` to surpress these warnings.
4. I removed the opinionated logging from `frb_codegen/src/library/utils/logs.rs`. I did not replaced it either ... I was not sure if this is logging you use for your own project (in which case you could change to this new implementation) or if this is used within the rust code of frb.
5. I removed the previous logger implementation in `frb_rust/Cargo.toml`and `frb_rust/src/misc/user_utils.rs`
6. I put the macro in `frb_rust/src/misc/logging.rs` - any better fitting place is good as well :)
7. I made the macro's integration dependent to the log feature in `frb_rust/src/misc/mod.rs`

I did not yet waited for CI to finish - will pick that up later.

fzyzcjy:
Good job and congratulations! 

I will review probably within a day or two, since quite busy now.

patmuk:
@fzyzcjy please have a look at CI as well - I don't know why it is failing :(

frb_rust/Cargo.toml
```diff
@@ -78,5 +72,8 @@ chrono = ["dep:chrono", "allo-isolate/chrono"]
 wasm-start = ["console_error_panic_hook"]
 thread-pool = ["dep:threadpool"]
 rust-async = ["dep:tokio", "dep:futures", "dep:wasm-bindgen-futures"]
-user-utils = ["dep:android_logger", "dep:oslog"]
+user-utils = []
 dart-opaque = ["dep:dart-sys-fork"]
+
+[lints.rust]
+unexpected_cfgs = { level = "allow", check-cfg = ['cfg(wasm)'] }
```
fzyzcjy, line 79: looks like we may not allow this. instead, let's migrate all `cfg(wasm)` to standard things like `#[cfg(target_family = "wasm")]` - maybe we can do this in another PR to keep things clear

frb_rust/Cargo.toml
```diff
@@ -78,5 +72,8 @@ chrono = ["dep:chrono", "allo-isolate/chrono"]
 wasm-start = ["console_error_panic_hook"]
 thread-pool = ["dep:threadpool"]
 rust-async = ["dep:tokio", "dep:futures", "dep:wasm-bindgen-futures"]
-user-utils = ["dep:android_logger", "dep:oslog"]
+user-utils = []
```
fzyzcjy, line 75: design choice question: shall we by default log to Flutter, or log to android/ios system log? If the former, if our Flutter somehow fails, then we cannot log things visibly.

fzyzcjy (commented):
Go through a first quick review, mainly focusing on the high-level things (only picked 1 nit, others not carefully checked). Good job!

(I will review more in depth soon)

fzyzcjy:
> I did not implement any tests

No worries, we can do it last. Yes just do things in pure_dart. Probably add a few lines mimicking e.g. https://github.com/fzyzcjy/flutter_rust_bridge/blob/4f5cf68f56ed56df62264227d81891744c356403/frb_example/pure_dart/test/api/misc_no_twin_example_a_test.dart#L10

frb_rust/src/misc/logging.rs
```diff
@@ -0,0 +1,233 @@
+#[macro_export]
+macro_rules! enable_frb_logging {
+    () => {
+#[allow(clippy::crate_in_macro_def)]
+use crate::frb_generated::StreamSink;
+use flutter_rust_bridge::frb as frb_4_log;
```
fzyzcjy, line 6: nit: maybe just do not `use` anything, since otherwise the user will get confused after executing this macro. Instead, just write down the super-long name directly (a bit ugly but acceptable).

frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template
```diff
@@ -8,3 +8,4 @@ crate-type = ["cdylib", "staticlib"]
 
 [dependencies]
 flutter_rust_bridge = REPLACE_ME_RUST_FRB_DEPENDENCY
+log = "0.4.20"
```
fzyzcjy, line 11:
```suggestion
log = "0.4.20" # Optional - can be removed if you do not want logging
```

frb_rust/src/misc/logging.rs
```diff
@@ -0,0 +1,233 @@
...
+import 'dart:io';
+import 'package:logging/logging.dart';
+
+static void default_log_function(Log2DartLogRecord record) {
+  print(
+    '[${DateTime.now()} ${log_level_from_number(record.levelNumber)} @${record.rustLog ? 'Rust' : 'Dart'} ${record.loggerName}] ${record.message}');
+}
+
+/// initialize the logging system, including the rust logger
+static Logger init_logger(
```
fzyzcjy, line 19: nit: IIRC we should have camelCase in Dart

frb_example/dart_minimal/lib/main.dart
```diff
@@ -1,8 +1,11 @@
 import 'package:frb_example_dart_minimal/src/rust/api/minimal.dart';
 import 'package:frb_example_dart_minimal/src/rust/frb_generated.dart';
 
+final LOGGER = FRBLogger.init_logger();
```
fzyzcjy, line 4: Hmm, this does not seem to be super perfect API... I hope the `RustLib.init` is enough to be the only function called to initialize everything.

To allow initialization logic to be called, maybe we can add support to some attribute like `#[frb(init_dart="this_Dart_code_will_be_called_when_frb_init")]` and use that to init logging.

frb_example/dart_minimal/lib/main.dart
```diff
@@ -1,8 +1,11 @@
...
+
 // If you are developing a binary program, you may want to put it in `bin/something.dart`
 Future<void> main() async {
   await RustLib.init();
-  print('Call Rust and get: 100+200 = ${await minimalAdder(a: 100, b: 200)}');
+  LOGGER.info(
```
fzyzcjy, line 9: maybe we can make the LOGGER *only* used from Rust. If some Dart users want to do logging, just follow the https://pub.dev/packages/logging doc and create another new logger.

frb_rust/src/misc/logging.rs
fzyzcjy, line 19: todo(fzyzcjy): review this core file again after it is modified

fzyzcjy (commented):
A deeper review, but mainly are nits :)

patmuk:
Now I am thinking if it wouldn't have been better to implement logging from the other side around: Send Dart logs to rust.
That way we might not need macros, as from the Dart code it would just call the rust log function ... I could implement that in addition, have the user configure the direction he likes.

For me personally it would depend how to handle logging configuration best - I here I lack the experience. I mean, would it be better to have log outputs configured via rust (using `opinionated_logger`, thus one logger for codegen and prod, OSLogger and Android crates, ... maybe some fancy Network logging ...) or better in the Flutter ecosystem (maybe better integrated in the mobile OSs, into firebase crashlytics, etc.).

For now I am undecided and would first gather some experience. I understood that in your product you forward the rust logs to dart as well, right?
Nevertheless, I think dart->rust would be very easy to implement (in fact easier than what I did ... duh).

fzyzcjy:
> Send Dart logs to rust

Hmm, if the user is "using Rust for e.g. computation-heavy work for the Flutter app", then I guess he or she will have already setup the Dart logging system before adding frb or Rust code, and thus "send rust logs to dart" would be a good default choice.

> That way we might not need macros

I guess it still needs macros (for complex scenarios at least)

patmuk:
> > Send Dart logs to rust
> 
> (...) I guess he or she will have already setup the Dart logging system before (...)

True.
 
> My personal choice for my app is also send Rust log to Dart.

Thanks for sharing.
 
> > That way we might not need macros
> 
> I guess it still needs macros (for complex scenarios at least)

True - at least the rust calls to be called by the Dart logging function (if Dart -> rust) need to be wired ... so the macro is a must.


I am currently simplifying the usage a bit. If you like you can already have a look - but I will extend the macro with the same optional parameters as the dart function `init_logger` so that less code is needed :)

patmuk:
I finished the improvements - feel free to review this PR!

patmuk:
Oh, I overlooked that you did already a review! Working through your comments now!

frb_rust/Cargo.toml
patmuk: Done in #2329 and removed the lint.

frb_rust/Cargo.toml
patmuk: AFAIK (from trying myself) the stdout is logged to the oslog anyways. I did not try it with Android and in iOS with a simulator only.
But I think it is unlikely that Flutter fails (but the iOS/Android logger not) and more user/beginner friendly to log to stdout per default.

I suggest that we improve/reintroduce these mobile loggers if really needed. I will stumble over that in this case anyways.

website/docs/guides/how-to/logging.md
```diff
@@ -1,89 +1,147 @@
 # Logging
 
-Since I have seen some questions asking how logging can be implemented with a Flutter + Rust application, here are some examples.
+Flutter Rust Bridge comes with logging build in - but you can override it with your own logging framework of choice.
```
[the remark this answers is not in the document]
patmuk: thanks, implemented that :)

website/docs/guides/how-to/logging.md
```diff
@@ -1,89 +1,147 @@
...
+More concrete, you can overwrite the log outputting function with the one(s) of your logging framework(s) of choice, which will automatically log statements from Rust as well.
```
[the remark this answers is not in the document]
patmuk: Okay.  Note that I rewrote the docs after making the logging better to use. Feel free to change it anywhere you like :)

frb_rust/src/misc/user_utils.rs
```diff
@@ -3,9 +3,6 @@ use crate::misc::panic_backtrace::PanicBacktrace;
 /// Setup defaults that is usually useful for a new project.
 /// Surely, you are free to customize everything.
 pub fn setup_default_user_utils() {
-    // setup log before others, such that we can see logs in other setup functions
-    #[cfg(feature = "log")]
-    setup_log_to_console();
```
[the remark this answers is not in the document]
patmuk: I agree to avoid breaking changes ... but are you sure about this one?

Here is the code
```Rust
#[cfg(feature = "log")]
fn setup_log_to_console() {
    #[cfg(target_os = "android")]
    let _ = android_logger::init_once(
        android_logger::Config::default().with_max_level(log::LevelFilter::Trace),
    );

    #[cfg(any(target_os = "ios", target_os = "macos"))]
    let _ = oslog::OsLogger::new("frb_user")
        .level_filter(log::LevelFilter::Trace)
        .init();

    #[cfg(wasm)]
    let _ = crate::misc::web_utils::WebConsoleLogger::init();

    // TODO add more platforms
}
```
It was called only indirectly, through setup_default_user_utils() - and I think it was added only in v.2.3 or v.2.4? I roughly remember that it broke my code - so removing might actually un-break code :)

Nevertheless, it just sets the log value to TRACE (I set INFO as a default) and initializes the mobile loggers.

So, the user's code would still compile. He would still see the log output, just additionally in stdout (as I assume, I might be wrong here). I can change the default log level to TRACE - but I think INFO is a better default?

frb_rust/src/misc/logging.rs
```diff
@@ -0,0 +1,233 @@
+#[macro_export]
+macro_rules! enable_frb_logging {
```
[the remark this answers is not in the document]
patmuk: I somehow this did not work in my VSCode ... but I formatted the code by hand now :)

frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template
patmuk: I needed to introduce a dependency to `chrono` as well ...
I marked both as
```
log = "0.4.20"  # Optional - can be removed if you do not want logging
chrono = "0.4.23"  # Optional - can be removed if you do not want logging (needed only for the default log output)
```

frb_rust/src/misc/logging.rs
patmuk: done :)

frb_rust/src/misc/logging.rs
patmuk: yes, I did this now everywhere in the user visible functions.

I did not do that in internal (_bla_bla) functions within the macro ... I think it is better readble this way - but I have not strong opinion on that, can do it there as well.

frb_example/dart_minimal/lib/main.dart
patmuk: :) That bothered me as well :)

That was the `ergonomics improvement` I did today. Now all can be configured via parameters to the macro. However, optionally a user can still call `FRBLogger.init_logger();` for customization - but doesn't have to.

When a user provides a custom log function he probably wants to write it in dart - thus he can use this function instead of providing a custom log function to the macro in rust.

frb_example/dart_minimal/lib/main.dart
patmuk: I refactored it now so that a user can just do that. 
Though instead of `final LOGGER = Logger.('my logger');` he would call `final LOGGER = FRBLogger.init_logger();` at least once.
As a bonus he doesn't need to add the logging package as a dependency.

frb_example/dart_minimal/rust/src/api/minimal.rs
```diff
@@ -1,10 +1,15 @@
+use flutter_rust_bridge::enable_frb_logging;
 use flutter_rust_bridge::frb;
+use log;
+
+enable_frb_logging!();
```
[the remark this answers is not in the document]
patmuk: I agree - there is a lot of code generated into `minimal.dart` now!

I failed to generate it into `frb_generated.dart`, because the stream glue code would not generate.
I think it would be best to do that.

I am afraid that a user might overlook the need for that extra file, forget to add this to the configuration, etc.
Though we could generate the setup when `frb create` and document it otherwise ...

Nevertheless, I think adding it to `frb_generated.dart` would be best - if you know how?

We might need to do the customizing differently, if we move the macro away from the users eyes ... though there would still be the (now) optional `FRBLogger.init_logger()` function.

frb_example/dart_minimal/pubspec.yaml
```diff
@@ -11,6 +11,7 @@ dependencies:
   flutter_rust_bridge_utils:
     path: ../../frb_utils
   freezed_annotation: ^2.2.0
+  logging: any
```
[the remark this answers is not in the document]
patmuk: Oh! Interesting ... Dart is transitively exporting dependencies (which Rust forbids) - but they solve it by automatically setting the dependency like this in the `pubspec.yaml`.

I set it to 1.2.0, like it was before. (My comment above of the user not needing to add the logging dependency was wrong.)

frb_rust/src/for_generated/boilerplate.rs
```diff
@@ -214,9 +214,11 @@ macro_rules! frb_generated_default_handler {
                 assert_eq!(
                     FLUTTER_RUST_BRIDGE_CODEGEN_VERSION,
                     flutter_rust_bridge::for_generated::FLUTTER_RUST_BRIDGE_RUNTIME_VERSION,
+                    // flutter_rust_bridge::for_generated::FLUTTER_RUST_BRIDGE_RUNTIME_VERSION,
```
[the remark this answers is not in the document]
patmuk: Ups ... removed now!

frb_rust/src/misc/logging.rs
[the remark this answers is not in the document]
patmuk: I changed the API, so that the configuration is simpler.

These settings need to happen at the same time.
The problem is this code:
```Rust
    pub fn initialize_log2dart(log_stream: crate::frb_generated::StreamSink<Log2DartLogRecord>, max_log_level: u16) {
      log::set_boxed_logger(Box::new(FRBLogger {
        stream_sink: log_stream,
      }))
      .map(|()| log::set_max_level(from_u16(max_log_level)))
      .expect("initialize_log2dart is called only once!");
(...)
```
Because I move the `FRBLogger` into `log::` I cannot change the log level subsequentially anymore. 
In between I had a version where one could set this and the logging function later in the program execution - but while I was struggling with lifetimes I thought that a user might never need to change these later in a program.

If it is just to make the function smaller and easier to read: Sure, we can split it up. But eventually it needs to be called at once.

patmuk:
I finished working through your review. Many thanks!

Now only until tests are to-be-done.
I am contemplating to implement a `Dart-to-Rust` option as well, while I have this fresh in mind. I would do so by splitting up the macro (internally) and giving it a 4th parameter, on which dependence different code parts are inserted, so either rust logs are forwarded via a stream to dart (current implementation) or dart logs are processed by dart calling the rust log function.

Sounds easy, but is probably not needed, as discussed before ...

For now I will hold still until we resolved the above discussions.

The blow-up of the `minimal.rs` is a bad think, would be good if we find a way to put the logging code into `frb_generated` or similar.

frb_rust/src/misc/user_utils.rs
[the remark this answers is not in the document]
fzyzcjy: > and I think it was added only in v.2.3 or v.2.4

Oh you are right, this is a very special case. But if they are in that version, then we break (unbreak) them again! I am not sure whether this is the correct way...

(INFO lgtm)

frb_rust/src/misc/logging.rs
fzyzcjy: > I think it is better readble this way

I guess following Dart's convention will be a bit more readable

frb_example/dart_minimal/lib/main.dart
fzyzcjy: (I will check this in the next round review)

frb_example/dart_minimal/rust/src/api/minimal.rs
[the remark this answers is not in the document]
fzyzcjy: > I am afraid that a user might overlook the need for that extra file, forget to add this to the configuration, etc.

We should generate it in the default template.

Also maybe creating a doc mentioning how to migrate.

frb_example/dart_minimal/lib/main.dart
fzyzcjy: Again it would be great to avoid such extra initialization (discussed below)

frb_rust/src/misc/logging.rs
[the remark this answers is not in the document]
fzyzcjy: > But eventually it needs to be called at once.

By default yes. But users should be free to customize etc.

Suppose a simple example: User may like to have Rust log forward to Dart, but at the same time, he or she already setup the dart logging (e.g. his own Logger.root.level, his own Logger.root.onRecord, let's say he sends it to Sentry instead of printing for some reason). Then he should *not* call our 2nd function.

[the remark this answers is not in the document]
fzyzcjy: > I cannot change the log level subsequentially anymore.

What about

```
pub fn another_function_that_freely_changes_log_level(){
  log::set_max_level(TRACE)
}
```

seems we can change the level

fzyzcjy:
Great and you are welcome!

> Dart-to-Rust option

I guess we can add this option in doc only and mention "if you want it, create an issue and discuss your scenarios". I personally think this will not be very frequently needed, thus we can save time efforts not implementing it, but I can be totally wrong!

> The bloating of the minimal.rs is a bad think, would be good if we find a way to put the logging code into frb_generated or similar.

Totally agree (see my reply above)

frb_rust/src/misc/user_utils.rs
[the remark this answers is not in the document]
patmuk: As always - it is ultimately your decision :)

I am thinking that these mobile logging crates might conflict with some Dart/Flutter packages.
I think that it would be better to keep the default logging simple.
Sure, one can change/not use it, if one has conflicts ...

But I wonder how breaking the change would actually be?
Unless I overlook something, it should really not be much noticeable to the user.

Meaning, there should be no compile error.
If I am wrong about stdout being automatically logged in the OsLogger & Android logger output the user would not see any difference.

frb_rust/src/misc/logging.rs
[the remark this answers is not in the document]
patmuk: Ok, I will look into it. I forgot why I wasn't able to change it after the initialization ... maybe it was a temporary limitation during refactoring ...

But let's first refactor where the code will be generated to, do unit tests - and then I will enable setting the log level after initialization.

As for your example, a user could create another logger for `Sentry`, that inherits from the FRB created root logger and change anything he likes there.

But in general, I think if a user changes something using this logging implementation, it should always apply to both sides, rust and dart. That way it is least surprising :)

patmuk: Ultimately I have to agree :) It affected only one call, `_log_level_from_str`, which is now `_logLevelFromStr`.

I actually mixed it up with `initialize_log2dart`, which is a rust function, that I changed to initialize_log_2_dart`.

frb_example/dart_minimal/rust/src/api/minimal.rs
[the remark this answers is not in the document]
patmuk: > Also maybe creating a doc mentioning how to migrate.

Yes, but I don't know what setup a user might have had before?
In my project, I implemented with manual "workaround" suggested in some issue ticket, loging with OsLogger and AndroidLogger. Then v.2.3 (or v.2.4) implemented that internally in FRB and broke my compilation.

After removing my code nothing was logged and I wondered what one needs to do to use the internal implementation ... which ultimately lead me to do this PR, so it is clearer to the user.
Originally I thought it is only improving the docs :)

So, my migration advice would simply be "remove whatever setup you have and use the setup & customization described in the doc. On the log statement itself you don't need to do any change".

> > Nevertheless, I think adding it to frb_generated.dart would be best - if you know how?
> 
> Yes I agree, that looks great! This may need another PR adding a feature, say, `#[frb(location_frb_generated)]`.

Ah, that sounds great! Would you do this PR? Then we keep this PR on hold until that PR is ready?

patmuk:
> > I did not implement any tests
> 
> No worries, we can do it last. Yes just do things in pure_dart. Probably add a few lines mimicking e.g.
>  https://github.com/fzyzcjy/flutter_rust_bridge/blob/4f5cf68f56ed56df62264227d81891744c356403/frb_example/pure_dart/test/api/misc_no_twin_example_a_test.dart#L10
> 

I looked into it, but it is not so easy to understand ... it would help, if you can give me a kick-start here: If you could implement a test for the default settings, meaning implementing a test that calls `enable_frb_logging!()` on the rust side and executes a `log::info!('whatever rust')` from Rust and a `FRBLogger.getLogger(`whatever dart`)` from Dart, that would be great!

Based on that I can do all the other scenarios.

As for asserting the test result, if it is too complicated to assert what has been written to the console we can set a custom log function that writes into a variable instead.

frb_rust/src/misc/user_utils.rs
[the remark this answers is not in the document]
fzyzcjy: > stdout being automatically logged in the OsLogger & Android logger

Hmm IIRC long ago people saying that `println` will *not* print anything to android or ios outputs. (Maybe they are talking about another output?)

Except for the breaking change issue, looks like this removal is pretty reasonable.
What about this:

1. We remove this (just like what you have done)
2. In migration doc, we tell users what happens and what should be done (e.g. "please add the new `setup_our_good_logging_system!()` macro")

(TODO, finish this later)

frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template
```diff
@@ -8,3 +8,5 @@ crate-type = ["cdylib", "staticlib"]
 
 [dependencies]
 flutter_rust_bridge = REPLACE_ME_RUST_FRB_DEPENDENCY
+log = "0.4.20"  # Optional - can be removed if you do not want logging
+chrono = "0.4.23"  # Optional - can be removed if you do not want logging (needed only for the default log output)
```
fzyzcjy, line 12: nit: if you want to use chrono only in the macro, try `flutter_rust_bridge::for_generated::chrono` (and at the same time export it in frb_generated.rs)

fzyzcjy (commented):
(just happen to see this, thus this isolated nit)

fzyzcjy:
For a few comments above, here are my two cents (not reply above because they are correlated):

[TODO]

frb_example/dart_minimal/rust/src/api/minimal.rs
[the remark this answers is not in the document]
fzyzcjy: > Ah, that sounds great! Would you do this PR? Then we keep this PR on hold until that PR is ready?

I may not have enough time for that PR recently, so feel free to PR for that! 

Before that is implemented, we can "imagine" it exists, i.e. ignore the ugly generated minimal.dart, knowing it will be beautiful in the future.

frb_example/dart_minimal/rust/src/api/minimal.rs
```diff
@@ -1,10 +1,37 @@
...
+// enable_frb_logging!(maxLoglevel = "warning");
+// enable_frb_logging!(
+//     customLogFunction = (|record: Log2DartLogRecord| {
+//         let logger_name = record.logger_name;
+//         let message = record.message;
+//         println!("[{logger_name}] {message})");
+//     })
+// );
 
 #[frb(init)]
```
fzyzcjy, line 29: nit: thinking about what should be exposed to end user, i.e. what should be the code end user put/see in minimal.rs

### choice 1

```rust
flutter_rust_bridge::setup_default_logging!();

#[frb(init)]
pub fn init_app() {
    flutter_rust_bridge::setup_default_user_utils();
}
```

### choice 2

```rust
flutter_rust_bridge::setup_default_user_utils!();
```

where `setup_default_user_utils`'s definition is, "call setup_default_logging + that fn init_app".
and we still allow customizations. For example, suppose one wants to remove logging, but keep others, then he or she can remove setup_default_user_utils, and copy-paste the `fn initapp` or whatever he likes.

pros: less confusing to users (in choice 1, users may think "why logging is not user utils?")

pros: less verbose (in choice 1, user may think "oh so many lines")

cons: a bit less explicit. (but users can click that macro to see what is going on, so maybe acceptable)

frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template
fzyzcjy, line 1: btw, maybe we can add a few lines in codegen to sanity check, if we find the string `setup_default_user_utils`, maybe we just `log::warn!("Please migrate to new default user utils, see some_url_link_here for details")` in codegen.

frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template
patmuk: Ah, that would be great - bothered me to having to add this second dependency.

But I didn't understand how? 

> flutter_rust_bridge::for_generated::chrono

I can't find it - your didn't mean
```Rust
    fn _default_log_fn (record: Log2DartLogRecord) {
      let timestamp = flutter_rust_bridge::for_generated::chrono::Local::now();
```?
I only use it at [line 27 in the macro](https://github.com/patmuk/flutter_rust_bridge/blob/b561524bcae9466a68c72dc845c2c657fcd78ffc/frb_rust/src/misc/logging.rs#L27)

> at the same time export it in frb_generated.rs

Do you mean changing `frb_codegen` so it contains `pub use flutter_rust_bridge::for_generated::chrono`?

File: frb_example/dart_minimal/rust/src/api/minimal.rs (hunk 16 above)
Comment (patmuk) [reply frb_example/dart_minimal/rust/src/api/minimal.rs:29]: I agree with choice 2 as well :) probably there will be more supporting tools in future - and then it would get very (too) verbose. So hiding that all behind something like `setup_default_(...)` sounds very good to me :)

I am quite confident, the I can rewrite the code so that the customizing can be done any time ... from setting the log level to a custom log function.

I think this would work from both sides - meaning one could call a Rust function `set_custom_log_fn` from `minimal.rs` or `main.dart`. So my next question is not very relevant, but: We are assuming a user would do the logging configuration preferably from Dart, right? 

Thus, for setting the log level one needs to use the Dart logging package's definition (which I find quite unusual ...), even when setting it from Rust. I could easily change this though, I was aiming for consistency.

As for the custom log function, if set from Dart it shall be Dart code, if set from Rust Rust code. I think we can change this as well, but that requires a bit more effort. But probably this is a good default.

So, to summarize, I think it would be best if:
1. logging gets automatically initialized with defaults, triggered as part of RustLib.init()
2. thus, there is no visible code in neither `main.dart`, nor `minimal.rs`
3. issuing log statements works as usual in the respective language (`Logger().info('from Dart)`, log::info(`from Rust`) - this should already work :)
4. Customizing the log level can be done from both, Dart or Rust code
5. the generated code ends up in frb_generated.rs/dart and not minimal.rs

For (5.), I wasn't able to get it there - the StreamSink was never processed. I tried calling the macro in `frb_rust/src/for_generated/boilerplate.rs`, `frb_rust/src/misc/user_utils.rs`, etc. Probably you see right away what needs to be fixed :)

File: frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template
Comment (patmuk) [reply frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template:1]: I think that would be great!

File: frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template (hunk 15 above)
Comment (fzyzcjy) [reply frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template:12]: 1. add a line mimicking https://github.com/fzyzcjy/flutter_rust_bridge/blob/36e702332e1fb9855a28472c2e060a46d8063daf/frb_rust/src/for_generated/mod.rs#L70
2. write `flutter_rust_bridge::for_generated::chrono::Something`

Comment (patmuk) [reply frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template]: Ah, great, that worked :)
I needed to add `chrono` to the default dependencies in `frb_rust`.

File: frb_example/dart_minimal/rust/src/api/minimal.rs (hunk 16 above)
Comment (fzyzcjy) [reply frb_example/dart_minimal/rust/src/api/minimal.rs:29]: > probably there will be more supporting tools in future

Totally agree. Choice 1 will require user to migrate again at that time which is not optimal.

Agree 1-6! Except minor details:

> A user would disable this by not having the log feature active

I guess by simply copy-paste content in setup_default_user_utilities and remove the one about logging.
Using feature for this, personally speaking, may be a little bit overkill, and a bit less flexible.

> Customizing the log level can be done from both, Dart or Rust code

Yes, for simplicity (of our implementation), users can use Rust and Dart's native mechanism to set logging level respectively. 

> the generated code ends up in frb_generated.rs/dart and not minimal.rs

I will handle that

File: frb_rust/Cargo.toml (hunk 2 above)
Comment (fzyzcjy) [reply frb_rust/Cargo.toml:77]: Hmm not sure why I missed this reply. Anyway we have discussed it below, so no worries!

File: frb_example/dart_minimal/rust/src/api/minimal.rs (hunk 16 above)
Comment (patmuk) [reply frb_example/dart_minimal/rust/src/api/minimal.rs:29]: > > A user would disable this by not having the log feature active
> 
> I guess by simply copy-paste content in setup_default_user_utilities and remove the one about logging. Using feature for this, personally speaking, may be a little bit overkill, and a bit less flexible.

Yes, true, that would be a way as well. I was referring to the already existing feature flag 'log' - I used `#[cfg(feature = "log")]` wherever it made sense :)


> > Customizing the log level can be done from both, Dart or Rust code
> 
> Yes, for simplicity (of our implementation), users can use Rust and Dart's native mechanism to set logging level respectively.

Oh yes, we can just use the usual functions for setting the log level - will do that.
 
> > the generated code ends up in frb_generated.rs/dart and not minimal.rs
> 
> I will handle that

Ah, great , thanks :)

Comment (fzyzcjy) [reply frb_example/dart_minimal/rust/src/api/minimal.rs:29]: That flag was there because I wanted frb_rust to have minimal dependency if a user wants it. It accidentally also controls that existing behavior...

Comment (patmuk) [reply frb_example/dart_minimal/rust/src/api/minimal.rs:29]: Good idea :) Ah, ok. If this flag is needed always for the existing behavior (mandatory parts?) I would suggest to remove it from these and repurpose it for this logging implementation. Probably users have used it for that. I actually assume that probably nobody deactivated it ... as this is not so straight-forward in Rust.

**Comment by fzyzcjy** [comment]
👀

**Comment by fzyzcjy** [comment]
Sure!

**Comment by patmuk** [comment]
Still don't have time ... could you please again reopen this PR?

**Comment by fzyzcjy** [comment]
Sure and looking forward to it!

**Comment by patmuk** [comment]
@fzyzcjy I found some time to update this PR and finalize it :) Currently I am merging the latest master - but I ran into a problem testing the code :(

I am running 
`dart --enable-experiment=native-assets run lib/main.dart` in `frb_example/dart_minimal`, but I am getting errors and wonder if you get them as well?

Since I am getting the same errors in the latest `master` and in my branch before the merge and upgrading the tooling, I am quite sure it is because of the tooling ... but I don't know why.
I am using
```
› dart --version                                          
Dart SDK version: 3.7.2 (stable) (Tue Mar 11 04:27:50 2025 -0700) on "macos_arm64"
```

The error I am getting is:
```
› dart --enable-experiment=native-assets run lib/main.dart
"/Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_example/dart_minimal/hook/build.dart" file not found.
Building native assets for package:frb_example_dart_minimal failed.
Compilation of hook returned with exit code: 255.
To reproduce run:
/Users/patmuk/code/own/oss/frb/.toolchain/flutter-local/flutter/bin/cache/dart-sdk/bin/dart compile kernel --packages=/Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_example/dart_minimal/.dart_tool/package_config.json --output=/Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_example/dart_minimal/.dart_tool/native_assets_builder/frb_example_dart_minimal/36bad3131645e96b3e5f196e7b126c8d/hook.dill --depfile=/Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_example/dart_minimal/.dart_tool/native_assets_builder/frb_example_dart_minimal/36bad3131645e96b3e5f196e7b126c8d/hook.dill.d /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_example/dart_minimal/hook/build.dart
stderr:
"/Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_example/dart_minimal/hook/build.dart" file not found.

stdout:

        
Error: Compiling native assets failed.
```

https://github.com/flutter/flutter/issues/138727 mentions this problem with https://github.com/dart-lang/sdk/issues/54111 solving it by having the `build.dart` in `hook/build.dart`. 
But I don't know if this build file is intended to get `cargo build` running?

But when I remove the file I get
```
› dart --enable-experiment=native-assets run lib/main.dart
Unhandled exception:
Bad state: Content hash on Dart side (742269791) is different from Rust side (-2119384465), indicating out-of-sync code. This may happen when, for example, the Dart code is hot-restarted/hot-reloaded without recompiling Rust code. (Note: This is just a sanity check. Even if content hash does not change, the code may still change and needs to be recompiled)
#0      BaseEntrypoint._sanityCheckContentHash (package:flutter_rust_bridge/src/main_components/entrypoint.dart:108:7)
#1      BaseEntrypoint.initImpl (package:flutter_rust_bridge/src/main_components/entrypoint.dart:52:5)
<asynchronous suspension>
#2      RustLib.init (package:frb_example_dart_minimal/src/rust/frb_generated.dart:27:5)
<asynchronous suspension>
#3      main (package:frb_example_dart_minimal/main.dart:8:3)
<asynchronous suspension>
```
when I move it to `hook/build.dart` 
```
 dart --enable-experiment=native-assets run lib/main.dart
> cargo build --release (pwd: /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_example/dart_minimal/rust, env: {})
    Finished `release` profile [optimized] target(s) in 0.13s
> cargo build --release (pwd: /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_example/dart_minimal/rust, env: {})
    Finished `release` profile [optimized] target(s) in 0.02s
> cargo build --release (pwd: /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_example/dart_minimal/rust, env: {})
    Finished `release` profile [optimized] target(s) in 0.02s
> cargo build --release (pwd: /Users/patmuk/code/own/oss/frb/flutter_rust_bridge_fork_pat/frb_example/dart_minimal/rust, env: {})
    Finished `release` profile [optimized] target(s) in 0.02s
```
`cargo build` runs in an endless loop.

However, the dart issue tickets I mentioned are over a year old - I wonder how this is impacting now? Something seems very off with my setup, but I can't figure out what. Gemini (surprisingly good with supporting flutter coding) suggests to modify the `build.dart` file quite heavily - but that can't be the solution, if it runs fine for others.

**Comment by patmuk** [comment]
It seems not to compile in CI either:
https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/15052130037/job/42309299553
```
[2025-05-15T18:50:23.075Z INFO frb_codegen/src/library/commands/fvm.rs:15] Has .fvmrc but no fvm binary installation, thus skip using fvm.
[2025-05-15T18:50:23.198Z WARN frb_codegen/src/library/commands/command_runner.rs:161] command=cd "/Users/runner/work/flutter_rust_bridge/flutter_rust_bridge/frb_example/dart_minimal/rust" && RUSTFLAGS="-C instrument-coverage --cfg=coverage --cfg=trybuild_no_target --cfg frb_expand" "cargo" "expand" "--lib" "--theme=none" "--ugly" stdout= stderr=error: no such command: `expand`

	View all installed commands with `cargo --list`
	Find a package to install `expand` with `cargo search cargo-expand`

[2025-05-15T18:50:23.198Z INFO frb_codegen/src/library/commands/cargo_expand/real.rs:111] Cargo expand is not installed. Automatically install and re-run.
[2025-05-15T18:55:15.769Z INFO frb_codegen/src/library/commands/fvm.rs:15] Has .fvmrc but no fvm binary installation, thus skip using fvm.
[2025-05-15T18:55:23.648Z INFO frb_codegen/src/library/commands/fvm.rs:15] Has .fvmrc but no fvm binary installation, thus skip using fvm.
[2025-05-15T18:55:24.652Z WARN frb_codegen/src/library/codegen/polisher/auto_upgrade.rs:27] Auto upgrader find wrong Dart/Rust flutter_rust_bridge dependency version, please enable `auto_upgrade_dependencies` flag or upgrade manually.
Done!

    Finished report saved to /Users/runner/work/flutter_rust_bridge/flutter_rust_bridge/target/coverage/GenerateRunFrbCodegenCommandGenerate/codecov.json
transformCodecovReport act on /Users/runner/work/flutter_rust_bridge/flutter_rust_bridge/target/coverage/GenerateRunFrbCodegenCommandGenerate/codecov.json
> /bin/sh -c git diff --exit-code  (pwd: /Users/runner/work/flutter_rust_bridge/flutter_rust_bridge/, env: {CARGO_TERM_COLOR: always})
```

**Comment by patmuk** [comment]
Oh, I see that his is just running the code generation. I don't see any test for running the project. 
https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/15052130037/job/42309302248 runs dart test successfully.

**Comment by fzyzcjy** [comment]
About hook file, could you please try to make Dart version same as the one on CI - iirc the newer dart changes hook mechanism and I have not had time to upgrade the examples to support that (note this only affects examples and not real users)

**Comment by patmuk** [comment]
downgrading dart worked :) It builds again locally. Will continue later.

**Comment by patmuk** [comment]
I could run dart-minimal locally, and it looks like all tests in pure-dart, except the ones a socket, are passing locally.

But in CI I am getting errors in form of
```
> /bin/sh -c git diff --exit-code  (pwd: /Users/runner/work/flutter_rust_bridge/flutter_rust_bridge/, env: {CARGO_TERM_COLOR: always})
diff --git a/frb_example/pure_dart/rust/Cargo.lock b/frb_example/pure_dart/rust/Cargo.lock
index eef23f3..90669fe 100644
--- a/frb_example/pure_dart/rust/Cargo.lock
+++ b/frb_example/pure_dart/rust/Cargo.lock
@@ -17,15 +17,6 @@ version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 ```
(...)
```
Unhandled exception:
ProcessException: Bad exit code (1). If you want to see extra information, set FRB_DART_RUN_COMMAND_STDERR=1
  Command: /bin/sh -c git diff --exit-code 
#0      runCommand (package:flutter_rust_bridge/src/cli/run_command.dart:67:5)
<asynchronous suspension>
#1      SimpleExecutor.call (package:flutter_rust_bridge_internal/src/utils/makefile_dart_infra.dart:32:12)
<asynchronous suspension>
#2      _maybeSetExitIfChanged (package:flutter_rust_bridge_internal/src/makefile_dart/generate.dart:427:5)
<asynchronous suspension>
#3      wrapMaybeSetExitIfChangedRaw (package:flutter_rust_bridge_internal/src/makefile_dart/generate.dart:419:3)
<asynchronous suspension>
#4      testDartNative.<anonymous closure> (package:flutter_rust_bridge_internal/src/makefile_dart/test.dart:360:7)
<asynchronous suspension>
#5      withLlvmCovReport (package:flutter_rust_bridge_internal/src/makefile_dart/test.dart:405:15)
<asynchronous suspension>
#6      testDartNative (package:flutter_rust_bridge_internal/src/makefile_dart/test.dart:339:3)
<asynchronous suspension>
#7      SimpleConfigCommand.run (package:flutter_rust_bridge_internal/src/utils/makefile_dart_infra.dart:88:31)
<asynchronous suspension>
#8      CommandRunner.runCommand (package:args/command_runner.dart:212:13)
<asynchronous suspension>
#9      main (file:///Users/runner/work/flutter_rust_bridge/flutter_rust_bridge/tools/frb_internal/bin/flutter_rust_bridge_internal.dart:30:3)
<asynchronous suspension>
Error: Process completed with exit code 255.
```

What does this mean? How can I move further?

patmuk:
All tests of `pure-dart` are passing locally :)
(with
```
› dart --enable-experiment=native-assets --enable-vm-service  test 
```
)

Do you know what the CI errors mean?

patmuk:
@fzyzcjy Unfortunately I am stuck - I went through all changes and can't see what is breaking the CI.
Can you please have a look?

fzyzcjy:
Ah I thought I replied to you!

fzyzcjy:
> But in CI I am getting errors in form of

For diffs, just copy those diffs that causes CI failure and apply locally to code and CI will be green

patmuk:
No problem :)
I re-applied my changes to a newly forked branch (logging_overhaul_redo), but I get there CI errors as well. 

frb_internal —slow is not working locally - I am getting a file not found error - but on the latest master as well.
Otherwise, frb_internal generate and test and run work in pure_dart and dart_minimal. 

I don’t know how to progress.

fzyzcjy:
Then just don't run that --slow command, use 

```
./frb_internal generate-internal-frb-example-pure-dart ; ./frb_internal precommit-generate
```

which is a subset.

patmuk:
> ```
> ./frb_internal generate-internal-frb-example-pure-dart ; ./frb_internal precommit-generate
> ```
I just did this: They work fine in both branches (`log_overhaul` and `log_overhaul_redo`).

Don't know why CI doesn't :(

fzyzcjy:
Hmm, is your local version all match CI?

fzyzcjy:
Oh I checked your code. You need to change things in https://github.com/fzyzcjy/flutter_rust_bridge/tree/master/frb_codegen/assets/integration_template

patmuk:
> Hmm, is your local version all match CI?

You mean if my local files are the same as on CI? Yes, I even did `sudo rm -rf *` and `git reset`

patmuk:
> Oh I checked your code. You need to change things in https://github.com/fzyzcjy/flutter_rust_bridge/tree/master/frb_codegen/assets/integration_template

Ok … this should be the same as in master, except for a dependency on log. However, I am currently not on my computer, will check this in the evening. 
Thanks for the hint!

patmuk:
Ah, I misunderstood: You meant there needs to be changes, similar to dart_minimal, done?

I will try to figure out which changes exactly. What puzzles me is that before merging master CI worked - but I assume the testing setup has been improved since :)

How can I trigger the CI build mistakes locally so I get better feedback?

Many thanks

fzyzcjy:
> How can I trigger the CI build mistakes locally so I get better feedback?

Just check CI yaml files and it tells you which command it runs

patmuk:
> > How can I trigger the CI build mistakes locally so I get better feedback?
> 
> Just check CI yaml files or logs and it tells you which command it runs

That is what I don't understand: In CI (e.g. https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/15113789286/job/42479269730?pr=2308) the command
`/bin/sh -c git diff --exit-code`
is failing. I don't understand what the actual issue is.

patmuk:
> Oh I checked your code. You need to change things in https://github.com/fzyzcjy/flutter_rust_bridge/tree/master/frb_codegen/assets/integration_template

I did changes here, for example `frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template`

I don't see what is missing, or why `git diff` is stopping CI ... can you please have a deeper look?

fzyzcjy:
It says that, after the commands shown in the log lines above this line, now the code has one line of diff like the screenshot

image

patmuk:
> It says that, after the commands shown in the log lines above this line, now the code has one line of diff like the screenshot
> 

Ah, now I understand! It should have no changes! I misunderstood that these are the PRs changes that shall be there :)
Ok, will fix that one-by-one now, maybe in the `logging_overhaul_redo` PR, as that one is cleaner. Will update here after I have done that.

patmuk:
@fzyzcjy I fixed all CI complaints :)
I did this in another Pr, in which my changed are cleaner (re-did them):
https://github.com/fzyzcjy/flutter_rust_bridge/pull/2735

Ready for review and merge (in the other PR - let's delete this PR).
Best way: Read the docs first. I know you did already (and I didn't changed it), but it has been a while ...

Issue #2384, opened 2024-11-02
Web target with hash included in wasm pkg files to solve cache issues


Slixe:
**Is your feature request related to a problem? Please describe.**
I have an web application using WebAssembly behind cloudflare, even if the cache is fully disabled, some browsers still keep the old WASM & js files when we do an update.

**Describe the solution you'd like**
Having a hash concatened to the filename for easier update would be better.

**Describe alternatives you've considered**
Doing by hand using a bash script. But I don't know where it's exactly loaded by flutter.

fzyzcjy:
Looks reasonable! Looks like it is loaded in https://github.com/fzyzcjy/flutter_rust_bridge/blob/acf0d3248b2cd1edc2514d17a03e982f0fa6df78/frb_dart/lib/src/wasm_module/_web.dart#L12, loaded from https://github.com/fzyzcjy/flutter_rust_bridge/blob/master/frb_dart/lib/src/loader/_web.dart. It may not be too hard, and we even has some (pseudo) content hash computed in https://github.com/fzyzcjy/flutter_rust_bridge/blob/acf0d3248b2cd1edc2514d17a03e982f0fa6df78/frb_example/flutter_via_create/lib/src/rust/frb_generated.dart#L71.

svr2kos2:
I encountered this cache issue too and after following the discussion in PR https://github.com/fzyzcjy/flutter_rust_bridge/pull/2387, I found the proposed solution is great but it might not be implemented immediately.
As a workaround, I’ve created my own temporary solution, I hope it can help someone.
https://github.com/fzyzcjy/flutter_rust_bridge/compare/master...svr2kos2:flutter_rust_bridge:master
In your `pubspec.yaml`, use the following configuration:
```
flutter_rust_bridge: 
    git:
      url: https://github.com/svr2kos2/flutter_rust_bridge
      path: frb_dart
      ref: d8d23f7
```
For CI, since I’m using GitLab, my build command looks like this:
 `flutter build web --release --build-number $CI_PIPELINE_IID`
This approach allows me to avoid the cache problem for now while waiting for the official fix.

fzyzcjy:
@svr2kos2 Btw feel free to PR for this (indeed just click "create pull request" from that diff looks enough)! That sounds reasonable, because when someone make a new release of an app, it is reasonable (indeed mandatory?) to bump the version number.

PR #2387, opened 2024-11-02
implement hash concat for wasm pkg

Slixe:
## Changes

Closes https://github.com/fzyzcjy/flutter_rust_bridge/issues/2384

## Checklist

- [X] An issue to be fixed by this PR is listed above.

## Remark for PR creator

### DIFF
--- a/frb_dart/lib/src/cli/build_web/executor.dart
+++ b/frb_dart/lib/src/cli/build_web/executor.dart
@@ -67,7 +67,11 @@ Future<void> executeBuildWeb(BuildWebArgs args) async {
   final rustCrateName =
       await _getRustCreateName(rustCrateDir: args.rustCrateDir);
 
-  await _executeWasmPack(args, rustCrateName: rustCrateName);
+  final contentHash =
+      await _getRustContentHash(rustCrateDir: args.rustCrateDir);
+
+  await _executeWasmPack(args,
+      rustCrateName: rustCrateName, contentHash: contentHash);
 
   if (args.enableWasmBindgen) {
     await _executeWasmBindgen(args, rustCrateName: rustCrateName);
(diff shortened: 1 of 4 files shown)


frb_dart/lib/src/cli/build_web/executor.dart
```diff
@@ -149,8 +153,12 @@ Future<String> _getRustCreateName({required String rustCrateDir}) async {
   return rustCrateName;
 }
 
+Future<int> _getRustContentHash({required String rustCrateDir}) async {
+  throw UnimplementedError();
```
Slixe, line 157: @fzyzcjy how would you fetch the rust content hash from the build-web command?

fzyzcjy: One way may be parsing the generated code and find it.

But I am thinking about something more general: How does normal wasm-pack + js users (i.e. *not* wasm + dart users) handle this? I guess there may be some ways to add content hashing within wasm-pack ecosystem, and maybe we can utilize it.

Slixe: Maybe we can just build it like it was, and then at the final packing we hash the wasm file, append it to the filename and put it in the loader config too?

fzyzcjy: Looks reasonable! More brainstorms based on that:

* Instead of `rust_lib-abcdef.js`, we do `rust_lib.js?dummy_query=abcdef` (where `abcdef` is the content hash). Then we do not need to rename anything to keep it simple.
* Then, after `build-web`, if this feature is enabled, we just compute `hash(read_file(rust_lib.js) + read_file(rust_lib.wasm))`, and write it to a field in `frb_generated.dart`.

I am still wondering whether there are more "native" solutions instead of hand crafting it...

Slixe: You’re right, we could do that. Unfortunately the generated js file split the filename.js and add _bg.wasm as suffix. We would need to do a find and replace to add _bg.wasm?dummy=abcdef

so maybe having it in the directly would be better / easier.

Slixe: So, show we ask wasm-pack if they want to handle the hash on their side, or should I continue to work on a implementation here?

fzyzcjy: I guess we can create an issue there to have a discussion with them (feel free to cc me if you like). Because if this feature is done there, then not only frb but also every other package that depend on wasm-pack will benefit from this!

PR #2488, opened 2025-01-08
Add `json_serializable` package check for `freezed` to/fromJSON

satvikpendem:
## Changes

Fixes #2468

## Checklist

- [x] An issue to be fixed by this PR is listed above.
- [x] New tests are added to ensure new features are working. Please refer to [this page](https://cjycode.com/flutter_rust_bridge/guides/contributing/tip) to see how to add a test.
- [x] `./frb_internal precommit --mode slow` (or `fast`) is run (it internal runs code generator, does auto formatting, etc).
- [x] If this PR adds/changes features, documentations (in the `./website` folder) are updated.
- [ ] CI is passing. Please refer to [this page](https://cjycode.com/flutter_rust_bridge/guides/contributing/tip) to see how to solve a failed CI.

## Remark for PR creator

- `./frb_internal --help` shows utilities for development.
- If fzyzcjy does not reply for a few days, maybe he just did not see it, so please ping him.

PR #2495, opened 2025-01-13
Update CI versions

satvikpendem:
## Changes

Continuation of CI issues from #2488

## Checklist

## Remark for PR creator


satvikpendem:
@fzyzcjy looks like same bad exit code issues here too, with just the CI update

fzyzcjy:
Hmm, maybe we can look at logs above "bad exit code" line. For example, looks like some are just git diff that can be quickly fixed, e.g.

image

For this, looks like we need to upgrade android version according to the hint:

image

etc

satvikpendem:
Thanks, is there a way to force the git diff to pass? Looks like it's exiting if the new version for dependencies is not matching the old one so how do we tell it to prefer the newer version?

fzyzcjy:
> is there a way to force the git diff to pass?

Just to apply the git diff. Either by running command locally, or copy-paste the git diff section in the CI output to a local file and do `git apply` the patch.

fzyzcjy:
> I think the Java version is still 1.8, should we upgrade?

If upgrade solves the problem then I guess it is OK.

satvikpendem:
Would you mind taking a look at why the rest are failing? I fixed what I can but I don't understand why they'd suddenly start failing when it seems they passed just fine before.

satvikpendem:
Ok I fixed the Rust lint issue. For the snapshot file not being found, I think I read the solution is to `flutter pub get` and `flutter create .` but it doesn't seem like that solved it. For the packages, I think actually they need to be updated as Dart is at 3.6 and some of the packages have been updated 2 years ago. I just don't know how much will break if we do so.

satvikpendem:
Sorry for the delay, I don't think I'll continue to work on this updating the CI PR as it seems like it's breaking for many people not just for my particular feature and I don't have the time to fix the issues here.

fzyzcjy:
It's OK! I will handle it later but may not be very soon, since this is not a real bug for end users.

PR #2540, opened 2025-02-09
Fix web cache issues by appending version query parameter to wasm/js requests

svr2kos2:
## Changes
Temporarily fix (so don't close) #2384 

## Checklist

- [x] An issue to be fixed by this PR is listed above.
- [ ] New tests are added to ensure new features are working. Please refer to [this page](https://cjycode.com/flutter_rust_bridge/guides/contributing/tip) to see how to add a test.
- [ ] `./frb_internal precommit --mode slow` (or `fast`) is run (it internal runs code generator, does auto formatting, etc).
- [ ] If this PR adds/changes features, documentations (in the `./website` folder) are updated.
- [ ] CI is passing. Please refer to [this page](https://cjycode.com/flutter_rust_bridge/guides/contributing/tip) to see how to solve a failed CI.

## Remark for PR creator

- `./frb_internal --help` shows utilities for development.
- If fzyzcjy does not reply for a few days, maybe he just did not see it, so please ping him.

PR #2545, opened 2025-02-10
Remove package_info_plus, query version.json

svr2kos2:
## Changes
pre: #2540

## Checklist

## Remark for PR creator

### DIFF
--- a/frb_dart/lib/src/wasm_module/_web.dart
+++ b/frb_dart/lib/src/wasm_module/_web.dart
@@ -11,16 +11,17 @@ import 'dart:html' as html;
 /// {@macro flutter_rust_bridge.internal}
 Future<void> initializeWasmModule({required String root}) async {
   _ensureCrossOriginIsolated();
-  final versionQuery = await _getVersionQuery();
+  final packageInfo = await _getPackageInfo();
+  final versionQuery = _assembleVersionQuery(packageInfo);
   
-  final script = web.HTMLScriptElement()..src = '$root.js?$versionQuery';
+  final script = web.HTMLScriptElement()..src = '$root.js$versionQuery';
   web.document.head!.append(script);
 
   await script.onLoad.first;
 
   jsEval('window.wasm_bindgen = wasm_bindgen');
 
-  await _jsWasmBindgen('${root}_bg.wasm?$versionQuery').toDart;
+  await _jsWasmBindgen('${root}_bg.wasm$versionQuery').toDart;
 }
 
 @JS('wasm_bindgen')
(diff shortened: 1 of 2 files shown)


svr2kos2:
> Looks great! Some nits:
> 
> * Maybe we can make function return the version, and _another_ function that assembles query string from the version
> * Maybe we can add a comment pointing to package_info_plus's documentation and explicitly mention we follow their approach, and also mention why we cannot depend on that package, etc.

I've done that, but I encountered another issue.  
After reviewing [this](https://github.com/fluttercommunity/plus_plugins/blob/50d12990f88b66c8a1c76426295cd28ad11e3194/packages/package_info_plus/package_info_plus/lib/src/package_info_plus_web.dart#L28), I realized that the process involving baseUri is more complex than I initially thought.  
However, if I implement this function, it would essentially be a copy of the original implementation.  
Since the Plus plugins follow the BSD-3-Clause license, I'm not sure if it's okay to do this.  
But since it's not a complicated function, I think maybe it's ok to reference it?

fzyzcjy:
> if I implement this function, it would essentially be a copy of the original implementation.

Hmmm... Originally the implementation was only a few lines, and thus I thought adding comments and references might be ok, but now it seems to be longer...

fzyzcjy:
Oh I realize another issue: For a pure dart package, does it still contain package.json or not?

svr2kos2:
@fzyzcjy 
I just tested it, and unfortunately, it seems that a pure Dart project doesn't have a package.json or similar file. I also came across this https://github.com/dart-lang/sdk/issues/56017#issue-2354626866, where it looks like some people are exploring this further.

Given this, I'm thinking maybe we should reconsider the approach in https://github.com/fzyzcjy/flutter_rust_bridge/pull/2540#issuecomment-2646770368 where you mentioned earlier. Perhaps we should just extract the version and build_number from the `pubspec.yaml` instead.

However, one thing to note is that users can override the *build_number* using the `--build_number` flag. So, we should probably provide this parameter to users as well, allowing them to override it if needed.

svr2kos2:
The issue is that Flutter faces the same problem, but it allows users to define their own request URL for `main.dart.js`. However, frb currently doesn't support this.

Perhaps we can simplify this by letting users define their own URL through RustLib.init()? This would provide flexibility without overcomplicating the implementation.

fzyzcjy:
I see...

> Perhaps we can simplify this by letting users define their own URL

Yes that sounds like another alternative solution, somehow similar to https://github.com/fzyzcjy/flutter_rust_bridge/pull/2540#issuecomment-2646776125

fzyzcjy:
Indeed, directly generatinghashes during frb's build-web is also not hard to implement. I am mainly worried about, that approach may cause generated code to change on *each* compilation, and thus we may have to introduce a file like frb_dynamic.dart and suggest users to gitignore it.

svr2kos2:
So which solution do you prefer?
If the solution is simple, you can implement it yourself for efficiency.
Or I'd be happy to try implementing it if you like. 😄

fzyzcjy:
> So which solution do you prefer?

I guess the "directly generate hashes during frb build-web" may be a little bit better

> Or I'd be happy to try implementing it if you like. 😄

Feel free to have a try and looking forward to it!

PR #2596, opened 2025-03-07
Run CI to check allo-isolate change

alexlapa:
No changes, just want to run CI to check whether shekohex/allo-isolate/pull/65 will break anything


fzyzcjy:
No worries about `Command Integrate`, which is about the default templates. Just check whether tests pass

fzyzcjy:
Looks like CI is happy :)

image

PR #2639, opened 2025-03-13
Fix for strict-casts lint

PaulGrandperrin:
I started using the `strict-casts` lint (https://dart.dev/tools/analysis#enabling-additional-type-checks) and then my `frb_generated.dart` contained errors.

It's not a big issue as this file can be excluded from the analyzer and this lint is optional anyway, but I thought it would be nice that everything would work out-of-the box for people enabling this lint.

Feel free to disagree, it's not important.


fzyzcjy:
Looks good!

Just a nit: https://github.com/fzyzcjy/flutter_rust_bridge/blob/b92f6832e2713fa9b178544bb8a7c8073394b836/frb_dart/lib/src/manual_impl/_web.dart#L10

It seems it is not a `int` on Web. 

However, maybe we can change `dynamic` to `Object`. Will that make the lint error disappear?

fzyzcjy:
@all-contributors please add @PaulGrandperrin for code

PaulGrandperrin:
@fzyzcjy Sorry I didn't get back to you before.. I work on both the backend and the frontend of my project, and I try to stay focused on just one at a time (or else my head explodes). I didn't switch back to the frontend with flutter since then.

Anyway, thx for the merge

fzyzcjy:
@PaulGrandperrin It's OK and you are welcome!

PR #2653, opened 2025-03-21
Fix Dart_PostCObject double free

alexlapa:
## Changes

Fixes https://github.com/shekohex/allo-isolate/issues/64
For more information: https://github.com/dart-lang/native/issues/2079

## Procedure and Checklist

Utility commands
* Run tests: `./frb_internal test-dart-native --package frb_example/dart_minimal`


alexlapa:
@fzyzcjy,

This is not ready for merge, I just want to know what you think about this approach? If its okay then I will do wome additional testing and add some docs. Haven't ran internal script so diff would stay clean.

frb_dart/lib/src/generalized_frb_rust_binding/_io.dart
```diff
@@ -26,8 +37,14 @@ class GeneralizedFrbRustBinding {
   }
 
   /// {@macro flutter_rust_bridge.only_for_generated_code}
-  void initFrbDartApiDl() =>
-      _binding.frb_init_frb_dart_api_dl(ffi.NativeApi.initializeApiDLData);
+  void initShutdownWatcher() {
+    _shutdownWatcher = ShutdownWatcher(_binding.frb_shutdown_callback);
```
fzyzcjy, line 41: nit: add `assert (_shutdownWatcher == null);` as an extra sanity check

frb_dart/lib/src/generalized_frb_rust_binding/_io.dart
```diff
@@ -5,6 +5,17 @@ import 'package:flutter_rust_bridge/src/generalized_uint8list/generalized_uint8l
 import 'package:flutter_rust_bridge/src/platform_types/_io.dart';
 import 'package:flutter_rust_bridge/src/platform_types/platform_types.dart';
 
+final class ShutdownWatcher implements ffi.Finalizable {
```
fzyzcjy, line 8: nit: maybe move this to bottom of this file

fzyzcjy, line 8: nit: Make it a private class, i.e. `_ShutdownWatcher`

frb_rust/src/ffi_binding/mod.rs
```diff
@@ -44,3 +44,23 @@ fn vec_resize(vec: &mut Vec<u8>, new_len: i32) {
...
+    use std::ffi::c_void;
+
+    use allo_isolate::{store_dart_post_cobject, ffi::DartCObject};
+
+    /// Called by Dart's NativeFinalizer on isolate group shutdown.
+    #[unsafe(no_mangle)]
+    pub unsafe extern "C" fn frb_shutdown_callback(_: *mut c_void) {
+
+        unsafe extern "C" fn devnull(_: i64, _: *mut DartCObject) -> bool {
+            true
```
fzyzcjy, line 59: hmm, shall we do `true` or `false` here

From https://github.com/dart-lang/sdk/blob/6ad56c68dc01466d1a51fbed1ac60fd82ca648f6/runtime/include/dart_native_api.h#L115:

> If false is returned, the message was not enqueued and ownership of external typed data in the message remains with the caller

It seems false is a bit more reasonable, since we really mean the message remains with the caller

frb_dart/lib/src/generalized_frb_rust_binding/_io.dart
```diff
@@ -5,6 +5,17 @@ import 'package:flutter_rust_bridge/src/generalized_uint8list/generalized_uint8l
...
+  static var _finalizer;
+
+  ShutdownWatcher(ffi.Pointer<ffi.NativeFinalizerFunction> callback) {
+    _finalizer = ffi.NativeFinalizer(callback);
```
fzyzcjy, line 12: nit: maybe we can follow the pattern shown in https://api.flutter.dev/flutter/dart-ffi/NativeFinalizer-class.html to handle this, e.g. demo there uses

```
  static final _finalizer =
      NativeFinalizer(_nativeDatabaseBindings.closeDatabaseAddress.cast());
```

etc

fzyzcjy (commented):
Great job! Only a logical issue and some nits (and need a bit of testing as you mentioned) :)

frb_dart/lib/src/generalized_frb_rust_binding/_io.dart
alexlapa: You are right that it should be `??=`, but cant do assert here. There might be multiple frb's in scope. Like your project is using frb, and one of your dependencies also use frb. So `initShutdownWatcher` will end up being called multiple times, but for different libraries and we want `_ShutdownWatcher` to be either a global variable or a static field so there will be only one.

frb_dart/lib/src/generalized_frb_rust_binding/_io.dart
alexlapa: We dont have `ExternalLibrary` available in a static context - it is provided via `GeneralizedFrbRustBinding` constructor, so can't make it exactly like this. But having `_ShutdownWatcher` as a global variable or a static field should have the same guarantees.

frb_rust/src/ffi_binding/mod.rs
alexlapa: Not sure about this. While returning `false` is technically correct that would make `StreamSink::add` to error and that might trigger some logging, requests to error tracking services, maybe there will even be some panic in user code. And at this point application is shutting down and we just want it to die silently.

I've added doc comment to explain this. Can change it to return `false` if you want, that's not a big deal.

fzyzcjy: Ah I see, good point! But if we have true, then we are leaking memory...

frb_dart/lib/src/generalized_frb_rust_binding/_io.dart
fzyzcjy: I see. Could you please also make the reasoning in comments - then it would be more clear why everything is like that for future readers

fzyzcjy:
Btw, feel free to mark it `ready for review` and ping me when ready!

frb_rust/src/ffi_binding/mod.rs
alexlapa: > But if we have true, then we are leaking memory...

Well, that's also does not really matter since memory will be reclaimed anyway when shutdown is completed. 

I think it't ready for review, but let me know if you want this changed.

frb_rust/src/ffi_binding/io.rs
```diff
@@ -29,3 +36,44 @@ pub unsafe extern "C" fn frb_free_wire_sync_rust2dart_dco(value: WireSyncRust2Da
...
+/// # Safety
+///
+/// This function should never be called manually.
+#[no_mangle]
+pub unsafe extern "C" fn frb_get_shutdown_callback() -> unsafe extern "C" fn(*mut c_void) {
+    /// Counter for how many active shutdown callbacks there are.
+    ///
+    /// It is incremented when `frb_shutdown_callback` is returned and
+    /// decremented when it is called.
+    static ISOLATES_NUM: AtomicU64 = AtomicU64::new(0);
```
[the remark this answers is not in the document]
alexlapa: Don`t think so. 

So the problem is that `RustLib` can be initialized in multiple isolates so there will be multiple `ShutdownCallback`s created on the dart side (despite it being `static`), and we don't want to disable `Dart_PostCObject` on first isolate shutdown, because there are others still running. E.g. this happened on CI when running tests cause Dart runs each file in separate isolate.

So to be able to run shutdown callback only once we would either need to be sure that `ShutdownCallback` was created in the "main" (first) isolate, which's lifetime is the same as application's. Or we would need some kind a isolate-agnostic-static-singleton, so there can be only one instance even if frb is initialized in multiple isolates. And I'm not sure anything like this is possible.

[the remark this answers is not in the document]
fzyzcjy: Good point! Then maybe we just add a comment to discuss why there must be multiple ones.

frb_rust/src/ffi_binding/io.rs
fzyzcjy, line 1: btw, is it possible we add a test to check this behavior? (I guess not easy...)

frb_rust/src/ffi_binding/io.rs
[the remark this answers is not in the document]
alexlapa: I don't think this changes anything hot-restart wise.

The `frb_shutdown_callback` will be called on hot-restart, and it will disable `Dart_PostCObject`. But hot-restart also drops any Dart-side `ReceiverPort`s so they stop working anyway. And `Dart_PostCObject` will be re-enabled on the next `RustLib.init` call. 

But let me know if I'm wrong in my assumptions, I get you being cautious about a thing that can silently break everything :)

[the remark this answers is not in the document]
fzyzcjy: I see, looks reasonable!

> I get you being cautious about a thing that can silently break everything :)

Yes exactly - I am trying to discuss whatever that is in my mind to ensure we are getting every detail right.

frb_rust/src/ffi_binding/io.rs
alexlapa: Welp, there are few thing we can do.

1. Actually confirming that the problem exists and that this PR fixes it. So for that we would need to see segfault on CI, and I can try that but I'm not sure that's going to happen. The only way I've been able to reproduce it is by closing the app on macOS (linux is fine and I haven't check Windows). And the problem is that it can only happen when app is closed via the red "quit" button, `Ctrl-C`'ing, `kill`ing, closing via `Activity Monitor` won't trigger segfault. So there is definitely something about  letting Dart run this "graceful" shutdown and I don't think we can achieve that on CI. Alternatively, creating and closing dart-vm on the native side might work. Cause what we actually need is to call `Dart_Cleanup` while there are `Dart_PostCObject` with `external_typed_data` payload in progress. But setting up this kind of test seems like too much effort tbh.

2. Making sure that `Dart_PostCObject` stops working after isolate with `RustLib` is killed. That would probably need some feature-gated `extern "C"` functions on the Rust side so it would be able to do some ffi avoiding frb. Should be doable.

Let me know what you think about it.

fzyzcjy: Hmm totally agree, looks not easy... So maybe just merge it now, and if you are interested in it, feel free to add tests to make it more guaranteed!

Issue #2588, opened 2025-03-04
"Atomics.wait cannot be called in this context" when using sync API


aran:
### Describe the bug

I have an object "State".

It is imported from another crate via `pub use automerge::sync::State;`

I have: 

```
#[frb(external)]
impl State {
    pub fn new() -> Self {}
}
```

In my `flutter_rust_bridge.yaml`, I have `full_dep: false` and `default_dart_async: false`.

My Dart code calls `State()` to build a new instance of this type, in flutter web, not using wasm.

Intermittently, this causes:

```
══╡ EXCEPTION CAUGHT BY MAIN ZONE ╞═════════════════════════════════════════════════════════════════
js_primitives.dart:28 The following PanicException was thrown:
js_primitives.dart:28 PanicException(EXECUTE_SYNC_ABORT RuntimeError: Atomics.wait cannot be called in this context
js_primitives.dart:28 pkg/rust_lib_bg.wasm 1:2006906                                                     1412
js_primitives.dart:28 pkg/rust_lib_bg.wasm 1:2255675                                                     2494
js_primitives.dart:28 pkg/rust_lib.js 230:26
js_primitives.dart:28 frb_pde_ffi_dispatcher_sync
js_primitives.dart:28 packages/flutter_rust_bridge/src/generalized_frb_rust_binding/_web.dart 37:12
js_primitives.dart:28 pdeFfiDispatcherSync
js_primitives.dart:28 packages/flutter_rust_bridge/src/codec/pde.dart 24:37                              pdeCallFfi
js_primitives.dart:28 packages/plans/src/rust/frb_generated.dart 1964:16                                 <fn>
js_primitives.dart:28 packages/flutter_rust_bridge/src/main_components/handler.dart 25:24                executeSync
...
```

Debug version:
```
══╡ EXCEPTION CAUGHT BY MAIN ZONE ╞═════════════════════════════════════════════════════════════════
js_primitives.dart:28 The following PanicException was thrown:
js_primitives.dart:28 PanicException(EXECUTE_SYNC_ABORT RuntimeError: Atomics.wait cannot be called in this context
js_primitives.dart:28 pkg/rust_lib_bg.wasm 1:11453490
js_primitives.dart:28 core::core_arch::wasm32::atomic::memory_atomic_wait32::h7ba7f7f3997c71f5
js_primitives.dart:28 pkg/rust_lib_bg.wasm 1:9062326
[...]

PR #2662, opened 2025-03-27
RFC: feat: allow omission of dart-opaque feature

### LINKED ISSUE #2661: Allow omission of dart-opaque feature
If you are not using some of the FRB features it should be possible to omit them in Cargo.toml and avoid the related codegen.

In this case, it is difficult to omit dart-opaque because some generated code expects related methods in the handler.

aran:
## Changes

Fixes #2661 

In this implementation `flutter_rust_bridge::frb_generated_boilerplate_io` and web are still present in the generated code, and now they have a reference to a `dart-opaque` feature that isn't present by default in the client's Cargo.toml. This can be worked around by adding the feature declaration. Fixing this so the code isn't generated at all would be a much bigger change.


aran:
Let me know if you have a preferred solution for the warning generated because of the 'dart-opaque' feature reference. I can take a look at it. I don't know what would be a good way to structure feature-gated generated code.

Overall I'm trying to understand the FRB code better and trying to minimize surface area on the way to solving https://github.com/fzyzcjy/flutter_rust_bridge/issues/2588.

fzyzcjy:
> Let me know if you have a preferred solution 

Not checked in detail but usually I just add more feature gates on it...

fzyzcjy (commented):
Briefly checked it, what about removing the `feat_no_dart_opaque` example, and instead add a test that modifies dart_minimal's rust features temporarily and ensure it compiles

frb_example/dart_minimal/rust/Cargo.toml
```diff
@@ -7,7 +7,20 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-flutter_rust_bridge = { path = "../../../frb_rust" }
+flutter_rust_bridge = { path = "../../../frb_rust", default-features = false, features = [
+    "anyhow",
+    # "dart-opaque",
```
fzyzcjy, line 12:
```suggestion
    # Deliberately remove this feature to ensure frb can work without it
    # "dart-opaque",
```

fzyzcjy (commented):
It would be great to keep dart_minimal have the feature by default (to make users easier to use dart_minimal to reproduce some bugs), but yes your approach looks also ok :)

frb_example/dart_minimal/rust/Cargo.toml
```diff
@@ -7,7 +7,20 @@ edition = "2021"
...
+    "user-utils",
+    "wasm-start",
+] }
 
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(frb_expand)'] }
+
+[features]
+default = ["dart-opaque"]
+dart-opaque = ["flutter_rust_bridge/dart-opaque"]
```
fzyzcjy, line 26: nit: is it possible we make it as simple as possible, say, is it possible to remove the dart-opaque line, and directly specify as `flutter_rust_bridge/dart-opaque`? (just a quick nit, I forget whether rust supports so or not, if not then ignore)

fzyzcjy:
Great job! I will handle it soon (hopefully within a few days)

frb_example/dart_minimal/rust/Cargo.toml
aran: I don't know of another way to do this.

aran:
Github isn't letting me push more commits to this with the merge. Created https://github.com/fzyzcjy/flutter_rust_bridge/pull/2681.

PR #2664, opened 2025-03-27
feat: allow configuring rust opaque implementation

### LINKED ISSUE #2663: Allow setting rust opaque codec to Nom without full_dep
### Describe the bug

In https://github.com/fzyzcjy/flutter_rust_bridge/issues/2588 I encounter concurrency-related issues involving reference counting on the web platform. 

The issue is seemingly avoided by forcing `generate_default_rust_opaque_codec` to `RustOpaqueCodecMode::Nom`. Of course this could be wrong and just making the issue more rare. 

In looking at this code, I am not sure what "nom" is vs "moi", but it appears that the implementation of "nom" is more straightforward and uses "Arc" directly whereas "moi" re-implements part of the reference counting of Arc. I admittedly don't have a complete diagnosis of what is going wrong. 

That said, I would like to be able to configure FRB to use `RustOpaqueCodecMode::Nom` without all the other implications of `full_dep`. Or, if there is no advantage to the "Moi" implementation, perhaps it can just be removed so there is nothing to configure.

### Steps to reproduce

Continuation of https://github.com/fzyzcjy/flutter_rust_bridge/issues/2588

### Logs

```shell
n/a
```

### Expected behavior

_No response_

### Generated binding code

```shell

```

### OS

_No response_

_No response_

### Flutter info

```shell

```

### Version of `clang++`

_No response_

### Additional context

_No response_

aran:
## Changes

Fixes #2663


frb_codegen/src/binary/commands.rs
```diff
@@ -145,6 +145,9 @@ pub(crate) struct GenerateCommandArgsPrimary {
     #[arg(long)]
     pub full_dep: bool,
 
+    /// Configure rust opaque implementation, "moi" or "nom"
+    pub rust_opaque: Option<String>,
```
fzyzcjy, line 149: nit:

1. Maybe name it default_rust_opaque_codec
2. maybe make it type `Option<RustOpaqueCodecMode>`

fzyzcjy:
Good job! Just some nits

fzyzcjy:
Looks great! I will also fix CI and merge it in the next batch together with other PRs and issues. (If you/anyone need it immediately, either use this branch or directly ping me and I will do it immediately)

PR #2681, opened 2025-04-07
feat: (continued) allow omission of dart-opaque feature

aran:
## Changes

Fixes https://github.com/fzyzcjy/flutter_rust_bridge/issues/2661

In this implementation flutter_rust_bridge::frb_generated_boilerplate_io and web are still present in the generated code, and now they have a reference to a dart-opaque feature that isn't present by default in the client's Cargo.toml. This can be worked around by adding the feature declaration. Fixing this so the code isn't generated at all would be a much bigger change.


aran:
This addresses https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/14270290721/job/40001795109?pr=2662 by adding a `dart-opaque` feature across all the Cargo.toml.

The "pseudo-manual" tests using dart-opaque aren't fixed yet, I'm not quite sure what's wrong there or what would be a good way to approach it.

fzyzcjy:
Hmm, firstly do we have this symbol or is it accidentially removed?

aran:
> Hmm, firstly do we have this symbol or is it accidentially removed?

That symbol is supposed to be gated on `dart-opaque`, so need to find all the places where `dart-opaque` is needed but not enabled. Found one more in 3682d0c9eb966928ea5374df7eaa43f7f7f6adbf.

fzyzcjy:
Looks good! Now we only have 2 failing checks and seems it is still related

image

aran:
I had to dig into a couple screens to figure out why the ubuntu CI jobs were "canceled". They failed with:

```
This is a scheduled Ubuntu 20.04 brownout. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101
```

fzyzcjy:
Ah, looks like we need to upgrade the ubuntu version in the CI. Do you have interest in making a PR for this?

aran:
> Ah, looks like we need to upgrade the ubuntu version in the CI. Do you have interest in making a PR for this?

I put up a quick find/replace PR but if there are issues due to the change I won't have time this week to look into them

fzyzcjy:
Wow I did not realize CI became green, good job! I will have a review on it hopefully within a day.

frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template
```diff
@@ -11,3 +11,6 @@ flutter_rust_bridge = REPLACE_ME_RUST_FRB_DEPENDENCY
 
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(frb_expand)'] }
+
+[features]
+dart-opaque = ["flutter_rust_bridge/dart-opaque"]
```
fzyzcjy, line 16: nit: this file is for all new projects' Cargo.toml, so it would be great to keep it unchanged

EDIT: Oh I see the reason

>  This can be worked around by adding the feature declaration.

Hmm, however, it would be great if we do not require all user code to have such a new feature in Cargo.toml :(

EDIT: This seems interesting https://stackoverflow.com/questions/64546459/how-can-my-crate-check-the-selected-features-of-a-dependency - is it possible we somehow utilize this approach

frb_example/dart_minimal/rust/Cargo.toml
```diff
@@ -7,7 +7,21 @@ edition = "2021"
...
+    "user-utils",
+    "wasm-start",
+] }
 
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(frb_expand)'] }
+
+[features]
+default = ["dart-opaque"]
+dart-opaque = ["flutter_rust_bridge/dart-opaque"]
```
fzyzcjy, line 27: nit: what about naming it `flutter-rust-bridge--dart-opaque` or something like that, so that it is more clear we are forwarding to there when we read our test infra code.

frb_example/pure_dart/rust/Cargo.toml
```diff
@@ -8,7 +8,12 @@ crate-type = ["cdylib"]
 
 [dependencies]
 anyhow = { version = "1.0.64", features = ["backtrace"] }
-flutter_rust_bridge = { path = "../../../frb_rust", features = ["chrono", "uuid", "backtrace"] }
+flutter_rust_bridge = { path = "../../../frb_rust", features = [
+    "backtrace",
+    "chrono",
+    "dart-opaque",
```
fzyzcjy, line 14: nit: given that dart_opaque is a default feature, maybe we can keep these lines unchanged

frb_example/pure_dart/rust/Cargo.toml
```diff
@@ -26,5 +31,7 @@ tokio = { version = "1.34.0", features = ["rt"] }
 
 [features]
 internal_feature_for_testing = []
+dart-opaque = ["flutter_rust_bridge/dart-opaque"]
```
fzyzcjy, line 34: nit: is it possible we delete this line (same as above)

frb_example/pure_dart/build.dart
```diff
@@ -2,4 +2,4 @@ import 'package:flutter_rust_bridge_utils/flutter_rust_bridge_utils.dart';
 
 void main(List<String> args) async => simpleBuild(args,
     // This is solely used for testing
-    features: ['internal_feature_for_testing']);
+    features: ['dart-opaque', 'internal_feature_for_testing']);
```
fzyzcjy, line 5: nit: (same, can we del it)

frb_codegen/assets/integration_template/shared/REPLACE_ME_RUST_CRATE_DIR/Cargo.toml.template
aran: That's an interesting hack. Never would have thought of it. Not sure if it's a good idea but I think it might work as long as the number of actually-needed feature combinations was small. 

I would generalize that suggestion and say that it comes down to modifying `flutter_rust_bridge::frb_generated_boilerplate_io` and `flutter_rust_bridge::frb_generated_boilerplate_web` not to generate the code requiring `dart-opaque`. Perhaps that StackOverflow macro approach is the best way to do that, or there might be another way.

aran:
Unfortunately I'm not going to be able to prioritize reworking this to use another approach so I'm going to close it out and maybe revisit another time. 

A couple of the semi-related commits like ae8c875821b3327477cf4f0951b9e23668434c39 and a75df11b9e1fb9658b4d4b2e41c9ca209eb6921f  might be independently useful even without the `dart-opaque` change, not sure.

frb_example/dart_minimal/rust/Cargo.toml
fzyzcjy: Oh yes, the "EDIT" above is the last one, so we may use that approach etc.

fzyzcjy:
Sad to hear that :( 

> A couple of the semi-related commits like https://github.com/fzyzcjy/flutter_rust_bridge/commit/ae8c875821b3327477cf4f0951b9e23668434c39 and https://github.com/fzyzcjy/flutter_rust_bridge/commit/a75df11b9e1fb9658b4d4b2e41c9ca209eb6921f might be independently useful even without the dart-opaque change, not sure.

Yes that looks useful - do you have interest in making one (or two) separate PRs for these?

PR #2716, opened 2025-05-08
Update users.md

hicaru:
added ZilPay wallet, who use it


fzyzcjy:
Thank you, but that list is automatically generated instead of manually updated. I will re-run the script to update it later.

PR #2718, opened 2025-05-09
Generate Freezed classes as sealed

### LINKED ISSUE #2715: Support for Freezed 3
When using Freezed, generated classes are neither `sealed` nor `abstract`, thus making it incompatible with Freezed 3, where they are a requirement. [^1]

I think it'd be fine to just mark all generated classes as sealed? That would still work even with Freezed 2, although I fear it may introduce a breaking change.

For now I'm sticking to Freezed 2 - as editing the files manually would be nightmarish

**Additional context**
[^1]: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/migration_guide.md

borgoat:
## Changes

Fixes #2715 - 

Fixes https://github.com/fzyzcjy/flutter_rust_bridge/issues/2714

## Procedure and Checklist

Utility commands


fzyzcjy:
Good job! Could you please check this part to ensure it works (at least minimally)?

image

borgoat:
It seems `dart_minimal` works properly, but it also doesn't use `freezed` as far as I can see (although it is a dependency)

I was trying to fix the rest of the CI though, at least `pure_dart` is definitely impacted by this change. I was struggling to make it work locally, but now I think it's just because I'm using Dart 3.7.2 (CI is on 3.6.2), and the native assets API changed: it expects `build.dart` under `hook/build.dart` now.

fzyzcjy:
> but it also doesn't use freezed as far as I can see

You need to add a few demos into it (here maybe just copy-paste some interesting tests in pure_dart) and check it still works.

> but now I think it's just because I'm using Dart 3.7.2 (CI is on 3.6.2), and the native assets API changed: it expects build.dart under hook/build.dart now.

Yes, I have not upgraded this yet, since it is not directly used by end users.

vdawg-git:
Also fixes https://github.com/fzyzcjy/flutter_rust_bridge/issues/2714

fzyzcjy:
@all-contributors please add @borgoat for code

PR #2721, opened 2025-05-14
Implement `#[frb(ignore)]` for struct fields

### LINKED ISSUE #2720: Ability to disable auto-accessors from opaque types
#### Is your feature request related to a problem? Please describe.

As a bit of background, we're trying to share a set of types between our Android app and another device running Rust software which define their communication protocol.  Some of the types contains other types which cannot be serialized and which are not `Clone`, but we only need access to the fields from the Rust side so we decided to make the types opaque so the Flutter side would only get a smart pointer to the types.

The problem we have is that since the types are in a shared crate all fields are public, which means that FRB's codegen automatically generates accessors for all fields.  This in turn means that our inner types which cannot be serialized get automatic accessors and FRB's codegen tries to expose the type to the Flutter side.  Even explicitly flagging the inner types as opaque as well doesn't help because that requires the types to be `Clone` which they are not.  And since these types are shared we cannot wrap the inner fields in `RustAutoOpaque`.

#### Describe the solution you'd like

An FRB codegen attribute to disable auto-accessors, on individual fields and/or on the type as a whole.

This would be our preferred solution as we don't by default want Flutter code to be able to access most of these fields.

#### Describe alternatives you've considered

Alternatively an attribute that would conditionally wrap the fields in `RustAutoOpaque` could work, but that would get very messy and would mean that any code that should work in both build configurations would get very complicated.

kyrias:
## Changes

Fixes #2720.

This additionally adds a `#[frb(ignore_all)]` attribute to skip all fields, but I could drop it if you'd prefer not to add that.

## Procedure and Checklist

- [x] Add tests in `frb_example/dart_minimal`.

Utility commands

### DIFF
--- a/frb_codegen/src/library/codegen/ir/mir/field.rs
+++ b/frb_codegen/src/library/codegen/ir/mir/field.rs
@@ -18,6 +18,7 @@ pub struct MirField {
 #[derive(Deserialize, Default)]
 pub struct MirFieldSettings {
     pub is_in_mirrored_enum: bool,
+    pub skip_accessors: bool,
 }
 }
 
--- a/frb_codegen/src/library/codegen/parser/mir/parser/attribute.rs
+++ b/frb_codegen/src/library/codegen/parser/mir/parser/attribute.rs
@@ -98,6 +98,10 @@ impl FrbAttributes {
         self.any_eq(&FrbAttribute::Ignore)
     }
 
+    pub(crate) fn ignore_all(&self) -> bool {
+        self.any_eq(&FrbAttribute::IgnoreAll)
+    }
+
     pub(crate) fn unignore(&self) -> bool {
         self.any_eq(&FrbAttribute::Unignore)
     }
@@ -278,6 +282,7 @@ mod frb_keyword {
     syn::custom_keyword!(setter);
     syn::custom_keyword!(init);
     syn::custom_keyword!(ignore);
+    syn::custom_keyword!(ignore_all);
     syn::custom_keyword!(unignore);
     syn::custom_keyword!(opaque);
     syn::custom_keyword!(non_opaque);
@@ -327,6 +332,7 @@ enum FrbAttribute {
     External,
     Getter,
     Ignore,
+    IgnoreAll,
     Unignore,
     Init,
     Mirror(FrbAttributeMirror),
@@ -381,6 +387,7 @@ impl Parse for FrbAttribute {
             .or_else(|| parse_keyword::<setter, _>(input, &lookahead, setter, Setter))
             .or_else(|| parse_keyword::<init, _>(input, &lookahead, init, Init))
             .or_else(|| parse_keyword::<ignore, _>(input, &lookahead, ignore, Ignore))
+            .or_else(|| parse_keyword::<ignore_all, _>(input, &lookahead, ignore_all, IgnoreAll))
             .or_else(|| parse_keyword::<unignore, _>(input, &lookahead, unignore, Unignore))
             .or_else(|| parse_keyword::<opaque, _>(input, &lookahead, opaque, Opaque))
             .or_else(|| parse_keyword::<non_opaque, _>(input, &lookahead, non_opaque, NonOpaque))
@@ -828,6 +835,11 @@ mod tests {
     #[test]
     fn test_ignore() {
         simple_keyword_tester("ignore", FrbAttribute::Ignore);
+    }
+
+    #[test]
+    fn test_ignore_all() {
+        simple_keyword_tester("ignore_all", FrbAttribute::IgnoreAll);
     }
 
     #[test]
--- a/frb_codegen/src/library/codegen/parser/mir/parser/function/auto_accessor/mod.rs
+++ b/frb_codegen/src/library/codegen/parser/mir/parser/function/auto_accessor/mod.rs
@@ -92,7 +92,11 @@ fn parse_auto_accessors_of_struct(
     let ty_struct = &type_parser.struct_pool()[&ty_struct_ident].to_owned();
 
     (ty_struct.fields.iter())
-        .filter(|field| field.is_rust_public.unwrap() && !is_ty_opaque_reference_type(&field.ty))
+        .filter(|field| {
+            field.is_rust_public.unwrap()
+                && !field.settings.skip_accessors
+                && !is_ty_opaque_reference_type(&field.ty)
+        })
         .flat_map(|field| {
             [MirFuncAccessorMode::Getter, MirFuncAccessorMode::Setter]
                 .into_iter()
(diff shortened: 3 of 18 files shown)


kyrias:
> Good job! Could you please add some demos in the frb_example/dart_minimal as mentioned in the "Procedure and Checklist" above

I got confused as to how exactly they should be added since there weren't any in there right now, but sure, I can add some.

> I am not very sure whether we should add the ignore_all, because after that, users can add both ignore and ignore_all on a struct, and then users will get confused :(

What would be confusing about it?  There's no difference between this and e.g. `serde`'s `rename` and `rename_all` attributes.  And it's rather user-unfriendly to have to add the same attribute on every field of a struct.

kyrias:
Additionally the instructions in the template don't work.  `./frb_internal test-dart-native --package frb_example/dart_minimal` gives me the following error:

```
> /bin/sh -c dart --enable-experiment=native-assets  test  (pwd: /home/remmy/src/e/flutter_rust_bridge/frb_example/dart_minimal, env: {CARGO_TERM_COLOR: always})
"/home/remmy/src/e/flutter_rust_bridge/frb_example/dart_minimal/hook/build.dart" file not found.
Building native assets for package:frb_example_dart_minimal failed.
Compilation of hook returned with exit code: 255.
To reproduce run:
/home/remmy/.local/share/flutter/bin/cache/dart-sdk/bin/dart compile kernel --packages=/home/remmy/src/e/flutter_rust_bridge/frb_example/dart_minimal/.dart_tool/package_config.json --output=/home/remmy/src/e/flutter_rust_bridge/frb_example/dart_minimal/.dart_tool/native_assets_builder/frb_example_dart_minimal/649d50974e631e218f6ed46b9c0f7bba/hook.dill --depfile=/home/remmy/src/e/flutter_rust_bridge/frb_example/dart_minimal/.dart_tool/native_assets_builder/frb_example_dart_minimal/649d50974e631e218f6ed46b9c0f7bba/hook.dill.d /home/remmy/src/e/flutter_rust_bridge/frb_example/dart_minimal/hook/build.dart
stderr:
"/home/remmy/src/e/flutter_rust_bridge/frb_example/dart_minimal/hook/build.dart" file not found.

stdout:

        
Error: Compiling native assets failed.
Unhandled exception:
ProcessException: Bad exit code (65). If you want to see extra information, set FRB_DART_RUN_COMMAND_STDERR=1
  Command: /bin/sh -c dart --enable-experiment=native-assets  test 
#0      runCommand (package:flutter_rust_bridge/src/cli/run_command.dart:67:5)
<asynchronous suspension>
#1      SimpleExecutor.call (package:flutter_rust_bridge_internal/src/utils/makefile_dart_infra.dart:32:12)
<asynchronous suspension>
#2      testDartNative.<anonymous closure>.<anonymous closure> (package:flutter_rust_bridge_internal/src/makefile_dart/test.dart:361:9)
<asynchronous suspension>
#3      wrapMaybeSetExitIfChangedRaw (package:flutter_rust_bridge_internal/src/makefile_dart/generate.dart:420:3)
<asynchronous suspension>
#4      testDartNative.<anonymous closure> (package:flutter_rust_bridge_internal/src/makefile_dart/test.dart:360:7)
<asynchronous suspension>
#5      withLlvmCovReport (package:flutter_rust_bridge_internal/src/makefile_dart/test.dart:387:12)
<asynchronous suspension>
#6      testDartNative (package:flutter_rust_bridge_internal/src/makefile_dart/test.dart:339:3)
<asynchronous suspension>
#7      SimpleConfigCommand.run (package:flutter_rust_bridge_internal/src/utils/makefile_dart_infra.dart:88:31)
<asynchronous suspension>
#8      CommandRunner.runCommand (package:args/command_runner.dart:212:13)
<asynchronous suspension>
#9      main (file:///home/remmy/src/e/flutter_rust_bridge/tools/frb_internal/bin/flutter_rust_bridge_internal.dart:30:3)
<asynchronous suspension>
```

fzyzcjy:
> What would be confusing about it? There's no difference between this and e.g. serde's rename and rename_all attributes. And it's rather user-unfriendly to have to add the same attribute on every field of a struct.

It seems we already support `#[frb(ignore)] pub struct S {}`. Thus wondering what's the difference of `#[frb(ignore_all)] pub struct S {}`.

fzyzcjy:
> hook/build.dart" file not found.

Hmm, maybe ensure Dart version is the same as in the CI

kyrias:
> > I got confused as to how exactly they should be added since there weren't any in there right now, but sure, I can add some.
>
> That's the point of "minimal". You just add some into there temporarily because that's quick to do, and I will later move them to frb_example/pure_dart and handle all auto generated things etc. But if you like you can directly do the latter as well.

The problem is that I'm not a Dart developer so I didn't really know what I need to do to "just" add a test case since there seemed to be no relevant test cases in there for me to mimic, but I see 

> > What would be confusing about it? There's no difference between this and e.g. serde's rename and rename_all attributes. And it's rather user-unfriendly to have to add the same attribute on every field of a struct.
>
> It seems we already support `#[frb(ignore)] pub struct S {}`. Thus wondering what's the difference of `#[frb(ignore_all)] pub struct S {}`.

The former ignores the type, the latter adds `#[frb(ignore)]` to all fields, the exact same way that serde's attributes I referenced work.

fzyzcjy:
> The problem is that I'm not a Dart developer so I didn't really know what I need to do to "just" add a test case since there seemed to be no relevant test cases in there for me to mimic, but I see what I was missing now.

Oh there is a simple addition function you can mimic

fzyzcjy:
> The former ignores the type, the latter adds #[frb(ignore)] to all fields, the exact same way that serde's attributes I referenced work.

Ah I see! Could you please provide a link? I quickly searched and does not find that https://serde.rs/field-attrs.html

kyrias:
> > The former ignores the type, the latter adds #[frb(ignore)] to all fields, the exact same way that serde's attributes I referenced work.
>
> Ah I see! Could you please provide a link? I quickly searched and does not find that https://serde.rs/field-attrs.html

That page only lists the field attributes, and the one I mentioned is at the very top, `#[serde(rename = "name")] `.  The [Container attributes](https://serde.rs/container-attrs.html) page lists the.. container attributes and includes both `#[serde(rename = "name")] `, which affects the type itself, and `#[serde(rename_all = "...")] ` which applies the equivalent attribute to all fields.

fzyzcjy:
Get it. Feel free to ping me when needing a code review!

fzyzcjy:
Ignore the general internal thing, I will handle it now

fzyzcjy:
I am handling the rest in #2725. Btw do you have interest in updating doc for this? (Just create a new PR)

kyrias:
Sure, I can take a stab at that.

PR #2732, opened 2025-05-16
Add demo to use dart callbacks in other threads

fzyzcjy:
## Changes

Fixes #the-issue-number-here

## Procedure and Checklist

Utility commands


fzyzcjy:
seems no need

PR #2751, opened 2025-06-03
feat: bounded thread pool

### LINKED ISSUE #2749: Support bounding the size of the worker pool
On Web, if we run many async-dart sync-rust operations at once, each one spawns a worker from the web worker pool, which grows on demand. Each worker takes up memory proportional to the size of our Rust code, which can lead to out of memory errors. 

It would be ideal to have a built-in configuration that prevents growing the worker pool beyond some configuration, with a work queue.

I tried to see if this could be done with a custom handler on Rust side without code changes to FRB, but I don't see a way. The issue is the handler needs a way to subscribe to the availability of a worker in the worker pool and send work to it when available, and I don't see a way.

It may be possible to handle this with a different Dart-side handler, i.e. via a `executeNormal` override that uses the `pool` package. It's not ideal though because it can't coordinate with actual worker pool capacity.

**Additional context**
We are trying to write code that is safely portable across web and native. On web there is the hard platform constraint of never blocking the main thread. In the presence of any shared memory, in particular `&mut` refs, there is no safe way to intermingle `#[frb(sync)]` code with any kind of async code, as some kind of blocking is required. Locks are created automatically for `RustAutoOpaque`. Thus for any given `RustAutoOpaque`, or broadly, shared memory data, we must either choose to use only sync functions and avoid the use of the locks or only access it in worker threads where we always use the lock. 

In our code currently we do all sync, but this causes some performance issues as the rust operations contend with the main thread. I tried to switch to all async, but we ran into OOMs because then it spawns a lot of web workers.

aran:
## Changes

Fixes #2749 

## Procedure and Checklist

- [X] Implement the feature / fix the bug.

Utility commands

### DIFF
--- a/frb_codegen/src/binary/commands.rs
+++ b/frb_codegen/src/binary/commands.rs
@@ -200,6 +200,10 @@ pub(crate) struct GenerateCommandArgsPrimary {
     /// List of cargo feature flags to enable when generating
     #[arg(long)]
     pub rust_features: Option<Vec<String>>,
+
+    /// Maximum number of web workers for the worker pool (web only)
+    #[arg(long)]
+    pub web_worker_pool_max_workers: Option<usize>,
 }
 
 #[derive(Debug, Args)]
--- a/frb_codegen/src/library/codegen/generator/wire/rust/spec_generator/misc/mod.rs
+++ b/frb_codegen/src/library/codegen/generator/wire/rust/spec_generator/misc/mod.rs
@@ -53,6 +53,7 @@ pub(crate) fn generate(
             context.config.default_rust_opaque_codec,
             content_hash,
             &context.config.rust_preamble,
+            context.config.web_worker_pool_max_workers,
         ),
         wire_funcs: (context.mir_pack.funcs_with_impl().iter())
             .map(|f| generate_wire_func(f, context))
@@ -174,6 +175,7 @@ fn generate_boilerplate(
     default_rust_opaque_codec: RustOpaqueCodecMode,
     content_hash: i32,
     rust_preamble: &str,
+    web_worker_pool_max_workers: Option<usize>,
 ) -> Acc<Vec<WireRustOutputCode>> {
     let rust_preamble_formatted = if rust_preamble.is_empty() {
         "".to_owned()
@@ -194,18 +196,25 @@ fn generate_boilerplate(
                     .into(),
                 ]
             }
-            TargetOrCommon::Common => vec![format!(
-                r#"{rust_preamble_formatted}flutter_rust_bridge::frb_generated_boilerplate!(
-                    default_stream_sink_codec = {default_stream_sink_codec}Codec,
-                    default_rust_opaque = RustOpaque{default_rust_opaque_codec},
-                    default_rust_auto_opaque = RustAutoOpaque{default_rust_opaque_codec},
-                );
-                pub(crate) const FLUTTER_RUST_BRIDGE_CODEGEN_VERSION: &str = "{version}";
-                pub(crate) const FLUTTER_RUST_BRIDGE_CODEGEN_CONTENT_HASH: i32 = {content_hash};
-            "#,
-                version = env!("CARGO_PKG_VERSION"),
-            )
-            .into()],
+            TargetOrCommon::Common => {
+                let web_worker_config = match web_worker_pool_max_workers {
+                    Some(max) => format!("pub(crate) const CODEGEN_WEB_WORKER_POOL_MAX_WORKERS: Option<usize> = Some({max});"),
+                    None => "pub(crate) const CODEGEN_WEB_WORKER_POOL_MAX_WORKERS: Option<usize> = None;".to_string(),
+                };
+                vec![format!(
+                    r#"{rust_preamble_formatted}flutter_rust_bridge::frb_generated_boilerplate!(
+                        default_stream_sink_codec = {default_stream_sink_codec}Codec,
+                        default_rust_opaque = RustOpaque{default_rust_opaque_codec},
+                        default_rust_auto_opaque = RustAutoOpaque{default_rust_opaque_codec},
+                    );
+                    pub(crate) const FLUTTER_RUST_BRIDGE_CODEGEN_VERSION: &str = "{version}";
+                    pub(crate) const FLUTTER_RUST_BRIDGE_CODEGEN_CONTENT_HASH: i32 = {content_hash};
+                    {web_worker_config}
+                "#,
+                    version = env!("CARGO_PKG_VERSION"),
+                )
+                .into()]
+            }
         }
     })
 }
--- a/frb_rust/src/for_generated/boilerplate.rs
+++ b/frb_rust/src/for_generated/boilerplate.rs
@@ -225,7 +225,17 @@ macro_rules! frb_generated_default_handler {
 
         #[cfg(target_family = "wasm")]
         thread_local! {
-            pub static THREAD_POOL: $crate::for_generated::SimpleThreadPool = Default::default();
+            pub static THREAD_POOL: $crate::for_generated::SimpleThreadPool = {
+                match CODEGEN_WEB_WORKER_POOL_MAX_WORKERS {
+                    Some(max) => {
+                        $crate::for_generated::SimpleThreadPool::new_with_max_workers(max)
+                            .expect("Failed to create bounded WorkerPool")
+                    }
+                    None => {
+                        Default::default()
+                    }
+                }
+            };
         }
 
         #[cfg(target_family = "wasm")]
--- a/frb_rust/src/for_generated/mod.rs
+++ b/frb_rust/src/for_generated/mod.rs
@@ -80,3 +80,4 @@ pub use misc_rust_async::*;
 pub use pointer::*;
 #[cfg(target_family = "wasm")]
 pub use wasm_bindgen;
+
--- a/frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
+++ b/frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
@@ -6,6 +6,8 @@ use crate::misc::web_utils::script_path;
 use crate::web_transfer::transfer_closure::TransferClosure;
 use js_sys::{Array, Object, Reflect};
 use std::cell::RefCell;
+use std::collections::VecDeque;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::iter::FromIterator;
 use std::rc::Rc;
 use wasm_bindgen::prelude::*;
@@ -26,6 +28,19 @@ pub struct WorkerPool {
 struct PoolState {
     workers: RefCell<Vec<Worker>>,
     callback: Closure<dyn FnMut(Event)>,
+    max_workers: Option<usize>,
+    work_queue: RefCell<VecDeque<PendingWork>>,
+    total_workers: AtomicUsize,
+}
+
+struct PendingWork {
+    closure: TransferClosure<JsValue>,
+}
+
+#[derive(Debug)]
+enum ExecuteResult {
+    Executed(Worker),
+    Queued,
 }
 
 #[wasm_bindgen]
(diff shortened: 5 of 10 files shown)


aran:
Possible approach to #2749. Let me know what you think.

Couple things:
 * I had an issue with the new `forceSameCodegenVersion` feature. My `lib/src/rust/frb_generated.dart` had this bit of code in it:
 
 ```
  /// Initialize flutter_rust_bridge
  static Future<void> init({
    RustLibApi? api,
    BaseHandler? handler,
    ExternalLibrary? externalLibrary,
    bool forceSameCodegenVersion = true,
  }) async {
    await instance.initImpl(
      api: api,
      handler: handler,
      externalLibrary: externalLibrary,
      forceSameCodegenVersion: forceSameCodegenVersion,
    );
  }
  ```
  
  However, `instance.initImpl` doesn't have the `forceSameCodegenVersion` parameter on it. I hacked around it to test, but I'm not sure if I configured something wrong or if there's a bug in main.

 * There is a new version of Flutter out and I tested this one on the latest only.
 * Not sure how to test this in the current FRB test framework. I ran it on our own code with and without the the parameter and observed the number of created web workers.

fzyzcjy:
> I had an issue with the new forceSameCodegenVersion feature.

Could you please try to upgrade the frb dart runtime package (to use master)? It is only on master now

> Not sure how to test this in the current FRB test framework. I ran it on our own code with and without the the parameter and observed the number of created web workers.

Yes it is not very trivial to test. One way may be, firstly create a ton of concurrent requests (which you mentioned will create unbounded num threads), and then get the number of thread and assert it is small

frb_codegen/src/binary/commands.rs
```diff
@@ -200,6 +200,10 @@ pub(crate) struct GenerateCommandArgsPrimary {
     /// List of cargo feature flags to enable when generating
     #[arg(long)]
     pub rust_features: Option<Vec<String>>,
+
+    /// Maximum number of web workers for the worker pool (web only)
+    #[arg(long)]
+    pub web_worker_pool_max_workers: Option<usize>,
```
fzyzcjy: nit: for api, I personally suggest to not add it to codegen, but instead just let users write a custom handler

frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
```diff
@@ -26,6 +28,19 @@ pub struct WorkerPool {
 struct PoolState {
     workers: RefCell<Vec<Worker>>,
```
fzyzcjy: nit: it seems this means idle_workers, so maybe we can rename it. (Otherwise, it just reading these several lines and I will think "hey total_workers should be equal to workers.len()")

frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
```diff
@@ -167,14 +209,31 @@ impl WorkerPool {
...
     /// finishes.
     ///
-    /// ## Errors
+    /// ## Returns
     ///
-    /// Returns any error that may happen while a JS web worker is created and a
-    /// message is sent to it.
+    /// - `Ok(ExecuteResult::Executed(worker))` - Work was executed on the returned worker
+    /// - `Ok(ExecuteResult::Queued)` - Work was queued for later execution (when max workers reached)
+    /// - `Err(e)` - An error occurred while creating worker or sending message
```
fzyzcjy: nit: no need to doc here, but just doc `enum ExecuteResult`'s two branches

frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
```diff
@@ -167,14 +209,31 @@ impl WorkerPool {
...
-        closure.apply(&worker).map(|_| worker)
+    fn execute_raw(&self, closure: TransferClosure<JsValue>) -> Result<ExecuteResult, JsValue> {
+        if let Some(worker) = self.try_get_worker() {
+            closure.apply(&worker)?;
+            return Ok(ExecuteResult::Executed(worker));
+        }
+        
+        if self.can_spawn_worker() {
+            let worker = self.spawn()?;
+            self.state.total_workers.fetch_add(1, Ordering::SeqCst);
```
fzyzcjy: nit: wondering whether (may or may not be possible) we can make a stronger abstraction about all these. (currently we put all fields and logic flat in this WorkerPool)

frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
```diff
@@ -167,14 +209,31 @@ impl WorkerPool {
...
+            closure.apply(&worker)?;
+            return Ok(ExecuteResult::Executed(worker));
+        }
```
fzyzcjy: nit: what about

```
let worker = self.try_get_worker().unwrap_or(|| {
  self.can_spawn_worker().then(||self.spawn(); state.fetch_add())
});
if let Some(worker) = worker {
  closure.apply(&worker)?;
  return Ok(ExecuteResult::Executed(worker));
}
```

(may not use unwrap_or etc but use if let; but the main key here is we may unify the "closure.apply + return Executed")

frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
```diff
@@ -193,6 +252,11 @@ impl WorkerPool {
         let slot2 = reclaim_slot.clone();
         let reclaim = Closure::<dyn FnMut(_)>::new(move |_: MessageEvent| {
             if let Some(state) = state.upgrade() {
+                if let Some(pending_work) = state.work_queue.borrow_mut().pop_front() {
+                    let _ = pending_work.closure.apply(&worker2);
+                    return;
```
fzyzcjy: wondering whether it is a bit better to be a if-else branch

frb_rust/src/thread_pool/web.rs
```diff
@@ -1,4 +1,18 @@
-pub use crate::third_party::wasm_bindgen::worker_pool::WorkerPool as SimpleThreadPool;
+use crate::third_party::wasm_bindgen::worker_pool::WorkerPool;
+
+pub struct SimpleThreadPool(WorkerPool);
+
+impl SimpleThreadPool {
+    pub fn new_with_max_workers(max_workers: usize) -> Result<Self, wasm_bindgen::JsValue> {
```
fzyzcjy, line 26: nit: given we have several args here, wondering whether we just provide a config struct, then users can do sth like `StructName::new(MyConfig {max_workers: 42, ..Default::default()})`

frb_codegen/src/binary/commands.rs
aran: ok

frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
aran: This is a little clearer in the newer version but let me know what you have in mind

frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
aran: see if you like the new version better, happy to continue to iterate

frb_rust/src/thread_pool/web.rs
aran: introduced a config struct

frb_codegen/src/library/codegen/generator/wire/rust/spec_generator/misc/mod.rs
```diff
@@ -194,18 +194,20 @@ fn generate_boilerplate(
...
-                    default_rust_opaque = RustOpaque{default_rust_opaque_codec},
-                    default_rust_auto_opaque = RustAutoOpaque{default_rust_opaque_codec},
-                );
-                pub(crate) const FLUTTER_RUST_BRIDGE_CODEGEN_VERSION: &str = "{version}";
-                pub(crate) const FLUTTER_RUST_BRIDGE_CODEGEN_CONTENT_HASH: i32 = {content_hash};
-            "#,
-                version = env!("CARGO_PKG_VERSION"),
-            )
-            .into()],
+            TargetOrCommon::Common => {
```
fzyzcjy, line 197: nit: seems this part is not changed, thus maybe we can revert the change to keep code diff minimal

frb_rust/src/for_generated/boilerplate.rs
```diff
@@ -225,7 +225,9 @@ macro_rules! frb_generated_default_handler {
 
         #[cfg(target_family = "wasm")]
         thread_local! {
-            pub static THREAD_POOL: $crate::for_generated::SimpleThreadPool = Default::default();
+            pub static THREAD_POOL: $crate::for_generated::SimpleThreadPool = {
```
fzyzcjy, line 228: (same)

frb_rust/src/for_generated/mod.rs
```diff
@@ -80,3 +80,4 @@ pub use misc_rust_async::*;
 pub use pointer::*;
 #[cfg(target_family = "wasm")]
 pub use wasm_bindgen;
+
```
fzyzcjy, line 83: (same)

frb_rust/src/thread_pool/web.rs
```diff
@@ -1,4 +1,38 @@
...
+
+impl SimpleThreadPool {
+    pub fn new_with_config(config: SimpleThreadPoolConfig) -> Result<Self, wasm_bindgen::JsValue> {
+        match config.max_workers {
+            Some(max_workers) => WorkerPool::new_with_max_workers(max_workers).map(SimpleThreadPool),
+            None => Ok(SimpleThreadPool(WorkerPool::default())), // Unbounded
+        }
+    }
+
+    pub fn new_with_max_workers(max_workers: usize) -> Result<Self, wasm_bindgen::JsValue> {
```
fzyzcjy, line 26: nit: if we have new_with_config, maybe we do not need this extra constructor

frb_rust/src/thread_pool/web.rs
```diff
@@ -1,4 +1,38 @@
...
+        Self::new_with_config(SimpleThreadPoolConfig { max_workers: Some(max_workers) })
+    }
+}
+
+impl Default for SimpleThreadPool {
```
fzyzcjy, line 31: nit: is it possible to use `#[derive(Default)]` instead of manually impl it

frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
```diff
@@ -42,6 +59,16 @@ impl WorkerPool {
         )
     }
 
+    pub fn new_with_max_workers(max_workers: usize) -> Result<WorkerPool, JsValue> {
```
fzyzcjy, line 62: nit: wondering whether we should move the "new_with_config" thing here, i.e. remove

* new
* new_with_max_workers
* new_raw

and have only one single new_with_config that accepts a config

b/c otherwise we have a lot of constructors here...

For backward compatibility we may have to keep the new and new_raw there, but they can be deprecated

frb_rust/src/thread_pool/web.rs
```diff
@@ -1,4 +1,38 @@
-pub use crate::third_party::wasm_bindgen::worker_pool::WorkerPool as SimpleThreadPool;
+use crate::third_party::wasm_bindgen::worker_pool::WorkerPool;
```
fzyzcjy, line 1: nit: is it possible we still use the "pub use ... as ...", then everything in web.rs can be kept unchanged

fzyzcjy (commented):
Looks great! Only some tiny nits

frb_rust/src/third_party/wasm_bindgen/worker_pool.rs
```diff
@@ -172,9 +216,32 @@ impl WorkerPool {
     /// Returns any error that may happen while a JS web worker is created and a
     /// message is sent to it.
     // NOTE: It is originally named `execute`, but rename to align with crate `threadpool`
-    fn execute_raw(&self, closure: TransferClosure<JsValue>) -> Result<Worker, JsValue> {
-        let worker = self.worker()?;
-        closure.apply(&worker).map(|_| worker)
+    fn execute_raw(&self, closure: TransferClosure<JsValue>) -> Result<ExecuteResult, JsValue> {
```
fzyzcjy, line 219: TODO(@fzyzcjy): carefully recheck the core logic again after code cleanup

(you do not need to do anything to this comment, it is just a reminder for myself)

PR #2735, opened 2025-05-19
Logging overhaul redo

patmuk:
Re-applies the logging overhaul changes to the latest master

## Changes

Fixes #the-issue-number-here

## Procedure and Checklist

In order to quickly iterate and avoid slowing down development pace by making CI pass, only the following simplified steps are needed, and I (fzyzcjy) will handle the rest of CI / moving the tests currently (will have more automation in the future).

- [ ] Implement the feature / fix the bug.
- [ ] Add tests in `frb_example/dart_minimal`.
- [ ] Make `dart_minimal`'s CI green.

Utility commands
- Run codegen: `cargo run --manifest-path /path/to/flutter_rust_bridge/frb_codegen/Cargo.toml -- generate`
- Run tests: `./frb_internal test-dart-native --package frb_example/dart_minimal`

PR #2766, opened 2025-06-20
re-applying 'logging_overhaul_redo' changes to master

patmuk:
CI is broken for the PR #2735. It is easier to (again) re-apply the intended changes here.


patmuk:
@fzyzcjy CI is happy :) I rewrote the git history as well so that it is more comprehensive.

Hope you have time for a review :) As you know, I recommend to start with the documentation. I made a new sample project as well (frb_examples/dart_logging) and reverted dart_minimal to how it is in master.

I am available for any comments or discussions :)

fzyzcjy:
Great job! I will review once I have time!

patmuk:
> Great job! I will review once I have time!

Thanks!
Maybe make your review in small batches. You could one day look at the example, another day at the documentation, another day at the generated code in the example and a last day at add_to_lib.rs and the changed lib.rs files, which should be the only code mixing with existing code. 

As for the “why” I did this PR (its been a long time ;) ): The opinionated logging wasn’t working for me - getting to the logs of the iOS Simulator was too complicated and there was a package conflict in one time. With this PR logging to the console works out of the box, on the rust side one can use the log macros only, and for a more sophisticated logging (e.g. into a file) one can plug any framework one desires, dart and/or rust side :)

fzyzcjy:
I see, that looks reasonable! I may prefer to check everything in one time to get a overall understanding though :)

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+* Please refer to https://fzyzcjy.github.io/flutter_rust_bridge/guides/miscellaneous/whats-new for what's changed in V2.
+* Overhaul of the logging implementation, from an opinionated approach to a configurable one. Please consult the [documentation](https://fzyzcjy.github.io/flutter_rust_bridge/guides/logging) for more information. If you implemented logging yourself you might have to migrate your changes.
```
fzyzcjy, line 4: nit: maybe keep changelog unchanged, I will update them later :)

website/docs/guides/how-to/logging.md
```diff
@@ -1,89 +1,385 @@
...
+If you have not disabled automatically configuring your `lib.rs` in your `flutter_rust_bridge.yaml` (with `add_mod_to_lib: false`) these lines are added for you. 
 
-```rust
-fn setup_the_logger() {
-    #[cfg(target_os = "android")]
-    android_logger::init_once(android_logger::Config::default().with_max_level(LevelFilter::Trace));
+Otherwise, enter
+```
+// this export is needed for logging
+pub use crate::frb_generated::StreamSink as __FrbStreamSinkForLogging;
```
fzyzcjy: hmm this seems a bit suboptimal :/

brainstorm: what about

```
enable_frb_logging!();
```

by default (which uses frb_generated as the place). and if someone uses sth different, do 

```
enable_frb_logging!(frb_generated_mod = whatever_else_name);
```

and handle that in the macro

website/docs/guides/how-to/logging.md
```diff
@@ -1,89 +1,385 @@
...
 
-## Approach 1: Use the default one
+Under the hood (and in a nutshell) our implementation uses the [Rust 'log' crate](https://crates.io/crates/log) and the [Dart 'logging' package](https://pub.dev/packages/logging), which are maintained by the respective language teams.
 
-If using the template by `flutter_rust_bridge_codegen create/integrate`, the "print logs to console" is configured by default,
-via the auto-generated call to `flutter_rust_bridge::setup_default_user_utils()`.
+Log messages are sent via a FRB's Stream implementation from Rust to Dart.
 
-Thus, you do not need to do anything :)
+## Setup
```
fzyzcjy, line 10: nit: is it possible we simplify this process while still providing same flexibility, e.g.

1. add the logging dependency
2. add `enable_frb_logging!();`

and done

by default we setup everything s.t. the users can directly do logging and find the logs appear correctly. and if user wants to customize, just do a `enable_frb_logging!(some_flag = false);`

website/docs/guides/how-to/logging.md
```diff
@@ -1,89 +1,385 @@
...
+
+## How to use it
+
+All logs will be printed to `stdout`.
+
+You can issue log statements in Rust and Dart by following the usual way done in the respective 3rd party packages:
+
+In **Rust** simply call `log::info!()` (or any of the [log crate's log levels](https://docs.rs/log/latest/log/enum.Level.html)).
+
+In **Dart** you first need to initialise logging with a call to `final LOGGER = FRBLogger.initLogger()`.
```
fzyzcjy: wondering whether it is possible to avoid this to make it more fluent to use. for example, if you want some code to be called on init, you may want `#[frb(init)]`

website/docs/guides/how-to/logging.md
```diff
@@ -1,89 +1,385 @@
...
+
+All of Dart's logging package log levels are available, but the more exotic ones have been replaced with the more common levels used in Rust:
+- FINE -> trace
+- CONFIG -> debug
+- SEVERE -> error
+- SHOUT -> fatal
+
+While Rust's log crate captures the module, file and line number of the log statement, Dart's logging package does not. 
+Here it is idiomatic to get a new handle of the logger in each file, and give it a name that tells you where your log statement is originating from.
+Usually one calls `final OTHER_LOGGER = Logger("other_logger");` from the library directly, but again, use `final OTHER_LOGGER = FRBLogger.getLogger("other_logger");` instead.
```
fzyzcjy, line 65: nit: curious what does this do - it seems we only have one "logger that logs from rust"

website/docs/guides/how-to/logging.md
```diff
@@ -1,89 +1,385 @@
...
+// import FRBDartLogger and LogLevel to be used in the customization
+import 'package:flutter_rust_bridge/flutter_rust_bridge_for_generated.dart';
+```
+(...)
+```Dart
+late final FRBDartLogger logger;
+```
+(...)
+```Dart
+  // setup logging
```
fzyzcjy, line 174: hmm wondering whether users just directly setup dart logging

fzyzcjy (commented):
Great job! I firstly reviewed some docs as suggested :)

EDIT: see below for my general naive thoughts; these review comments may be (briefly) viewed after reading general thoughts

fzyzcjy:
My naive thoughts for API:

Main ideas
* As few setup operations as possible
    * => make users easy to setup
    * => and make user code clean
* Still have high flexibility
    * May be done by optional args to disable default things etc

Detailed design

when users want to use it, they just do

1. add `log` as a Rust dependency
2. add `enable_frb_logging!(setup_dart_logging_output = true);` to their code
3. tell users that, now it is printed to stdout, but later users should follow Dart's logging package and e.g. output to a service, and then they should remove this temp `setup_dart_logging_output=true` thing

what users get
* Rust `log::info` auto works by delegating it into Dart logging system, i.e. whenever users do log::info, it is as if they do a Dart log.info
* When users write down Rust log::info (and/or Dart log.info), things are printed to stdout
    * Implemented by `Logger.root.onRecord.listen((record) { print('${record.level.name}: ${record.time}: ${record.message}'); });` in the default impl code.
    * But should allow users to setup their own listeners (since printing is usually not the correct way, but just the adhoc way to let users see something). May be done by a bool flag, say, `enable_frb_logging!();` without `setup_dart_logging_output = true` will not trigger this

The current design is great! Just some nits and we may be possible to further improve it :)

fzyzcjy:
btw it seems we need to update the windows ci, could you please help bumping them (in a separate PR)?

image

patmuk:
> Great job! I firstly reviewed some docs as suggested :)
> 
> EDIT: see below for my general naive thoughts; these review comments may be (briefly) viewed after reading general thoughts

Wow, great, many thanks for this timely review! I am responding now ...

website/docs/guides/how-to/logging.md
patmuk: I agree .... that was quite hard to implement.

I tried it that way, but the problem is that the location of `frb_generated.rs` depends on the value of `rust_output`. And at the time of macro expansion I cannot access this value.

Thus a default value only works if the user did not change `rust_output` in his `frb.yaml` settings. I thought it would be too confusing and would look unnecessarily redundant for the user that `rust_output` needs to be configured for logging a second time.

Injecting the location into `lib.rs` is more automatic.

Nevertheless, if you think otherwise, we can change it to require the macro value. I implemented it that way first, it works, but seemed to user-unfriendly to me.

website/docs/guides/how-to/logging.md
patmuk: I rewrote it, please check if it is better understandable now :)

website/docs/guides/how-to/logging.md
patmuk: We need the `initLogger()` call for dart-side configuration (see further down for more explanation).

However, we could execute it automatically, if not done by the user:
As usual with logging frameworks one has to call `getLogger`. At the moment I am throwing an Exception, if `initLogger()` has not been called before:
```
    if (_singleton == null) {
      throw Exception("You have to call FRBLogger.initLogger() first!");
    }
```
However, instead we could just call `initLogger()` here and create the singleton. 
I did it this way so that the user notices if he accidentally refactored `initLogger` away and thus the options are reset to default, e.g. log level info.

But, as always, you are defining the ergonomics - so if you think `initLogger` should be called only if customization is done, I can easily change that.
----
Here is a bit longer explanation of initLogger:

Probably we can rewrite it ... but the `initLogger()` function is used to customize the logging as well:

If one doesn't want to do it via Rust (`enable_frb_logging!()`) the same parameters can be given to this call on the dart side. If we do it via init we either drop the option for configuration via dart or need a function for that anyways.

If we don't support configuration with dart code using dart logging frameworks gets more complicated (having a rust function, which is a dart callback, set up in the macro).

Besides this, I created the `initLogger()` function via 
```
    #[flutter_rust_bridge::frb(dart_code = "
    import 'package:logging/logging.dart';

    static FRBDartLogger initLogger(
```
so using `#[frb(init)]` might be more complicated ...

... thus, better leave it this way?

website/docs/guides/how-to/logging.md
patmuk: I don't fully understand: Do you mean why users can't just setup logging in Dart directly, not using FRBLogging - and/or that his is even more effort?

Well, they still can just setup Logging in Dart instead of using this implementation. But they would get Dart logs only this way, Rust's logs would only show up in `stdout`. So the extra effort is needed to get logs from both sides.

CHANGELOG.md
patmuk: Ok! If you don't mind I will leave this for now for you to rewrite, but remember in the future to not edit it. (Mainly did that for triggering CI.)

patmuk:
> The current design is great! Just some nits and we may be possible to further improve it :)

Thanks :) I think it is closer to your thoughts, I just did not describe it that well (did some refactoring to make things more automatic and did not keep up well with the documentation. Now I rewrote that a bit :) ).

> Main ideas
> 
> * As few setup operations as possible

Totally agree.

>   * => and make user code clean

Yes. I tried to minimize the `frb_generated.rs` code with putting code into `frb_dart/lib/src/utils/frb_logger.dart`, but could not do so for the remaining generated code. Appreciated if you can identify more code from `frb_generated.rs` which can be moved to `frb_logger.dart` to make it even cleaner :)

>   * May be done by optional args to disable default things etc
Yes. These go to `enable_frb_logging!()` or `initLogger()`. Options are mirrored, so the user can choose if he likes to set them in Rust or Dart.

> when users want to use it, they just do
> 
> 1. add `log` as a Rust dependency
> 2. add `enable_frb_logging!(setup_dart_logging_output = true);` to their code
> 3. tell users that, now it is printed to stdout, but later users should follow Dart's logging package and e.g. output to a service, and then they should remove this temp `setup_dart_logging_output=true` thing

Yes, it already works like this :) However, without the boolean flag. Either the user provides nothing or he overwrites a default setting directly.

> what users get
> 
> * Rust `log::info` auto works by delegating it into Dart logging system, i.e. whenever users do log::info, it is as if they do a Dart log.info
> * When users write down Rust log::info (and/or Dart log.info), things are printed to stdout
>   
>   * Implemented by `Logger.root.onRecord.listen((record) { print('${record.level.name}: ${record.time}: ${record.message}'); });` in the default impl code.
>   * But should allow users to setup their own listeners (since printing is usually not the correct way, but just the adhoc way to let users see something). May be done by a bool flag, say, `enable_frb_logging!();` without `setup_dart_logging_output = true` will not trigger this

Yes, it works like this. However, instead of `setup_dart_logging_output = true`, setting `customLogFunction =` is disabling stdout writing and uses whatever the user provides.


I probably misunderstood some things ... and I did not describe it good enough before.

patmuk:
> btw it seems we need to update the windows ci, could you please help bumping them (in a separate PR)?
> 
> image

Done (as you know ;) )

fzyzcjy:
> and make user code clean
Yes. I tried to minimize the frb_generated.rs code

btw no big worries about frb_generated.rs - would be great to keep it minimal when possible, but that is generated code and users will not read them, so it is ok to be somehow messy.

TODO: reply more

> my child got sick today :(

Sad to hear that :( Take your time, and wish your child get well soon!

website/docs/guides/how-to/logging.md
fzyzcjy: My thought is that,

1. if users use default rust_output, then tell them to write down `enable_frb_logging!();`
2. if users use custom rust_output, then tell them (in doc) to write down `enable_frb_logging!(frb_generated_mod = whatever_else_name);`

> I thought it would be too confusing and would look unnecessarily redundant for the user that rust_output needs to be configured for logging a second time.
Injecting the location into lib.rs is more automatic.

Oh I see...
I personally feel most users will never change the rust_output name, so it is fine to configure twice for the few users who change.

Btw another way may be:

```
# lib.rs
pub use crate::another_rust_output_name as frb_generated;
```

Then we just use frb_generate forever

website/docs/guides/how-to/logging.md
fzyzcjy: > Per default any log output is written to stdout, via println! in Rust.

Hmm confused, I thought it may be

* by default, (1) rust log is streamed to dart's logging package (2) dart's logging package is configured s.t. it `print` the msg

fzyzcjy:
Hmm maybe I will find another piece of time to give a more detailed draft (pseudocode), no worries

website/docs/guides/how-to/logging.md
patmuk: [EDITED}
Inspired by your last comment I found a way to automatically setup lib.rs :)
A user only has to do something, if he disabled `add_to_lib` AND set `rust_output` :)

I updated the documentation accordingly.

Maybe keep this for last, but review `add_to_lib.rs`. I did quite some improvements there:
- marking the injected code section, so that
- it is easier in future to rewrite the injected code
- migrating the former injected code to the new lines
- handling newlines correctly (following rust_fmt)

- injecting `use crate::...:::frb_generated` if a user sets `rust_output`
- setting `mod` to the top-level module of that user provided path.
... I did not yet updated the documentation for that - will wait for your approval, especially for the last point.
I think it is helpful for the user to have `mod` set as well, even if he has to setup the rest (sub-mod file structure) himself.

website/docs/guides/how-to/logging.md
patmuk: [EDITED]

Now in a new project only 
1. add `enable_frb_logging!()` 
is needed :)

Depending on the setup more steps might be needed - but I believe it is reduced to the necessary minimum :) See the updated documentation.

website/docs/guides/how-to/logging.md
patmuk: [EDITED]

This is how it is implemented:

Rust:
1. Rust's log-macro creates a log record, which includes the message but other information as well (e.g. log level, code line, ...)
2. I am using cronos to add a timestamp
3. now the log record is translated into a neutral format (MirLogRecord in frb_rust/src/misc/logging.rs:238).
4. the record is handled by the function `log_fn` (generated by the macro expansion), which uses the user's provided log output function or the default (which does println!()

Dart:
1. Dart's log-functions (e.g. LOGGER.info()) create a (Dart) log record 
2. (no need to add a timestamp, as this is already in the log record)
3. the log record is translated to the neutral format as well (`MirLogRecord`)
4. the record is handled by the function `logFn`, which is the FRB generated call to `log_fn`.

In both cases:
- `log_fn` either calls the rust-written default function to output log records (via println!())
- or, if done, the user provided function.

If the user's function was written:
- in Rust, configured as a parameter of `enable_frb_logging!()`, it is called by `log_fn` directly
- in Dart, configured as a parameter of `initLogger();`, it is called as a Dart-Callback by `log_fn`

patmuk:
> Hmm maybe I will find another piece of time to give a more detailed draft (pseudocode), no worries

Now the setup is even simpler:
1. put `enable_frb_logging!()` in a rust_input file

That's it - more steps are only needed if certain non-default settings have been done.

For customization:
Either in Rust: Set parameters of `enable_frb_logging!()`.
Or in Dart: Set parameters of `.initLogger()`.

website/docs/guides/how-to/logging.md
patmuk: [EDITED]

I changed it, so that `initLogger()` is needed only if customization is desired. Otherwise `getLogger()` is enough.

fzyzcjy:
Looks great! I see your long replies and will find time to sit down and think carefully :)

patmuk:
> Looks great! I see your long replies and will find time to sit down and think carefully :)

I will make it easier and edit my replies. I implemented some suggested ergonomic changes, so I can shorten the text :)

patmuk:
> My naive thoughts for API:
> 
> Main ideas
> 
> * As few setup operations as possible
>   
>   * => make users easy to setup
>   * => and make user code clean
> * Still have high flexibility
>   
>   * May be done by optional args to disable default things etc
> 
> Detailed design
> 
> when users want to use it, they just do
> 
> 1. add `log` as a Rust dependency
> 2. add `enable_frb_logging!(setup_dart_logging_output = true);` to their code
> 3. tell users that, now it is printed to stdout, but later users should follow Dart's logging package and e.g. output to a service, and then they should remove this temp `setup_dart_logging_output=true` thing
> 
> what users get
> 
> * Rust `log::info` auto works by delegating it into Dart logging system, i.e. whenever users do log::info, it is as if they do a Dart log.info
> * When users write down Rust log::info (and/or Dart log.info), things are printed to stdout
>   
>   * Implemented by `Logger.root.onRecord.listen((record) { print('${record.level.name}: ${record.time}: ${record.message}'); });` in the default impl code.
>   * But should allow users to setup their own listeners (since printing is usually not the correct way, but just the adhoc way to let users see something). May be done by a bool flag, say, `enable_frb_logging!();` without `setup_dart_logging_output = true` will not trigger this
> 
> The current design is great! Just some nits and we may be possible to further improve it :)

Thanks :)
I reworked it and believe that the setup steps are reduced to a minimum now :)

patmuk:
@fzyzcjy CI is fine now, all up-to-date :) I don't know why codacy is complaining though. Probably I did not integrate the new example dart_logging correctly?

fzyzcjy:
No worries about codacy!

fzyzcjy:
I will review the doc again now and express my thoughts (probably quite similar to yours, but just provide a full text to make it have context; and I will copy-and-elaborate on my old text).

### Main ideas

* As few setup operations as possible
    * => make users easy to setup
    * => and make user code clean
* Still have high flexibility
    * May be done by optional args to disable default things etc

### Detailed design

when users want to use it, they just do

1. add `log` as a Rust dependency (✅ I see it in the doc)
2. add `enable_frb_logging!(setup_dart_logging_output = true);` to their code (✅)
3. tell users that, now it is printed to the console, but later users should follow Dart's logging package and e.g. output to a service, and then they should remove this temp `setup_dart_logging_output=true` thing 
    * 🚧 it would be great for the default to be "users should see texts on their `flutter run` just like they `print` in dart"
        * but the doc currently says "println in rust and flutter *hides* it". then users will get confused "why my text is not anywhere" :(

### Sample implementation

note: I have not checked your code and below is just what is in my mind. it is just FYI instead of any requirements!

frb_rust

```rust
macro_rules! configure_logging {
  (setup_dart_logging_output, ...) => {
    #[frb(init_dart_code="""initDefaultLogging();""")]
    #[frb(dart_code="""
      void initDefaultLogging() {
        frbInitLogger().listen((msg) => logging.info(msg)); // listen to rust stream, forward to dart logging system
        if (setup_dart_logging_output) Logger.root.onRecord.listen((record) => print('${record.level.name}: ${record.time}: ${record.message}'));
      }
    """)]
    pub fn frb_init_logger(sink: StreamSink<SomeLogRecordStruct>) {
      ... do what is mentioned in https://cjycode.com/flutter_rust_bridge/guides/how-to/logging , https://github.com/MnlPhlp/flutter_logger/blob/main/src/lib.rs , etc ...
    }
  }
}
```

user app rust

```rust
configure_logging!();
```

frb dart

```dart
abstract class BaseEntrypoint {
  initImpl() {
    ... await executeRustInitializers(); // old code
    // ADD
    await executeDartInitializers();
  }
}
```

user app dart

```dart
// genereated code
class RustLib {
  executeDartInitializers() {
    // this is auto generated from the `#[frb(init_dart_code)]`
    initDefaultLogging();
  }
}

// this is auto generated from the `#[frb(dart_code)]`
void initDefaultLogging() {...}
```

frb_codegen

* Support `#[frb(dart_code)]` on a `fn`, which means copy-paste those dart code into frb_generate.dart
* Support `#[frb(init_dart_code)]`, which means execute that dart expression at app start time.
* note that these annotations are unrelated to functions. e.g. `#[frb(init_dart_code="g();")] fn f() {}` is possible, where g is something unrelated to f.

pros
* allow other future cases when any user want to expose a minimal rust-only operation which contains dart code
cons
* the annotations are unrelated to the functions indeed, which may make readers of this annotation a bit confused

another potential implementation is to let users do one manual initialization on the dart side. I personally slightly favor the "no manual call on dart" to make it a bit shorter to integrate.

patmuk:
Thanks for the new review!

> 1. add `log` as a Rust dependency (✅ I see it in the doc)
> 2. add `enable_frb_logging!(setup_dart_logging_output = true);` to their code (✅)

Yes, that's it :) Logging dependency needed only for old projects, as it is in `cargo-template` for new ones :)
Though I don't understand the benefit of `setup_dart_logging_output = true` - at the moment the user either adds `enable_frb_logging!();` to enable logging with all defaults or he adds customization as parameters to this macro.

>    * 🚧 it would be great for the default to be "users should see texts on their `flutter run` just like they `print` in dart"      
>      * but the doc currently says "println in rust and flutter _hides_ it". then users will get confused "why my text is not 
anywhere" :(

Maybe my wording in the docs ("Note that this is needed, if you are developing a _Flutter_ application, as `stdout` is not shown.") is not good ...

Default logging works for `cargo run` and `dart run`. However, in my Simulator on my Mac, when running a flutter application, it is not showing stdout in the console, thus no logging output.

I assume that this is highly dependent on which environment is used - and maybe stdout is indeed shown in console.app or somewhere else. I did not searched for that.

So, basically, what I mean to say is "unless you are executing a cli app (rust or dart) you should use a logging framework that supports the environment where you run your app. Look at the example in this doc for how to do that in the Flutter Simulator.

But that was a bit too verbose ;) Shall I reformulate that in the docs? I don't think we should do something more complicated than println in the default version.

> ### Sample implementation
> note: I have not checked your code and below is just what is in my mind. it is just FYI instead of any requirements!

sure, appreciated!

> frb_rust
> ```rust
> macro_rules! configure_logging {
>     #[frb(init_dart_code="""initDefaultLogging();""")]
>     #[frb(dart_code="""
> ```

Yes, this is basically what I did, see `frb_rust/src/misc/logging.rs`. 

In Dart usually one gets an instance of a logger singleton per class - and calling this function the first time does the initialization. Thus automatically for the user :)

> user app rust
> 
> ```rust
> configure_logging!();
> ```

... now called `enable_frb_logging!();` - but we could rename it. 

> frb dart
and
> user app dart

Not needed. Just

```dart
final LOGGER = FRBLogger.getLogger();
```
as one does with any dart logging framework.

or, if the user want's to customize in dart and not rust:

```dart
final LOGGER = FRBLogger.initLogger(maxLogLevel: LogLevel.trace);
```
instead.  

> frb_codegen
> 
> * Support `#[frb(dart_code)]` on a `fn`, which means copy-paste those dart code into frb_generate.dart
> * Support `#[frb(init_dart_code)]`, which means execute that dart expression at app start time.
> * note that these annotations are unrelated to functions. e.g. `#[frb(init_dart_code="g();")] fn f() {}` is possible, where g is something unrelated to f.

Would be nice :) At the moment I am using #[frb(dart_code)]` on the struct `FRBLogger`, which gives me the function `initLogger()`, see frb_rust/src/misc/logging.rs:65). This would be `#[frb(init_dart_code)]` - but at the moment I am calling this function if the singleton is not initialized yet.

So - nice to haves, but not necessary.

> 
> pros
> 
> * allow other future cases when any user want to expose a minimal rust-only operation which contains dart code
>   cons
> * the annotations are unrelated to the functions indeed, which may make readers of this annotation a bit confused

Then IMHO it is better to attach the dart-fn to a struct, like I did. Thus the translated dart class contains this function. It is not callable from rust - but that is ok - most probably it is need in dart only.

> 
> another potential implementation is to let users do one manual initialization on the dart side. I personally slightly favor the "no manual call on dart" to make it a bit shorter to integrate.

Do you mean the general case or this specific? Here it works, as a user is used to call `getLogger()` in Dart, where I sneak in the initialization. So a user would not notice he needs to do something additional :)

fzyzcjy:
Thanks for the reply! I will reply as soon as having time

patmuk:
> Thanks for the reply! I will reply as soon as having time

Appreciated :) I will try to be more brief next time ;)

patmuk:
@fzyzcjy just pinging you so we can finally merge this PR.
Please ping me once you post a question or suggestion - I might overlook it in the above threads :/

fzyzcjy:
sure, this is on my list and I will review once having time! but sadly I do not have time to review this recently :( you can see how much work I need to handle in https://github.com/sgl-project/sglang for example

patmuk:
> sure, this is on my list and I will review once having time! but sadly I do not have time to review this recently :( you can see how much work I need to handle in https://github.com/sgl-project/sglang for example

Yes, absolutely! Totally understand that!

fzyzcjy:
bump (i sent an email months ago)

rubiktubik:
@patmuk @fzyzcjy 

What is currently blocking this feature? It would make the library even better than it already is. :-)

fzyzcjy:
The main blocker is that the PR adds a nontrivial great feature, thus needs some careful review, but I have been having tight ddls and are not able to review this kind of big PRs :( But feel free to directly use the fork version!

fzyzcjy:
(no worries about review results, I may handle it myself)

patmuk:
> /gemini review

Thanks, good idea :) I am now commenting on the findings.

frb_dart/lib/src/utils/frb_logger.dart
```diff
@@ -0,0 +1,307 @@
...
+      case 'WARN':
+        return LogLevel.warn;
+      case 'ERROR':
+        return LogLevel.error;
+      case 'OFF':
+        return LogLevel.off;
+      default:
+        print(
+            'Unknown LOG_LEVEL: "$levelStr". For potential values, refer to LogLevel enum definition.');
+        exit(1); // Exit, as this is a critical configuration error
```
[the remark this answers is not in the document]
patmuk: My intention here was to abort so the user sees the error as early as possible - and not silently overlook and wonder why logging doesn't work. However, we should do it in the philosophy like the rest is implemented. If there is a critical misconfiguration in any of FRB's settings, what is usually done?

frb_codegen/src/library/codegen/polisher/add_mod_to_lib.rs
```diff
@@ -24,35 +23,604 @@ pub(super) fn try_add_mod_to_lib(rust_crate_dir: &Path, rust_output_path: &Path)
...
-        let comments = " /* AUTO INJECTED BY flutter_rust_bridge. This line may not be accurate, and you can change it according to your needs. */";
-        let modified_content_lib_rs = format!("{expect_code}{comments}\n{raw_content_lib_rs}");
+    let final_content = process_lib_rs_content(
+        &lib_rs_content_normalized,
+        &rust_output_path_relative_to_src_folder,
+    )?;
 
-        fs::write(&path_lib_rs, modified_content_lib_rs).unwrap();
+    // --- Write back only if content actually changed from the original normalized state, ignoring white spaces ---
+    if final_content.replace("/n", "") != lib_rs_content_normalized.replace("/n", "") {
```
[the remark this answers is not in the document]
patmuk: I think this is accurate and a good finding! My intention was to not touch lib.rs if only non-visible characters changed.

I probably did not explain what I do here in general (in add_mod_to_lib.rs): In main only the generated mod is added, one line. But in this branch we need to a dependency as well, so we need to add multiple lines. However, this cannot be after or before certain other elements, needs to sometimes be in the middle.

Thus I needed to rewrite add_mod_to_lib.rs and introduced markers for the generated part, so that hand written lines are not accidentally overwritten.

frb_example/dart_logging/rust/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+// AUTO INJECTED BY flutter_rust_bridge.
+// The following lines may not be accurate; change them according to your needs.
+mod bridge;
+use crate::bridge::frb_generated;
```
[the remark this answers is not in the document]
patmuk: I forgot - but I think the re-export (pub) is not necessary. I roughly remember the I got compilation errors when it was present ...

website/docs/guides/how-to/logging.md
```diff
@@ -1,89 +1,379 @@
...
+    pub level_number: u16,   // The log level encoded. Decode with `FRBLogger.logLevelFromNumber(x)` in Dart or `from_u16(x)` in Rust. : `Rust::log::Record::Level`, `Dart::Logger::LogRecord::Level`
+    pub message: String, // The String given to the log statement: `Rust::log::Record::args`, `Dart::Logger::LogRecord::message`
+    pub logger_name: String, // The name of the logger given by `FRBLogger.initLogger(name: "MyClass");`, `Rust::log::Record::target`, `Dart::Logger::LogRecord::loggerName`
+    // pub time: String, // log::Record::?, Dart::Logger::LogRecord::time --> omitted, as there is no time record in the log crate's Record
+    pub rust_log: bool, // true, if the log statement originates from Rust code
+    pub module_path: Option<String>, // `Rust::log::Record::module_path`, None for Dart
+    pub file_name: Option<String>, // `Rust::log::Record::file_name`, None for Dart
+    pub line_number: Option<u32>, // `Rust::log::Record::line_number`, None for Dart
 }
 ```
```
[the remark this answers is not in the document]
patmuk: 1. Let's double check in the code. I think I copy & pasted, so if the log levels are not matching, this is probably wrong in the implementation as well. But somehow I don't believe it is wrong ...
2. Yes, good catch!
3. maybe it is `time` in rust and `timestamp` in dart? Let's check the code here as well. I wonder why these lines are commented - maybe I just copy & pasted and this needs to be removed.

fzyzcjy:
No worries! Indeed w/ AI agents, I am thinking about maybe have a try to see whether can directly handle all the PRs, i.e. let agent help me modify where I think is improveable, and then merge your PR (and others' PR) directly.

fzyzcjy:
I am working on it, and will firstly merge this pr to main, and then revert the commit, and then implement the code in a new pr, s.t. agents do not get confused about historical code, and at the meantime your work is fully recognized to main :)

PR #2790, opened 2025-07-04
Add `disable_dart_fix`, `disable_dart_format` and `no_enable_overlay_lib` to `flutter_rust_bridge_codegen integrate`.

### LINKED ISSUE #2789: Add `no-overwrite` or related options to `flutter_rust_bridge_codegen integrate`.
I'm using `flutter_rust_bridge_codegen integrate` to generate `rust_builder` and other files in CI/CD. But it always break my code: it overwrites `lib/main.dart` and applies dart fixes.

Add a `no-overwrite` or `no-template` option to disable the overwrite. And a `no-fix` to disable dart fix and rust format.

I run `git reset --hard` after the command to recover. But this command takes a few minutes to run.

**Additional context**

xuxiaocheng0201:
Fixes #2789

The new `flutter_rust_bridge_codegen integrate --help` is:


xuxiaocheng0201:
Another question: Are the names of these options appropriate?

fzyzcjy:
> Another question: Are the names of these options appropriate?

Yes I am also thinking about this - would be great if we do not have both "no enable" and "disable". Maybe follow existing params of frb generate command

frb_codegen/src/library/integration/integrator.rs
```diff
@@ -76,11 +104,19 @@ pub fn integrate(config: IntegrateConfig) -> Result<()> {
...
+        info!("Dart fixes is disabled.")
+    }
 
-    info!("Format Dart code");
-    dart_format(&dart_root, 80)?;
+    if config.enable_dart_format {
+        info!("Format Dart code");
+        dart_format(&dart_root, 80)?;
+    } else {
+        info!("Dart formats is disabled.");
```
fzyzcjy:
```suggestion
        info!("Dart format is disabled.");
```

frb_codegen/src/library/integration/integrator.rs
```diff
@@ -76,11 +104,19 @@ pub fn integrate(config: IntegrateConfig) -> Result<()> {
     info!("Setup cargokit dependencies");
     setup_cargokit_dependencies(&dart_root, &config.template)?;
 
-    info!("Apply Dart fixes");
-    dart_fix(&dart_root)?;
+    if config.enable_dart_fix {
+        info!("Apply Dart fixes");
+        dart_fix(&dart_root)?;
+    } else {
+        info!("Dart fixes is disabled.")
```
fzyzcjy:
```suggestion
        info!("Dart fix is disabled.")
```

frb_codegen/src/library/integration/integrator.rs
```diff
@@ -36,27 +39,52 @@ pub fn integrate(config: IntegrateConfig) -> Result<()> {
...
+            Template::Plugin => execute_overlay_dir(
+                &TemplateDirs::PLUGIN_LIB,
+                &replacements,
+                &dart_root,
+                &config,
+                Some(&[format!("{dart_package_name}.dart")]),
+            )?,
+        }
+    }
+    if config.enable_integration_test && config.template == Template::App {
```
fzyzcjy: hmm, seems the original code already handles integration test folder via filtering. Wondering whether we can mimic that to handle  the "overlay lib" and minimize code changes

frb_codegen/src/binary/commands.rs
```diff
@@ -221,10 +221,22 @@ pub(crate) struct CreateCommandArgs {
 
 #[derive(Debug, Args)]
 pub(crate) struct IntegrateCommandArgs {
-    /// Generate code related to integration test
+    /// Do NOT generate code related to lib/example etc.
+    #[arg(long)]
+    pub(crate) no_enable_overlay_lib: bool,
```
fzyzcjy: nit: would be great to change the name a bit, since "overlay lib" may not be super clear

fzyzcjy (commented):
Great job! Only some nits

frb_codegen/src/library/integration/integrator.rs
xuxiaocheng0201: I think that since the template file is placed in another folder, the file content should be decoupled from the code section. But using filters is not unacceptable. Which one do you think would be better?

fzyzcjy: I personally slightly favor minimizing folder complexity, i.e. the filtering approach :)

xuxiaocheng0201: This could be a bug: `flutter_rust_bridge_codegen create tester` runs well, but `flutter_rust_bridge_codegen create integration_test` will fail.

xuxiaocheng0201: This could be a bug. Due to the fact that the `path` in `filter_file` is a full path that starts with `dart_root`, the filter may return wrong values.

xuxiaocheng0201: Ok, I fixed it.

frb_codegen/src/binary/commands.rs
```diff
@@ -221,9 +221,21 @@ pub(crate) struct CreateCommandArgs {
...
+    #[arg(long)]
+    pub(crate) no_integration_test: bool,
+
+    /// Do NOT apply dart fix after generating code.
+    #[arg(long)]
+    pub(crate) no_dart_fix: bool,
+
+    /// Do NOT format dart code after generating code.
+    #[arg(long)]
+    pub(crate) no_dart_format: bool,
```
fzyzcjy:
```suggestion
    /// Do not generate code related to lib/example etc.
   #[arg(long)]
    pub(crate) no_write_lib: bool,

    /// Do not generate code related to integration test
    #[arg(long)]
    pub(crate) no_integration_test: bool,

    /// Do not apply dart fix after generating code.
    #[arg(long)]
    pub(crate) no_dart_fix: bool,

    /// Do not format dart code after generating code.
    #[arg(long)]
    pub(crate) no_dart_format: bool,
```

frb_codegen/src/library/integration/utils.rs
```diff
@@ -17,12 +17,12 @@ pub(super) fn overlay_dir(
     filter: &impl Fn(&Path) -> bool,
 ) -> Result<()> {
     for entry in current_reference_dir.entries() {
-        let target_sub_path = base_target_path.join(entry.path());
-
-        if !filter(&target_sub_path) {
+        if !filter(entry.path()) {
```
fzyzcjy, line 20: hmm, curious why we need to change the logic here?

fzyzcjy (approved):
LGTM! only a nit and ready to merge

frb_codegen/src/library/integration/utils.rs
xuxiaocheng0201: I commented above. https://github.com/fzyzcjy/flutter_rust_bridge/pull/2790#discussion_r2188373921

xuxiaocheng0201: If the dart_root contains 'rust_builder' or 'integration_test' etc. the filter may work unexpectedly. Although this is unusual, it truly is a bug.

fzyzcjy: oh I see, reasonable!

fzyzcjy:
@all-contributors please add @xuxiaocheng0201 for code

fzyzcjy:
@all-contributors please add xuxiaocheng0201 for code

PR #2820, opened 2025-08-01
fix text input bug in example code

Lightfire228:


fzyzcjy:
Hmm my naive guess seems the "=" is correct since it is full text. Could you please have a test and check whether "+=" or "=" owkrs

Lightfire228:
What's happening, is the text provided by `onChanged: (text) => {}` is only the incoming text, not all of it.
So, just assigning `state.inputText = text` overwrites what was previously typed

Here's an example video 


https://github.com/user-attachments/assets/400a260f-51d5-4706-93da-d1c057f30d4a



[The code is here](https://github.com/Lightfire228/Timekeeper/blob/15ab95aedec189cde6821c09da6495625f37dd3b) , but it's mostly the todo list example, copy+pasta into [`state.rs`](https://github.com/Lightfire228/Timekeeper/blob/15ab95aedec189cde6821c09da6495625f37dd3b/rust/src/frb_api/state.rs) and [`main.dart`](flutter/lib/main.dart)

fzyzcjy:
Interesting... But in https://api.flutter.dev/flutter/material/TextField/onChanged.html it is mentioned onChanged does provide full text instead of the new char

Lightfire228:
this is strange
clicking away and back causes the next keystroke to return the entire text buffer, not just the new text

Maybe a bug in my version of flutter...?

----

ok, checking the android build, i think it's a bug in the linux build
the android build behaves as per the docs

Lightfire228:
Ok, I created a [test example](https://gist.github.com/Lightfire228/6bd86e51d1c5df3d42ced0167834fcea) in pure dart, and the bug isn't happening

I'll need to look deeper

Should i create an issue?

PR #2826, opened 2025-08-12
Add GitHub Actions workflow for precompiling binaries and documentation

iota9star:
This pull request introduces support and documentation for using precompiled Rust binaries in the integration template for plugins, aiming to reduce developer friction by allowing builds without a local Rust toolchain. The changes provide a workflow for generating, verifying, and distributing precompiled binaries via GitHub Actions and releases.

Support for precompiled binaries:

* Added a sample `cargokit.yaml` configuration file in the Rust crate directory, specifying the URL prefix for binaries and a placeholder for the public key used for verification.

Documentation improvements:

* Added a comprehensive `precompiled_binaries.md` guide explaining the process of using precompiled binaries with Cargokit, including configuration, key management, and verification steps.
* Included instructions and a sample GitHub Actions workflow for building and uploading precompiled binaries for multiple platforms (Linux, macOS, Windows, and Android).

Automation via CI:

* Added a reusable GitHub Actions workflow (`precompile-binaries.yml`) to automate the building and uploading of precompiled binaries for all supported platforms, including setup steps for Dart, Rust, Java, Android SDK, and NDK.

### DIFF
--- a/frb_codegen/assets/integration_template/plugin/.github/workflows/precompile-binaries.yml
+++ b/frb_codegen/assets/integration_template/plugin/.github/workflows/precompile-binaries.yml
@@ -1,0 +1,93 @@
+name: Precompile Binaries
+
+on:
+  workflow_dispatch:
+    inputs:
+      TAG:
+        description: 'Set a Tag'
+        required: true
+        default: ''
+  push:
+    tags:
+      - v*
+
+env:
+  GITHUB_REPOSITORY: REPLACE_ME_GITHUB_USERNAME/REPLACE_ME_DART_PACKAGE_NAME
+  RUST_CRATE_DIR: REPLACE_ME_RUST_CRATE_DIR
+  NDK_VERSION: 27.0.12077973
+  MIN_SDK_VERSION: 21
+
+jobs:
+  precompile:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Dart
+        uses: dart-lang/setup-dart@v1
+
+      - name: Setup Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install Linux dependencies
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang cmake ninja-build pkg-config libgtk-3-dev liblzma-dev
+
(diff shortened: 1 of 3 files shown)


iota9star:
@fzyzcjy 
image
How should I test the modifications? The test seems to be stuck

fzyzcjy:
Great job! I will check it ASAP (hopefully today)

fzyzcjy:
Great job! But it would be great if we put it as a pure doc (and optionally some folders and files) instead of modifying the template, since by default users may want to use the non-precompile approach to avoid binding a github repo etc.

Therefore, what about 

* Put the thorough doc in the doc folder
* For the sample files, one way may be put it in a folder like frb_examples/precompiled_binary, with only these files exist

iota9star:
I have deleted the template file and only modified the document. However, I think it would be better to keep the template file, because I have found that all libraries (at least those I have seen) do not use precompiled binary files, and various issues are mentioned. I guess they may not be aware of the precompiled binary files.

**Deleting is always easier than creating**

fzyzcjy:
Good point! My main worry is that, if we make it the default in generated new projects, that cargokit.yaml may make cargokit find that github repo and use precompile mode (correct me if I am wrong), then users get confused and wrongly think they have to configure all these things. Another minor issue is that it is hard to test these code.

I have another brainstorm: what about we still only put doc to frb, but point out a live example - your https://github.com/fluttercandies/fjs - in the doc, and explain to users how to understand the related parts in fjs to use in their projects.

Yet another brainstorm is, we may put these files inside the code gen template, but make it be on a flag that is by default disabled. Then we will not have the issue mentined above.

iota9star:
The logic for cargokit to obtain precompiled binaries is straightforward. If unable to obtain them, it will automatically proceed with local compilation. Of course, we can first attempt to refine the current documentation, as with the latest submitted modifications, which merely involve improving the documentation

https://github.com/irondash/cargokit/blob/ec49f0ee89744a8e80ee8df764bfb1faea95319b/build_tool/lib/src/artifacts_provider.dart#L96
https://github.com/irondash/cargokit/blob/ec49f0ee89744a8e80ee8df764bfb1faea95319b/build_tool/lib/src/artifacts_provider.dart#L181

fzyzcjy:
> The logic for cargokit to obtain precompiled binaries is straightforward. If unable to obtain them, it will automatically proceed with local compilation.

I see, looks reasonable. But it would be great if we do not do the fallback by default, since users may still get confused about the behavior when they do not need a github repo case.

fzyzcjy:
Hmm I re-checked the doc and it seems to be copied from cargokit's doc and make some changes. Therefore, maybe we can remove the duplicated doc (and let users to view the one on cargokit), and only write the different parts in frb.

PR #2859, opened 2025-09-05
Support having a different crate name than the flutter plugin

### LINKED ISSUE #2853: Changing Cargo.toml name causes failed to load dynamic library on iOS and MacOS
### Describe the bug

This issue is similar to https://github.com/fzyzcjy/flutter_rust_bridge/issues/2747 so I apologize for brining it back up. I have a rust library with the library name foo which I want to create a flutter plugin for. Since I wanted to use the same name foo for the flutter plugin (similar to the prior mentioned issue) I used the same name foo for the flutter plugin. The issue is that flutter_rust_bridge will then run into an issue when scanning the third-party crate due to two cates with the same name foo - again just like the other mentioned issue. To try and get around this issue I just renamed the rust project to foo_wrapper. Here is the command that I used to do that: `flutter_rust_bridge_codegen create foo --template plugin --rust-crate-name foo_wrapper --rust-crate-dir rust`.

This was working just fine for Windows and Android, however, this causes issues on MacOS and iOS. The podspec file for both platforms had to be updated to search for `libfoo_wrapper.a` instead of `libfoo.a`. This works for things to compile, however, during runtime flutter cannot find the dynamic library:
```
Unhandled Exception: Invalid argument(s): Failed to load dynamic library 'foo_wrapper.framework/foo_wrapper'
```
This is because the podspec has the name `foo` so after cargokit is done, CocoPods uses this name to create the framework target which ends up being `foo.framework`.

During code generation with flutter_rust_bridge, it uses the name of the rust crate for the stem: https://github.com/fzyzcjy/flutter_rust_bridge/blob/845e45a4e032f022eceb005b2859ab37d2f1b9a8/frb_codegen/src/library/codegen/config/internal_config_parser/generator_parser.rs#L142 Manually changing this after generation fixes this issue (obviously since it now points to the right framework).

Perhaps supporting Cargo.toml dependency aliasing or an option to change the stem only for iOS and MacOS would resolve this problem? I would need to spend more time investigating a proper fix but that is what I know so far.

### Steps to reproduce

Run the following:
```bash
flutter_rust_bridge_codegen create foo --template plugin --rust-crate-name foo_wrapper --rust-crate-dir rust
```
Update the crate name to `foo_wrapper` if not already set (it didn't for me on MacOS but did on Windows for some reason).
Try to run iOS or MacOS.
Linker will fail.
Run (make sure you are in the example directory):
```bash
flutter clean
```
Update `macos/foo.podspec` to have:
```podspec
...
    :script => 'sh "$PODS_TARGET_SRCROOT/../cargokit/build_pod.sh" ../rust foo_wrapper',
...
    :output_files => ["${BUILT_PRODUCTS_DIR}/libfoo_wrapper.a"],
...
    'OTHER_LDFLAGS' => '-force_load ${BUILT_PRODUCTS_DIR}/libfoo_wrapper.a',
  }
```
Try to run iOS or MacOS.

### Logs

```shell
N/A
```

### Expected behavior

_No response_

### Generated binding code

```shell

```

### OS

Windows and MacOS

2.11.1

### Flutter info

```shell

```

### Version of `clang++`

_No response_

### Additional context

_No response_

NathanKolbas:
## Changes

Fixes #2853

This PR allows flutter plugins to have a different name than the crate by supporting the `--rust-crate-name` argument when `plugin` is chosen as the `--template`.

The `frb_example/plugin_crate_name` directory was created to show and test that this works as intended. Manually tested on Windows, Android, Linux, MacOS, iOS, and Web.

For this to work on MacOS and iOS, `module_name` is used in the `.podspec` file so that the built output can use the same `stem` as all the other platforms.

## Procedure and Checklist

Utility commands


frb_codegen/assets/integration_template/plugin/ios/REPLACE_ME_DART_PACKAGE_NAME.podspec
```diff
@@ -0,0 +1,46 @@
+#
```
fzyzcjy, line 1: qq: why do we have this new file, will we override user things

frb_example/plugin_crate_name/android/src/main/AndroidManifest.xml
fzyzcjy, line 1: hmm I am not sure whether we should put it in frb_example, since it is a bit heavy to maintain one more example

frb_codegen/assets/integration_template/plugin/android/build.gradle
fzyzcjy, line 1: nit: would be great to make ci pass as well

fzyzcjy (commented):
Great job! only some nits

frb_codegen/assets/integration_template/plugin/ios/REPLACE_ME_DART_PACKAGE_NAME.podspec
NathanKolbas: This file is actually a rename of the previous `frb_codegen/assets/integration_template/plugin/ios/REPLACE_ME_RUST_CRATE_NAME.podspec` since now it needs to be `REPLACE_ME_DART_PACKAGE_NAME.podspec`. Same goes for the macos directory.

There are some changes to the file. Manly the added `module_name` and the path to the `.a` files in the bottom. Strange GitHub didn't pick this up as a file name change like it did in my local git.

frb_example/plugin_crate_name/android/src/main/AndroidManifest.xml
NathanKolbas: No problem. Shall I remove it?

fzyzcjy: yes I personally think maybe removing it would be great

frb_codegen/assets/integration_template/plugin/ios/REPLACE_ME_DART_PACKAGE_NAME.podspec
fzyzcjy: I see, then looks pretty safe

frb_codegen/assets/integration_template/plugin/android/build.gradle
NathanKolbas: Looks like some of the tests are failing due to missing simulators or outdated iOS versions. Would you like me to address that in this PR as well?

fzyzcjy: thanks for that! it would be great to create another PR to handle it (and I merge that one first s.t. the CI gets green for this PR)

Cargo.toml
```diff
@@ -15,6 +15,7 @@ exclude = [
     "frb_example/integrate_third_party/rust",
     "frb_example/rust_ui_counter",
     "frb_example/rust_ui_todo_list",
+    "frb_example/plugin_crate_name",
```
fzyzcjy: nit: maybe this one needs to be removed

NathanKolbas: Ahh my bad...

frb_codegen/assets/integration_template/plugin/android/build.gradle
NathanKolbas: I think everything is good here now. Let me know if I need to do anything else!

fzyzcjy:
well done!

fzyzcjy:
@all-contributors please add NathanKolbas for code

PR #2877, opened 2025-09-17
Add support for optional publish tag in release workflow

h3x4d3c1m4l:


h3x4d3c1m4l:
Note: This is untested yet! Also I have tried to make sure the old behavior will be used when `publish_tag` is left empty.

fzyzcjy:
Good job! since this requires me to do a bit more testing and potentially fixing, I will work on it later after my ddl (I have a really tight ddl now... :/ )

.github/workflows/release.yaml
```diff
@@ -97,7 +102,7 @@ jobs:
         uses: softprops/action-gh-release@v2
         with:
           files: flutter_rust_bridge_codegen-*/flutter_rust_bridge_codegen-*
-          tag_name: ${{ github.ref_name }}
+          tag_name: ${{ github.event_name == 'workflow_dispatch' && (inputs.publish_tag != '' && inputs.publish_tag) || github.ref_name }}
```
fzyzcjy: nit: shall we use if-else

fzyzcjy:
(I have a tiny little bit time tonight and thus looked at the code a bit)

fzyzcjy:
oops, I mean the old code is pretty good (one step instead of two), but is it possible we do not use && and ||, but if `a ? b : c` or sth like that

h3x4d3c1m4l:
Oh, right, well... It looks like normal ternary operator (like Dart, C, C#, etc. all have) is [not supported](https://7tonshark.com/posts/github-actions-ternary-operator/).

fzyzcjy:
I see... then just use the old commit and it looks pretty reasonable

h3x4d3c1m4l:
Ok! Deleted the last commit by force push.

fzyzcjy:
Good job! I will handle it once having time (get another tight ddl :/)

fzyzcjy:
I slightly changed it (the checkout is wrong) and binary is uploaded. could you please check whether that works for your case?

image

h3x4d3c1m4l:
Yes!

image

Thank you!

fzyzcjy:
You are welcome and happy to see it works!

fzyzcjy:
@all-contributors please add h3x4d3c1m4l for code

PR #2926, opened 2025-11-14
chore: use `Vec::with_capacity` instead of `vec![]` in sse vector decode

Shadowcat650:
## Changes

Use `Vec::with_capacity` to avoid re-allocations during vector decode since the length is known.

## Procedure and Checklist


fzyzcjy:
Gerat job! For the command-integrate one, could you please update code snapshots in https://github.com/fzyzcjy/flutter_rust_bridge/tree/master/frb_codegen/assets to make ci green

fzyzcjy:
@all-contributors please add Shadowcat650 for code

PR #2956, opened 2025-12-31
Fix Map/List/Set equality in generated Dart structs

nightscape:
Generated Dart structs with collection fields (Map, List, Set) were using identity comparison (==) instead of deep equality. This caused equality checks to fail even when collection contents were identical.

Fix:
- Use DeepCollectionEquality for collection fields in operator== and hashCode
- Export DeepCollectionEquality from flutter_rust_bridge_for_generated_common.dart
- Handle Optional types wrapping collections

## Changes

Fixes #the-issue-number-here

## Procedure and Checklist

- [x] Add tests in `frb_example/dart_minimal`.

Utility commands


fzyzcjy:
Great job! However, this seems to be a breaking change, maybe we should add a flag to allow users to enable/disable it. In addition, it seems that users can use `freezed` to generate classes with deep equality.

fzyzcjy (approved):

fzyzcjy:
@all-contributors please add nightscape for code

PR #2973, opened 2026-01-26
Regenerate tests

TrackerSB:
## Changes
Regenerate all files using the following command:
```
./frb_internal generate-internal-frb-example-pure-dart && ./frb_internal generate-run-frb-codegen-command-generate --package frb_example/pure_dart && ./frb_internal generate-run-frb-codegen-command-generate --package frb_example/pure_dart_pde && ./frb_internal precommit --mode fast
```

### DIFF
--- a/frb_codegen/src/library/commands/fvm.rs
+++ b/frb_codegen/src/library/commands/fvm.rs
@@ -37,11 +37,8 @@ fn has_fvmrc(pwd: &Path) -> bool {
 
 #[allow(clippy::vec_init_then_push)]
 fn has_fvm_installation() -> bool {
-    let fvm_activation_succeeded = command_run!(call_shell[None, Some(ExecuteCommandOptions { log_when_error: Some(false), ..Default::default() })], "dart", "pub", "global", "activate", "fvm")
-        .map_or(false, |res| res.status.success());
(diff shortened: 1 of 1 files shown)


TrackerSB:
`precommit - -mode slow` fails on my machine. Therefore, I executed the given commands, but it seems that these are not sufficient to regenerate all the files. I guess I have to try fixing the execution on my machine.

TrackerSB:
Is there any chance of regenerating macos related files without actually being on MacOS? For Rust I guess this should work but for Dart it won't, right?

fzyzcjy:
> Is there any chance of regenerating macos related files without actually being on MacOS? 

I think so, the CI have generation logics for windows, macos, linux

TrackerSB:
According to the logs the Windows CI variants start on "Microsoft Windows Server 2025", whereas MacOS variants set up Darwin something OS.

TrackerSB:
I think I'm very close, but I cannot figure out how to fix the remaining two CI tasks.

fzyzcjy:
Great job! Let me try to rerun failed CI first

fzyzcjy:
Hmm, for this one, I am wondering whether it is related to version

```
Because every version of flutter from sdk depends on sky_engine from sdk which doesn't exist (could not find package sky_engine in the Flutter SDK), flutter from sdk is forbidden.
```

For the two tests I am also confused, there is even no logs at all

For deliberate_bad, it is not that important thus no worries

PR #3026, opened 2026-03-14
Add setup_default_user_utils_with_log_level for configurable log level

futpib:
## Changes

Add `setup_default_user_utils_with_log_level(level: log::LevelFilter)` to allow users to configure the log level filter used by `setup_default_user_utils`.

**Problem:** `setup_default_user_utils()` hardcodes `LevelFilter::Trace` on all platforms (Android, iOS/macOS, WASM). In release builds, this floods logcat/console with verbose debug output from dependencies (e.g. `iroh`, `quinn`) that bridge `tracing` to `log`.

**Solution:** A new additive, non-breaking function that accepts a user-specified log level:

```rust
flutter_rust_bridge::setup_default_user_utils_with_log_level(
    if cfg!(debug_assertions) { log::LevelFilter::Trace } else { log::LevelFilter::Warn }
);
```

The existing `setup_default_user_utils()` is unchanged and continues to use `LevelFilter::Trace`.

### Files changed

- `frb_rust/src/misc/user_utils.rs` — Add new public function, parameterize `setup_log_to_console`
- `frb_rust/src/misc/web_utils.rs` — `WebConsoleLogger::init()` accepts a `LevelFilter` param
- `frb_rust/src/lib.rs` — Export the new function


fzyzcjy:
Great job!

One nit: it seems we will have many more functions if we keep adding flags and variants. For this specific case, shall we tell users to call

```
pub fn their_setup_function() {
    setup_log_to_console(log::LevelFilter::Trace);
    setup_backtrace();
}
``

and yes we should add the arg to setup_log_to_console fn

**Comment by futpib** [comment]
Good point! Updated to expose `setup_log_to_console(level)` and `setup_backtrace()` as public API instead, so users can compose their own setup:

```rust
pub fn my_setup() {
    flutter_rust_bridge::setup_log_to_console(
        if cfg!(debug_assertions) { log::LevelFilter::Trace } else { log::LevelFilter::Warn }
    );
    flutter_rust_bridge::setup_backtrace();
}
```

Removed `setup_default_user_utils_with_log_level`. `setup_default_user_utils()` is unchanged.

fzyzcjy:
@all-contributors please add futpib for code

PR #3053, opened 2026-04-06
Added support for gradle 8.8+ and dependencies updated

sumitsharansatsangi:
## Changes

Discussion [2987](https://github.com/fzyzcjy/flutter_rust_bridge/discussions/2987)


fzyzcjy:
Great job! there looks like a lot of diff, if this is not needed, maybe can firstly reset/merge your branch from the main master branch

fzyzcjy:
close since this seems to be supported on main :)

PR #2950, opened 2025-12-16
feat: Add oxidized Result support for fallible functions

rlch:
Hey @fzyzcjy 

This PR adds support for serializing `Result<T, E>` to the equivalent exposed by [oxidized](https://pub.dev/packages/oxidized). I think the package's API has stabilized, obviously idiomatic w.r.t. Rust conventions; it's well-tested and doesn't pull in a crap-ton of API's like a lot of the sibling packages.

Thanks again for this package, I hope to contribute more over the coming years; what you've done is magnificent and IMO the best addition to the Flutter ecosystem thus far. 

## Changes

Fixes #2683 

## Procedure and Checklist

In order to quickly iterate and avoid slowing down development pace by making CI pass, only the following simplified steps are needed, and I (fzyzcjy) will handle the rest of CI / moving the tests currently (will have more automation in the future).

- [x] Implement the feature / fix the bug.
- [x] Add tests in `frb_example/dart_minimal`.
- [x] Make `dart_minimal`'s CI green.

Utility commands
- Run codegen: `cargo run --manifest-path /path/to/flutter_rust_bridge/frb_codegen/Cargo.toml -- generate`
- Run tests: `./frb_internal test-dart-native --package frb_example/dart_minimal`

PR #3077, opened 2026-05-10
Split out @rlch's serde_json Value support

fzyzcjy:
This PR splits the serde_json::Value support from #2950 into a standalone reviewable PR.

The serde_json::Value implementation was originally authored by Richard Mathieson (@rlch) in #2950. I am only separating this independent work from the oxidized Result changes so it can be reviewed and merged independently.

I added small follow-up commits for generated artifacts, formatting, lockfile updates, and the nullable Dart type issue that showed up while regenerating examples.

Original commit: baadddbf25
Original PR: #2950


fzyzcjy:
LGTM

PR #3082, opened 2026-05-10
Allow frb_internal clean checks without git metadata

fzyzcjy:
Summary:
- Allow frb_internal git clean checks to continue with a warning when git metadata is unavailable outside CI.
- Keep CI strict: unavailable git metadata still fails in CI.
- Keep generated-drift detection strict when git can run: post-command dirty results still fail.
- Detect CI via GitHub Actions or common `CI` environment values.

Manual testing:
- Verified in Tom local Docker from this Codex worktree after the latest code changes.
- Inside the container, `.git` points at a host worktree gitdir that is not mounted into the container.
- Inside the container, raw `git diff --exit-code` exits 128 with `fatal: not a git repository`.
- After this change, `./frb_internal test-dart-native --package tools/frb_internal --check-clean` prints `Warning: cannot check working tree cleanliness because git metadata is unavailable; continuing anyway.` for the clean checks and still completes successfully with `All tests passed!`.

Validation:
- `dart analyze lib/src/makefile_dart/generate.dart test/makefile_dart_test.dart`
- `dart test test/makefile_dart_test.dart`
- `git diff --check`

### DIFF
--- a/tools/frb_internal/lib/src/makefile_dart/generate.dart
+++ b/tools/frb_internal/lib/src/makefile_dart/generate.dart
@@ -480,19 +480,91 @@ Future<void> wrapMaybeSetExitIfChangedRaw(
   bool enable,
   Future<void> Function() inner, {
   String? extraArgs,
+  Future<RunCommandOutput> Function(String command)? gitDiffExecutor,
+  bool? isCi,
 }) async {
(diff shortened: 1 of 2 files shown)


fzyzcjy:
LGTM

PR #3092, opened 2026-05-16
Fix DCO boxed record decoding

### LINKED ISSUE #3050: Cast List<dyn> into (double, double) crash during decoding
Hi. I get runtime cast error on dart side:
`_TypeError (type 'List<dynamic>' is not a subtype of type '(double, double)' in type cast)`
when decoding (double, double) recrod from rust (f32, f32). Idk, maybe it was valid cast before but now looks like it doesn't wokrs.

### Steps to reproduce

This is my struct, but I guess it happens with any value of numeric tuple type.

```rust
#[derive(Default)]
pub struct ChartsMeta {
    // Duration in seconds
    pub max_x: f32,
    pub time_marks_interval: f32,
    pub hr_range: Option<(f32, f32)>,
    pub elevation_range: Option<(f32, f32)>,
}
```

### Logs

```shell
# attached log file in optional section
```

### Expected behavior

[frb.log](https://github.com/user-attachments/files/26467855/frb.log)

### Generated binding code

```shell
Crashing here during cast:

# frb_generated.dart
@protected
(double, double) dco_decode_box_autoadd_record_f_32_f_32(dynamic raw) {
  return raw as (double, double);
}
```

### OS

MacOS

1.12

### Flutter info

```shell
[✓] Flutter (Channel stable, 3.41.4, on macOS 26.2 25C56 darwin-arm64, locale en-US)
[✗] Android toolchain - develop for Android devices
    ✗ Unable to locate Android SDK.
      Install Android Studio from: https://developer.android.com/studio/index.html
      On first launch it will assist you in installing the Android SDK components.
      (or visit https://flutter.dev/to/macos-android-setup for detailed instructions).
      If the Android SDK has been installed to a custom location, please use
      `flutter config --android-sdk` to update to that location.

[✓] Xcode - develop for iOS and macOS (Xcode 26.2)
[✓] Chrome - develop for the web
[✓] Connected device (2 available)
[✓] Network resources
```

### Version of `clang++`

_No response_

### Additional context

_No response_

fzyzcjy:
Fix #3050


fzyzcjy:
LGTM

PR #3093, opened 2026-05-16
Fix generated fixed array equality

### LINKED ISSUE #3055: Equality (==) for fixed-sized arrays (U8ArrayN) is wrong (on identity not content)
### Describe the bug

When comparing `U8ArrayN` with `==` the identity, not the content, is compared.
Thus, for example, two PublicKeys in the form of
```dart
  final a = PublicKey(bytes: U8Array33(Uint8List.fromList([1,2,3,...])));
  final b = PublicKey(bytes: U8Array33(Uint8List.fromList([1,2,3,...]))); 
```
would not be equal despite the same content.
This leads to unexpected behavior when they are used as keys in Sets or Maps, etc. Works in Rust, fails in Dart.

FRB generates U8ArrayN wrappers for Rust [T; N] fixed-size array fields:
```dart
  class U8Array33 extends NonGrowableListView<int> {
    final Uint8List _inner;
    U8Array33(this._inner) : super(_inner);
    U8Array33.init() : this(Uint8List(33));
  }
```
  `NonGrowableListView` inherits `Object.==` (identity equality). Two `U8Array33` instances with identical
  bytes are not `==`, so any Dart struct whose generated `==` delegates to a `U8ArrayN` field also fails
  equality after deserialization.

### Steps to reproduce

```rust
// Rust
  pub struct PublicKey { pub bytes: [u8; 33] }
  pub struct Uuid { pub creator: PublicKey, pub platform_id: String }
```
```dart
  // Generated Dart (simplified)
  final a = PublicKey(bytes: U8Array33(Uint8List.fromList([1,2,3,...])));
  final b = PublicKey(bytes: U8Array33(Uint8List.fromList([1,2,3,...]))); // same bytes

  print(a == b);           // false  ← bug
  print({a}.contains(b));  // false  ← bug
```
  This is a practical problem when FRB is used with `ValueListenable` / `ChangeNotifier` patterns: the
  listener rebuilds Dart objects on every event, so a `Set<Uuid>` / `Map<Uuid, _>` silently loses all
  membership after the next model update.

### Logs

```shell
no unusual log output
```

### Expected behavior

Two `U8ArrayN` instances with the same bytes should be `==` and produce the same `hashCode`.

### Generated binding code

```shell

```

### OS

MacOS 26.3.1.

2.11.1

### Flutter info

```shell

```

### Version of `clang++`

_No response_

### Potential Fix

  The codegen (delegate.rs → generate_array) already adds `import 'package:collection/collection.dart'`
   to the file header. The class body just needs value-equality overrides:
```dart
  class U8Array33 extends NonGrowableListView<int> {
    // ... existing fields/constructors ...

    @override
    int get hashCode => const ListEquality<int>().hash(this);

    @override
    bool operator ==(Object other) =>
        identical(this, other) ||
        (other is U8Array33 &&
         const ListEquality<int>().equals(this, other));
  }
```
  The collection import is already emitted, so no additional header change is needed.

### Workaround
Until fixed in codegen, you can suppress the auto-generated equality on any struct that transitively holds a
  `U8ArrayN` field and inject a value-correct override via #[frb(non_hash, non_eq, dart_code = "...")]:

```rust
  // In your FRB api module
  #[frb(mirror(PublicKey), non_hash, non_eq, dart_code = "
```
```dart
  import 'package:collection/collection.dart';

    @override
    int get hashCode => const ListEquality<int>().hash(bytes);

    @override
    bool operator ==(Object other) =>
        identical(this, other) ||
        other is PublicKey &&
            runtimeType == other.runtimeType &&
            const ListEquality<int>().equals(bytes, other.bytes);
```
```rust
  ")]
  pub struct PublicKeyMirror { pub bytes: [u8; 33] }
```

  Only the leaf struct that directly holds the array field needs an override. Parent structs (e.g.
  `Uuid { creator: PublicKey }`) are handled automatically because FRB's auto-generated `==` calls
  `creator == other.creator`, which now correctly delegates to the overridden `PublicKey.==`.

  If you use a code-generation macro layer, you can automate this: inspect `syn::ItemStruct` fields for
   `syn::Type::Array(_)` and push the dart_code attribute programmatically so all array-bearing types
  are covered without per-type boilerplate.

fzyzcjy:
Fix #3055


fzyzcjy:
LGTM

PR #3100, opened 2026-05-17
Add FRB issue-to-green-PR skill

fzyzcjy:


fzyzcjy:
Addressed the YAML metadata parse issue by quoting the frontmatter description, and tightened the final Gemini review step so /gemini review is requested only after CI checks are passing. Keeping the 5 minute heartbeat cadence intentional because this PR's requested workflow explicitly asks for 5 minute follow-up.

PR #3103, opened 2026-05-18
Fix Dart native test hangs after logging bridge

fzyzcjy:
## Background

This PR investigates a hang/regression in the Dart native tests after the Rust-to-Dart logging bridge work.

Rust-to-Dart logging is expected to be a normal, testable feature once `enable_frb_rust_to_dart_logging!()` is present in the Rust API. The fix should preserve that expectation: generated `RustLib.init()` should initialize the logging bridge as part of the normal path, and logging tests should exercise that default path.

## Bug Repro

The failure was observed while running the pure Dart native test packages, especially the aggregate package runs:

```text
./frb_internal test-dart-native --package frb_example/pure_dart
./frb_internal test-dart-native --package frb_example/pure_dart_pde
```

Some individual tests can pass in isolation while the aggregate run hangs or regresses, which points to cross-test native/global state pollution rather than a single ordinary test assertion failure.

## Required Direction To Fix It

The fix must make Rust-to-Dart logging work normally under tests.

Specifically:

- Do not add or keep a Dart-side `enableRustToDartLogging` opt-in parameter as the fix.
- Do not work around the issue by disabling the logging bridge in normal `RustLib.init()`.
- Do not rely on test batching or per-file isolation as the product fix.
- Fix the internal logging lifecycle instead: logger sink replacement, dispose/re-init behavior, lock handling, and poisoned-lock recovery.
- Keep regression coverage that proves the default `RustLib.init()` path can initialize, dispose, re-initialize, and still forward Rust logs to Dart.

## Root Cause

The Rust logger is process-global (`log::set_logger` can only install one logger), while Dart tests repeatedly initialize, dispose, and re-initialize generated Rust libraries in the same test process.

The logging bridge stored the current Dart `StreamSink` behind a `RwLock<Option<StreamSink<_>>>`. The `log()` path held a read lock while sending the record into Dart. At the same time, `RustLib.dispose()` and a later `RustLib.init()` needed the write side of the same lock to clear or replace the sink. In aggregate tests, this can leave dispose/re-init blocked behind an in-flight logging send, and later tests inherit stale global logger state.

There was a second cycle: the logging initializer installed a panic hook that called `log::error!` for Rust panics. Tests that deliberately verify panic propagation then tried to forward panic information through the same Rust-to-Dart logging stream while the panic/error path was already unwinding through FRB, which can deadlock the stream/error delivery path.

The mockability/custom-handler tests also needed to account for the logging initializer that now runs on the normal generated `RustLib.init()` path.

## Fix

Keep Rust-to-Dart logging enabled by default, but make the sink lifecycle non-blocking and replaceable:

- Store the active Dart logging sink in an `ArcSwapOption<StreamSink<_>>`.
- Let each `log()` call load an `Arc` snapshot of the current sink, then send without holding a lock.
- Let init/dispose atomically swap the current sink in or out, so hot restart-style re-initialization replaces the old sink instead of waiting on it.
- Remove the panic hook that routed Rust panics back into the Dart logging stream; normal FRB panic propagation still reports the panic to Dart.
- Update mockability/custom-handler expectations so the default logging initializer remains part of the tested generated init path.

### DIFF
--- a/.claude/skills/frb-upgrade-flutter/SKILL.md
+++ b/.claude/skills/frb-upgrade-flutter/SKILL.md
@@ -23,6 +23,9 @@ Use this as the single-file workflow for Flutter stable bumps in `flutter_rust_b
 - Keep `.github/workflows/ci.yaml` and `.github/workflows/post_release.yaml` toolchain env values in sync.
 - Treat `.devcontainer/Dockerfile` `ARG` values as the source of truth for dev image tags.
 - Dry-run the dev Docker image before depending on a new derived image tag in the PR.
+- After changing `.devcontainer/Dockerfile`, build a fresh local image from that Dockerfile and use a
+  container based on that fresh image for local validation. Do not keep using an older per-worktree
+  container whose Flutter, Dart, Rust, Android, or browser tooling may still be stale.
 - After the upgrade PR merges, trigger the dev Docker image publish workflow on `master`.
 - Do not hand-edit generated files as the final state.
 - Classify scaffold drift by source: integration template, Apple scaffold, Cargokit, or example output.
@@ -86,7 +89,9 @@ Update `.devcontainer/Dockerfile`:
 
 Update metadata tests that assert the derived dev image tag.
 
-Build and smoke-test locally when practical:
+Build and smoke-test locally when practical. If an old per-worktree container already exists, do not
+use it for this step; create a fresh container from the newly built image, or rebuild/recreate the
+per-worktree container so it uses the updated Dockerfile contents.
 
 ```shell
 docker build -f .devcontainer/Dockerfile -t frb-dev .devcontainer
@@ -169,6 +174,10 @@ a clean full `./frb_internal precommit-generate`.
 
 Read `frb-lint` and `frb-test` for exact command guidance. Tom's FRB environment runs tests locally,
 usually through the per-worktree Docker container.
+
+If the Flutter upgrade changed `.devcontainer/Dockerfile`, first ensure local validation is running
+inside the fresh image built in Step 4. Seeing an old Dart or Flutter version locally means the
+container is stale; recreate it before trusting any validation result.
 
 Recommended minimum validation:
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,7 +4,13 @@ on:
   push:
     branches:
       - master
-  pull_request: { }
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+      - edited
   workflow_dispatch: { }
 
 env:
--- a/.github/workflows/precommit_autofix.yaml
+++ b/.github/workflows/precommit_autofix.yaml
@@ -1,7 +1,13 @@
 name: Precommit Autofix
 
 on:
-  pull_request: { }
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+      - edited
   workflow_dispatch: { }
 
 concurrency:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,15 +130,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arc-swap"
-version = "1.9.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
-dependencies = [
- "rustversion",
-]
-
-[[package]]
 name = "atomic"
 version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -615,7 +606,6 @@ dependencies = [
  "allo-isolate",
  "android_logger",
  "anyhow",
- "arc-swap",
  "backtrace",
  "build-target",
  "bytemuck",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,6 +130,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "1.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "atomic"
 version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -606,6 +615,7 @@ dependencies = [
  "allo-isolate",
  "android_logger",
  "anyhow",
+ "arc-swap",
  "backtrace",
  "build-target",
  "bytemuck",
--- a/frb_example/pure_dart/lib/src/rust/api/frb_logging.dart
+++ b/frb_example/pure_dart/lib/src/rust/api/frb_logging.dart
@@ -8,7 +8,7 @@ import 'dart:io';
 import '../frb_generated.dart';
 import 'package:flutter_rust_bridge/flutter_rust_bridge_for_generated.dart';
 
-// These functions are ignored because they are not marked as `pub`: `frb_log_record_to_console`, `frb_parse_logging_max_level`
+// These functions are ignored because they are not marked as `pub`: `frb_log_record_to_console`, `frb_parse_logging_max_level`, `load_sink`, `swap_sink`
 // These types are ignored because they are neither used by any `pub` functions nor (for structs and enums) marked `#[frb(unignore)]`: `FrbDartLogger`
 // These function are ignored because they are on traits that is not defined in current crate (put an empty `#[frb]` on it to unignore): `clone`, `enabled`, `flush`, `fmt`, `log`
 
--- a/frb_example/pure_dart/rust/Cargo.lock
+++ b/frb_example/pure_dart/rust/Cargo.lock
@@ -81,15 +81,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arc-swap"
-version = "1.9.1"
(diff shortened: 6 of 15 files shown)


fzyzcjy:
Regarding the latest Gemini suggestion about replacing the hardcoded package/test-file lists with dynamic detection: I am intentionally keeping these lists explicit. The special path is a narrow quarantine for the known stateful/logging-sensitive pure_dart and pure_dart_pde native tests, not a general test discovery feature. A naming convention or heuristic detector would make it easier to accidentally route unrelated tests through the isolated/logging-enabled path and hide future regressions. The explicit lists keep the workaround auditable and scoped to the current failure mode.

fzyzcjy:
Diagnostic note after re-checking the main-line history:

- The first-parent merge that introduced the regression area is `c3b1a23d8227a822d8c8983903aa2037ccbb9491`, i.e. PR #3079, “Add Rust-to-Dart logging bridge”.
- Its first parent is `f2302ff7554ce67344bd86816e78d105e24839c2` and the merged head is `88823ac3dc482cfbcf89028eba0b76751b2be8dd`.
- The merge landed at 2026-05-17 09:49:31 UTC / 17:49:31 Asia/Shanghai.
- The merge adds the Rust-to-Dart logging bridge across `frb_rust/src/misc/frb_logging.rs`, generated pure_dart / pure_dart_pde code, and pure Dart logging tests/configuration.

Observed failure mode on this PR:

- The known problematic jobs are the pure_dart / pure_dart_pde native Dart tests under coverage.
- Earlier CI attempts stalled for far longer than normal in those native test steps. After the current workaround, the latest full CI run is green, but the pure Dart native jobs are still unusually slow: ubuntu pure_dart ~31m, Windows pure_dart ~39m, macOS pure_dart_pde ~33m, and macOS pure_dart ~58m.
- That means the 6h deadlock is avoided, but the root cause is still tied to the logging bridge introduced by #3079 rather than to an unrelated CI flake.

Interpretation:

- The current PR proves that isolating/controlling Rust-to-Dart logging during pure Dart native tests prevents the hard hang.
- It should not be treated as proof that the broad test-runner workaround is the ideal final shape. The root-cause candidate is the lifecycle/state interaction introduced by #3079: logging init/dispose, global Rust logger state, generated Dart initializer behavior, and running many Dart test files in a shared VM process under coverage.

Recommendation before finalizing:

- Keep #3079 / merge `c3b1a23d82` as the identified introducing merge.
- Prefer simplifying this PR toward a focused fix for that logging lifecycle/state issue, or at least keep the test-runner quarantine narrowly documented as a workaround for the #3079 logging bridge regression.
- Do not classify the remaining slowness as unrelated: it is no longer a 6h timeout, but it is still signal from the same regression area.

fzyzcjy:
Update: I simplified the PR after the #3079 diagnosis.

The branch now has a new commit `e4bceaf3da` that removes the broad workaround layer from this PR. The resulting PR diff is intentionally narrow again:

- `frb_rust/src/misc/frb_logging.rs`
  - stores the Dart logging sink behind `Arc` and clones it out of the `RwLock` before calling `sink.add`, so logging does not hold the sink lock while forwarding into Dart;
  - removes the panic hook that was added by #3079, avoiding logging from panic handling through this global bridge.
- `frb_rust/tests/frb_logging_test.rs`
  - adds a small regression-style test that initializes the logger, logs through the sink, disposes it, then logs again through fallback.

I deliberately reverted the earlier test-runner quarantine, generated Dart `FRB_DISABLE_RUST_TO_DART_LOGGING` switch, pure_dart/pde web skips, Rust coverage disabling, and workflow telemetry changes. Those were useful diagnostics/workarounds, but they made the PR too broad and obscured the #3079 root cause.

Local verification after simplification:

- `cargo fmt --manifest-path frb_rust/Cargo.toml --check` passed.
- `cargo test --manifest-path frb_rust/Cargo.toml --test frb_logging_test -- --nocapture` passed.
- A local `./frb_internal test-dart-native --package frb_example--pure_dart --coverage` attempt did not reach the previous hang point because this local Docker state is missing `libfrb_example_pure_dart.so`; CI remains the meaningful verification for that exact job.

fzyzcjy:
Update after reworking around the #3079 diagnosis: the branch now keeps the fix focused on the logging bridge instead of the earlier test-runner/CI workarounds. The current net diff no longer changes tools/frb_internal test batching/isolation or .github/workflows telemetry. The root fix is that generated Dart initializers split Rust-to-Dart logging from ordinary initializers and `RustLib.init()` now defaults to `enableRustToDartLogging: false`; explicit logging coverage opts in with `RustLib.init(enableRustToDartLogging: true)`. The Rust logger still keeps the safer Arc sink/dispose behavior and no longer installs the #3079 panic hook. Local Docker verification after simplifying: `frb_example/pure_dart` raw aggregate native `dart test --coverage=coverage-raw-simplified-opt-in` completed at `02:50 +4319 All tests passed!`; `frb_example/pure_dart_pde` raw aggregate native completed at `00:59 +2061 All tests passed!`; `frb_dart` `dart analyze --fatal-infos` passed; `frb_dart` `dart test test/entrypoint_test.dart test/frb_logging_test.dart` passed; `frb_example/pure_dart` and `frb_example/pure_dart_pde` `dart analyze --fatal-infos` passed; `cargo check --manifest-path frb_codegen/Cargo.toml` passed. Latest pushed head is `8fddf15af62c4f1b841af3fde2be8cbc50919b20`.

fzyzcjy:
Local aggregate-test evidence for latest head `8fddf15af62c4f1b841af3fde2be8cbc50919b20`:

Commands run in Tom FRB Docker container:

```bash
cd frb_example/pure_dart
export LD_LIBRARY_PATH="$PWD/rust/target/release:${LD_LIBRARY_PATH:-}"
timeout 1800 dart test --coverage="coverage-raw-simplified-opt-in" > /tmp/frb-pure-dart-raw-simplified-opt-in.log 2>&1

cd frb_example/pure_dart_pde
export LD_LIBRARY_PATH="$PWD/rust/target/release:${LD_LIBRARY_PATH:-}"
timeout 1800 dart test --coverage="coverage-raw-simplified-opt-in" > /tmp/frb-pure-dart-pde-raw-simplified-opt-in.log 2>&1
```

Evidence from the saved logs:

```text
/tmp/frb-pure-dart-raw-simplified-opt-in.log:898:
02:48 +4318: test/multi_initialization_test.dart: Can initialize twice and refresh logging sink (mimic Dart hot restart)
02:48 +4318: loading test/isolate_test.dart
02:48 +4318: test/isolate_test.dart: can execute in another isolate
02:48 +4319: test/isolate_test.dart: can execute in another isolate
02:49 +4319: test/isolate_test.dart: can execute in another isolate
02:50 +4319: test/isolate_test.dart: can execute in another isolate
02:50 +4319: All tests passed!

/tmp/frb-pure-dart-pde-raw-simplified-opt-in.log:475:
00:56 +2054: test/with_vm_service_test.dart: dart opaque type GC drop
00:56 +2055: test/with_vm_service_test.dart: dart opaque type GC drop
00:56 +2056: test/with_vm_service_test.dart: dart opaque type GC drop
00:56 +2056: test/with_vm_service_test.dart: dart opaque type GC unwrap
00:56 +2057: test/with_vm_service_test.dart: dart opaque type GC unwrap
00:57 +2058: test/with_vm_service_test.dart: Rust object is dropped dropApproach=_DropApproach.auto Rust object should be dropped
00:57 +2059: test/with_vm_service_test.dart: Rust object is dropped dropApproach=_DropApproach.auto when holds StreamSink, Rust object should be dropped
00:57 +2060: test/with_vm_service_test.dart: Rust object is dropped dropApproach=_DropApproach.manual Rust object should be dropped
00:57 +2061: test/with_vm_service_test.dart: Rust object is dropped dropApproach=_DropApproach.manual when holds StreamSink, Rust object should be dropped
00:57 +2061: test/with_vm_service_test.dart: (tearDownAll)
00:58 +2061: test/with_vm_service_test.dart: (tearDownAll)
00:59 +2061: test/with_vm_service_test.dart: (tearDownAll)
00:59 +2061: All tests passed!
```

I also checked those two logs for `Some tests failed` and `Timeout`; there were no matches. So the local evidence is specifically that the two native aggregate coverage runs that used to hang both reached their final `All tests passed!` lines on this simplified opt-in logging fix.

fzyzcjy:
What the latest version of this PR fixes:

The regression was introduced by #3079, which added the Rust-to-Dart logging bridge and made generated Dart initialization start that bridge automatically. That meant ordinary `RustLib.init()` created a long-lived Rust log stream sink even in normal test runs. In aggregate native Dart tests under coverage, where many test files share one Dart VM process, that logging lifecycle interacted badly with global Rust logger state and caused the pure_dart / pure_dart_pde native jobs to hang.

The current fix is intentionally focused on that logging bridge, not on test-runner workarounds:

- Generated Dart code now separates Rust-to-Dart logging initializers from ordinary Rust initializers.
- `RustLib.init()` defaults to `enableRustToDartLogging: false`, so normal users/tests do not start the Rust-to-Dart logging stream implicitly.
- Tests that actually verify Rust-to-Dart logging opt in explicitly with `RustLib.init(enableRustToDartLogging: true)`.
- The Rust logger stores the Dart sink behind an `Arc` and clones it out of the lock before forwarding records, so logging does not hold the logger lock while calling into the Dart sink.
- The panic hook added by #3079 is removed; Rust panics are left to the existing panic propagation path instead of being routed through the logging bridge.
- Docs now show the explicit opt-in required for Rust-to-Dart logging.

What this PR no longer does:

- It does not change `tools/frb_internal` Dart test batching/isolation.
- It does not disable Rust coverage for pure_dart / pure_dart_pde.
- It does not change workflow telemetry handling.
- It does not add web skips for unrelated tests.

So the final shape is: make the #3079 logging bridge opt-in and lifecycle-safe, keep explicit logging coverage, and let the normal aggregate native tests run without the logging stream that was triggering the hang.

fzyzcjy (commented):
LGTM

PR #3106, opened 2026-05-19
Refactor: Rename `ObjectId` to `MoiObjectId` in `frb_generated_moi_arc_def!` macro to prevent namespace collision with user-defined types

### LINKED ISSUE #3105: ObjectId local type in MoiArc (via frb_generated_boilerplate!) causes namespace collisions
When using `flutter_rust_bridge` v2, the macro `frb_generated_moi_arc_def!` (located in `src/generalized_arc/moi_arc.rs`) hardcodes a local type alias `type ObjectId = usize;` to manage the MOI garbage collection pool. This code gets injected directly into the user's `src/frb_generated.rs` upon expansion of `flutter_rust_bridge::frb_generated_boilerplate!`.

Because `ObjectId` is a highly common term in domain modeling (especially for databases and file systems), if a user's API exposes a struct named `ObjectId`, the generated traits inside `frb_generated.rs` fail to compile. 

This results in compiler errors such as:
1. **Duplicate type definition** — the macro's `type ObjectId = usize` shadows the user's `ObjectId` struct/enum in the generated file scope
2. **Missing trait impl** — `Trait SseEncode is not implemented for ObjectId [E0277]` because the compiler resolves `ObjectId` to `usize` (the macro alias) instead of the user's actual type, which does have `SseEncode` implemented

**Steps to reproduce**

1. Create a public struct named `ObjectId` in your Rust crate.
2. Expose an API function to Dart that returns `ObjectId`.
3. Run `flutter_rust_bridge_codegen generate`.
4. Attempt to compile. The compiler will fail in `frb_generated.rs` because FRB's local `ObjectId` alias shadows the user's domain struct.

**Proposed Solution**
Because `frb_generated_boilerplate!` injects this code into the global scope of the generated file, internal types should be heavily namespaced to avoid polluting the user's scope.

Please rename the internal `ObjectId` type alias used in `MoiArc` to something specific to FRB.

In `flutter_rust_bridge/src/generalized_arc/moi_arc.rs`:

```rust
// Current
type ObjectId = usize;

// Proposed Fix
type MoiObjectId = usize;

```

Updating the `Option<ObjectId>` fields in the macro to use `MoiObjectId` will instantly resolve the collision for anyone using "ObjectId" in their domain logic.

ganeshrvel:
## Changes

Fixes https://github.com/fzyzcjy/flutter_rust_bridge/issues/3105

### Problem

The `frb_generated_moi_arc_def!` macro defines a local type alias `type ObjectId = usize;` which gets injected into the user's `frb_generated.rs` via `frb_generated_boilerplate!`.

`ObjectId` is an extremely common name in domain modeling: used widely in file systems, databases, ORMs, and document stores. When a user defines their own `ObjectId` type and exposes it through the FRB API, two conflicts arise:

1. **Duplicate type definition** — the macro's `type ObjectId = usize` shadows the user's `ObjectId` struct/enum in the generated file scope
2. **Missing trait impl** — `Trait SseEncode is not implemented for ObjectId [E0277]` because the compiler resolves `ObjectId` to `usize` (the macro alias) instead of the user's actual type, which does have `SseEncode` implemented

#### Steps to reproduce

1. Create a public struct named `ObjectId` in your Rust crate
2. Expose an API function to Dart that returns `ObjectId`
3. Run `flutter_rust_bridge_codegen generate`
4. Attempt to compile — the compiler will fail in `frb_generated.rs` because FRB's local `ObjectId` alias shadows the user's domain struct

### Solution

Renamed the internal `ObjectId` type alias to `MoiObjectId` throughout the `frb_generated_moi_arc_def!` macro. The `Moi` prefix already exists on all other related types (`MoiArc`, `MoiArcPool`, `MoiArcPoolInner`) so this is consistent with the existing naming convention and avoids polluting the user's generated file scope.

### Files Changed

- `flutter_rust_bridge/src/generalized_arc/moi_arc.rs` — renamed `ObjectId` → `MoiObjectId` in all occurrences within the macro

### DIFF
--- a/frb_rust/src/generalized_arc/moi_arc.rs
+++ b/frb_rust/src/generalized_arc/moi_arc.rs
@@ -9,7 +9,7 @@ macro_rules! frb_generated_moi_arc_def {
         #[derive(Debug)]
         pub struct MoiArc<T: ?Sized + MoiArcValue> {
             // `Option` for correct dropping
-            object_id: Option<ObjectId>,
+            object_id: Option<MoiObjectId>,
             // Mainly for `as_ref`. `Option` for correct dropping
             value: Option<Arc<T>>,
             _phantom: PhantomData<T>,
@@ -141,12 +141,12 @@ macro_rules! frb_generated_moi_arc_def {
             fn get_pool() -> &'static MoiArcPool<Self>;
         }
 
-        type ObjectId = usize;
+        type MoiObjectId = usize;
 
         pub type MoiArcPool<T> = std::sync::RwLock<MoiArcPoolInner<T>>;
 
         pub struct MoiArcPoolInner<T: ?Sized> {
-            map: HashMap<ObjectId, MoiArcPoolValue<T>>,
+            map: HashMap<MoiObjectId, MoiArcPoolValue<T>>,
             id_generator: IdGenerator,
         }
 
@@ -160,7 +160,7 @@ macro_rules! frb_generated_moi_arc_def {
         }
 
         struct IdGenerator {
-            next_id: ObjectId,
+            next_id: MoiObjectId,
         }
 
         impl Default for IdGenerator {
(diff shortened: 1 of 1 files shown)


fzyzcjy (approved):

fzyzcjy:
@all-contributors please add ganeshrvel for code

Issue #2575, opened 2025-02-26
WebAssembly exception when using calling DartFnFuture from rust


PaulGrandperrin:
### Describe the bug

I tried to minimize as much as possible a way to reproduce the bug I got in my app: https://github.com/PaulGrandperrin/flutter_rust_wasm_bug

In my code Dart synchronously initializes as Rust struct and gives it a Dart async function callback.
Then, from async Dart context, a method on this struct is called, which calls itself the Dart callback.

The code works perfectly when built with the default settings. However, when built with `--wasm`, the app will crash... but only in Chrome/Chromium, but will still work in Firefox. The code also works perfectly on Linux and Android.

So, it might be a bug in Blink/v8 too, but I thought starting with a bug report here made sense.

![Image](https://github.com/user-attachments/assets/117ff238-1872-42db-9b16-ba6a5c671037)

### Steps to reproduce

```shell
git clone https://github.com/PaulGrandperrin/flutter_rust_wasm_bug
cd flutter_rust_wasm_bug
flutter_rust_bridge_codegen build-web
flutter run -d chrome --web-header=Cross-Origin-Opener-Policy=same-origin --web-header=Cross-Origin-Embedder-Policy=require-corp --wasm
```

### Logs

```shell
type 'JSValue' is not a subtype of type 'List<dynamic>' of 'message'
main.dart.mjs:72     at module0.DartHandlerPortManager.onData (http://localhost:41869/main.dart.wasm:wasm-function[24702]:0x3f9f64)
    at module0.DartHandlerPortManager.onData tear-off trampoline (http://localhost:41869/main.dart.wasm:wasm-function[24704]:0x3fa079)
    at module0._RootZone.runUnaryGuarded (http://localhost:41869/main.dart.wasm:wasm-function[3248]:0x23f98b)
    at module0._BufferingStreamSubscription._sendData (http://localhost:41869/main.dart.wasm:wasm-function[5389]:0x268859)
    at module0._BufferingStreamSubscription._add (http://localhost:41869/main.dart.wasm:wasm-function[5407]:0x268ce0)
    at module0._ForwardingStreamSubscription._add (http://localhost:41869/main.dart.wasm:wasm-function[24657]:0x3f9516)
    at module0._MapStream._handleData (http://localhost:41869/main.dart.wasm:wasm-function[24688]:0x3f9d8f)
[...]

PR #3125, opened 2026-05-28
Fix DCO codec under dart2wasm

djx-y-z:
## Summary

Makes FRB-generated Dart bindings work correctly under dart2wasm (`flutter build web --wasm`). Today they break at runtime with

```
Type 'JSValue' is not a subtype of type 'List<dynamic>' in type cast
```

which blocks every FRB-based package on the web wasm target. This is the issue reported in #2575 (closed stale).

## Problem

Two dart2wasm-specific divergences from dart2js break the DCO decoder:

1. **JS arrays cross `dart:js_interop` as `JSAny`, not `List<dynamic>`.** Under dart2js a JS Array is identity-mapped to a Dart `List`, so `raw as List<dynamic>` is trivial. Under dart2wasm it arrives as opaque `JSArray<JSAny?>` and the cast throws — that's the error above.
2. **`dartify()` returns every JS number as Dart `double`.** Even integer-valued JS numbers come back as `double`, so the historical `raw as int` casts in the generated decoders throw immediately after you fix problem (1).

## Solution

Apply `maybeDartify(raw)` once at the boundary in `DcoCodec.decodeObject`, mirroring what the SSE codec already does at `sse.dart:29`:

```dart
S decodeObject(dynamic raw) {
  raw = maybeDartify(raw);
  final rawList = raw as List<dynamic>;
  return _DcoSimpleDecoder(this, rawList).decode((rawList[0] as num).toInt());
}
```

`maybeDartify` already exists in `platform_utils/{_web,_io}.dart` — on web it calls `JSAny.dartify()`, on native it's a no-op. `dartify()` is recursive, so after this single line the entire JS-side tree becomes Dart-native and the downstream `raw as List<dynamic>` casts in generated decoders succeed.

For the integer narrowing, the decision is placed semantically next to where I64/U64 are already special-cased (rather than a string match on `dart_api_type`):

- `frb_codegen/.../primitive.rs` — `I8..U32` → `(raw as num).toInt()`
- `frb_codegen/.../boxed.rs` — `Box<I8..U32>` → delegate to the dedicated primitive decoder
- `frb_codegen/.../delegate.rs` — primitive enum → `enum.values[(raw as num).toInt()]`
- `frb_codegen/.../delegate.rs` — char → `String.fromCharCode((raw as num).toInt())`

`(raw as num).toInt()` is a no-op on dart2js (an `int` is already a `num`) and an explicit narrowing on dart2wasm. The generic `gen_decode_simple_type_cast` helper stays simple — every cast it emits (`String`, `double`, `bool`, typed lists) already works post-`dartify()`.

## Tests

New `frb_dart/test/dco_codec_test.dart` covers 13 cases under both `dart test -p chrome` (dart2js) and `dart test -p chrome --compiler dart2wasm`:

- Top-level success message (boundary)
- Nested `Vec<i32>` (dartify recursion)
- Struct with mixed `int` + `String`
- Double pass-through — guards against accidentally narrowing `double` to `int`
- Bool
- Char via `String.fromCharCode`
- Primitive enum
- Optional null
- Optional present
- Negative int
- Complex enum with switch on discriminant — verifies `0 == 0.0` matches `case 0:`
- i64 via a real JS `BigInt(...)` construction — exercises `dcoDecodeI64` post-dartify
- `Uint8List` typed array

In CI, `test_dart_web` for `frb_dart` now runs this file under `--compiler dart2wasm` in addition to the existing dart2js run, and `test_flutter_web` matrixes over `wasm: [false, true]`, exercising the existing `flutter_via_create` (`greet → String`) and `gallery` integration tests under both compilers.

Verified end-to-end on a preview run on my fork (djx-y-z/flutter_rust_bridge#1):

- `Test :: Dart :: Web (frb_dart)` green — the new unit test passes under dart2wasm
- `Test :: Flutter :: Web (dart2wasm / frb_example--flutter_via_create)` green — real Rust call returning a String renders correctly in the browser under `--wasm`
- `Test :: Dart :: Native` green across Win/Mac/Linux × all packages — no native regression
- `Test :: Flutter :: Web (dart2js / *)` green — no dart2js regression
- `Test :: Rust` green across all variants
- `Generate :: FRB Codegen :: Command Generate` green for every package — the regen in this PR matches what CI produces

## Out of scope

- CST encoder path — the symmetric `.jsify()` direction may have related issues; haven't audited it here.
- `DartFn` / `DartFnFuture` under dart2wasm — DCO's DartFn decoder is `throw UnimplementedError`; DartFn flows through SSE, which already applies `maybeDartify`.
- `frb_dart/lib/src/wasm_module/_web.dart:24` loader incompatibility with wasm-bindgen 0.2.92 under dart2wasm. This is a separate infrastructure bug that will likely keep `Test :: Flutter :: Web (dart2wasm / frb_example--gallery)` red until addressed (gallery's mandelbrot uses `WorkerPool` which trips a different code path). The new `test_dart_web` dart2wasm run is the real runtime gate for this PR's DCO fix.

Addresses #2575. Real-world downstream report: djx-y-z/openmls_dart#5.


fzyzcjy:
Hi, could you please make the CI green (probably by using the agent skills)

djx-y-z:
@fzyzcjy quick update on CI — I pushed fixes for the failures you flagged:

- **`Command Integrate`** was template drift: the pre-generated `frb_generated.dart` in `integration_template/shared/` still had the old `raw as int` for `dco_decode_u_8`, so freshly-integrated examples diffed against it. Synced that one line to the codegen output.
- **`gallery` under dart2wasm** was hanging at startup (it hit the 6h timeout) on the WorkerPool / wasm-loader path (#2914) — unrelated to the DCO fix, and it runs before any FRB decode. I excluded that single matrix cell; `flutter_via_create × dart2wasm` + the new `dco_codec_test` (run under `--compiler dart2wasm`) stay as the runtime gate.

All of this PR's own checks are green now, including codecov.

The remaining red isn't from this PR: I merged latest `master` to confirm the fix is still consistent with current codegen (it is — `Command Generate` is green), and that surfaced that `master` currently doesn't compile. `frb_codegen/src/library/codegen/polisher/auto_upgrade.rs` has two `#[cfg(test)] mod tests` blocks (lines 118 and 193) → `error[E0428]: the name 'tests' is defined multiple times`. It appears to have come in with #3130, and master's own push run shows the same failure (run 26715398465 at `b0b4496` — `Test :: Rust` on all toolchains + `Generate :: Internal`). It's still present on current master (978bd28), so this PR just inherits those reds through the merge. Happy to re-merge once master builds again so everything goes green.

djx-y-z:
Thanks for picking this up in #3182! Looks like it covers the full dart2wasm surface — all the DCO decoders, the CST encode path, the WorkerPool fix (which is what made `gallery` hang under `--wasm` here), and the web CI matrix — and folds in the integer-decoder fix from this PR as `dcoDecodePrimitiveInt`.

Happy to close #3125 as superseded by #3182. If there's a specific piece you'd like me to keep or carry over instead, just let me know.

fzyzcjy:
(merge + revert this pr, s.t. the work is recognized by github + the dev work is in another pr)

PR #3065, opened 2026-05-03
Support HarmonyOS

star4277:
## Changes

Support HarmonyOS

## Changes

Fixes #2954

## Acknowledgements

Special thanks to the following projects for their work on bringing Rust support to the HarmonyOS / OpenHarmony ecosystem:

- [cargokit-ohos](https://github.com/tavimori/cargokit-ohos) by @tavimori
- [cargokit_ohos](https://github.com/Aloereed/cargokit_ohos) by @Aloereed
- [flutter_rust_bridge_ohos](https://github.com/Aloereed/flutter_rust_bridge_ohos) by @Aloereed

This contribution was made possible thanks to their exploration, implementation efforts, and open-source contributions to the HarmonyOS Rust ecosystem.

## Optional tools

* It is recommended to use the `/frb-develop-feature` skill, which contains detailed guidance for the agent to develop a feature.
* There is a devcontainer and Dockerfile at `.devcontainer`, which optionally helps creating a standardized development environment.

## PR chain

This PR is stacked on:

1. #3126
2. #3129
3. #3130

The coverage-only tests that exercise existing base-branch behavior were extracted to #3130 so this PR's diff stays focused on HarmonyOS support.

p.s. https://github.com/fzyzcjy/flutter_rust_bridge/pull/3065/commits/b54544f998e91d91e10ed24bb904ef823c5340e4 passed ci (but code is not cleaned up)

PR #3126, opened 2026-05-30
Support skipping FVM installation

fzyzcjy:
Authorship note: This extracts the skip_fvm_install / --skip-fvm-install functionality originally authored by @Star4277 in #3065 into a standalone PR. The goal is to review and land that FVM behavior independently from the HarmonyOS support work in #3065.

This PR only separates the FVM install-skip option and its generated command docs; it is not presented as original work by @fzyzcjy.


fzyzcjy:
LGTM

PR #3140, opened 2026-05-30
Timeout Flutter integration test commands

fzyzcjy:
## Summary

- Add timeout support to the internal Dart command runner.
- Apply a 20 minute timeout to Flutter integration test commands run by `frb_internal`.
- Kill the whole spawned process tree when a timeout fires, so shell wrappers do not leave child commands running.
- Cover the POSIX process-tree timeout path with a Dart test.

## Recovery Note

This PR was accidentally merged into the previous chain branch (`codex/web-test-browser-executable-env`) instead of `master`. The same change was recovered onto `master` via #3167.

## Verification

- `cd frb_dart && dart test test/run_command_test.dart`


fzyzcjy:
LGTM

PR #3181, opened 2026-06-06
Support native assets integration backend

fzyzcjy:
Support the native-assets integration backend for flutter_rust_bridge.

This PR is the final PR in a stacked chain. Please review/merge in order:

1. #3205 `codex/frb-integrator-split` -> `master`
   - Refactor integration integrator modules.
   - Intended as refactor-only prework.
2. #3206 `codex/frb-integration-template-layout` -> `master`
   - Restructure integration templates by backend.
3. #3208 `codex/frb-verify-apple-scaffold-source` -> `master`
   - Regenerate Apple scaffold on macOS and check the Apple scaffold source-of-truth paths.
4. #3212 `codex/frb-reduce-mobile-device-matrix` -> `codex/frb-verify-apple-scaffold-source`
   - Reduce duplicate Android/iOS device matrix entries.
5. This PR `codex/native-assets-backend` -> `codex/frb-reduce-mobile-device-matrix`
   - Add the native-assets integration backend.
   - Add `flutter_rust_bridge_hooks`.
   - Add checked-in native-assets examples and CI coverage for them.

Keeping these as a stack is intentional so this PR's diff focuses on the native-assets backend rather than mechanical integration refactors, Apple scaffold regeneration, and CI matrix cleanup.

## Native assets loading notes

- The build hook registers the generated FRB Dart library as the Native Assets code-asset id (for example `src/rust/frb_generated.io.dart`) and routes the Rust build output into the app bundle through `native_toolchain_rust` / `ToAppBundle()`.
- The Dart loader intentionally stays on the normal Flutter FFI loading paths instead of opening a `package:` asset URI. It opens `lib<stem>.so` on Android/OHOS, `<stem>.dll` or `lib<stem>.so` on desktop, and `<stem>.framework/<stem>` on Apple platforms, with the existing non-packaged `ioDirectory` path still serving Dart/non-packaged tests.
- The two reference repos use the same basic hook shape. `kushalmahapatro/frb_native_assets` uses `RustBuilder(assetName: ...frb_generated.io.dart, cratePath: ../rust)` and then calls the normal generated `RustLib.init()` path, with no generated-loader patch. `Yegair/flutter-rust-bridge-code-assets-sample` adds a temporary wrapper that reads `.dart_tool/hooks_runner/.../output.json` for desktop/`flutter test` and opens platform-specific names on iOS/Android; that wrapper compensates for trying to load a code asset by `package:` URI directly. This PR avoids that `package:`-URI path and relies on the bundled platform library names, which are covered by the checked-in native-assets examples and Flutter native CI matrix.

## Why this differs from the reference examples

- `kushalmahapatro/frb_native_assets` is already close to this PR: it uses `native_toolchain_rust.RustBuilder` and then calls the normal generated `RustLib.init()` path, without patching the generated loader.
- `Yegair/flutter-rust-bridge-code-assets-sample` adds a user-side `initBundled()` wrapper that reads `.dart_tool/hooks_runner/.../output.json` for desktop / `flutter test`. That wrapper is a conservative workaround for opening a Native Assets code-asset id such as `package:sample_flutter_lib/src/rust/frb_generated.io.dart` directly; `DynamicLibrary.open()` does not resolve that `package:` URI.
- This PR deliberately avoids opening the `package:` code-asset id. The build hook still registers the generated FRB Dart library as the asset id, but runtime loading uses the ordinary Flutter FFI platform names (`lib<stem>.so`, `<stem>.dll`, `<stem>.framework/<stem>`). That is why the Yegair-style `output.json` lookup is not needed here.
- This is not only theoretical: after `d99721638d` removed the hook-output fallback, CI run `27444477653` on commit `3d1e4488982d4981c16b4c9f93e5447cfff9cf3c` passed desktop `flutter test integration_test/simple_test.dart` for the native-assets examples on macOS, Windows, and Linux. The current head does not change the relevant loader, hook builder, or native-assets example tests from that passing commit.


fzyzcjy:
LGTM

PR #3182, opened 2026-06-06
Run dart2wasm in web CI

fzyzcjy:
## Summary

- Add dart2wasm variants to the Dart and Flutter Web CI matrices while retaining dart2js lanes.
- Keep coverage artifacts compiler-specific.
- Update CI-plan tests and snapshots for the expanded matrices.

## Stack

- This is the top PR in the native GitHub stack that starts at #3428.

## Validation


tools/frb_internal/lib/src/makefile_dart/test.dart
```diff
@@ -616,7 +626,9 @@ Future<void> flutterIntegrationTestRaw({
 Future<void> testFlutterWeb(TestFlutterWebConfig config) async {
   await _runFlutterDoctor();
   await runPubGetIfNotRunYet(config.package);
-  await _installDartCoverage();
+  if (config.coverage) {
+    await _installDartCoverage();
+  }
```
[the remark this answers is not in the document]
fzyzcjy: Addressed in later commits. At current head 6f6d11ef3c, `executeFrbCodegen` only receives `--dart-coverage` when `config.coverage` is true, matching the conditional `_installDartCoverage()` call. I also verified the relevant local Docker cases after the latest fix: `./frb_internal test-flutter-web --package frb_example--flutter_via_create --coverage --wasm` passed, and `./frb_internal test-dart-web --package frb_example--pure_dart` passed.

frb_dart/lib/src/manual_impl/_web.dart
```diff
@@ -1,10 +1,45 @@
...
+
+/// {@macro flutter_rust_bridge.only_for_generated_code}
+String dcoDecodeString(Object? raw) {
+  if (raw is String) return raw;
+  if (raw is JSAny) {
+    final dartified = raw.dartify();
+    if (dartified is String) return dartified;
+  }
+  return String.fromCharCode(dcoDecodePrimitiveInt(raw));
+}
```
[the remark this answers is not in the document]
fzyzcjy: Already present in the current head: `dcoDecodeString` checks `raw is JSString` and returns `raw.toDart` before falling back to `dartify()`.

tools/frb_internal/lib/src/makefile_dart/test.dart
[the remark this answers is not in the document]
fzyzcjy: Addressed in the current head: `testFlutterWeb` now only passes `--dart-coverage` when `config.coverage` is true, matching the conditional `_installDartCoverage()` call.

frb_dart/lib/src/cli/run_command.dart
```diff
@@ -31,20 +31,33 @@ Future<RunCommandOutput> runCommand(
...
   Duration? timeout,
 }) async {
+  final processEnvironment = removedParentEnvKeys.isEmpty
+      ? env
+      : {
+          for (final entry in Platform.environment.entries)
+            if (!removedParentEnvKeys.contains(entry.key))
+              entry.key: entry.value,
+          ...?env,
+        };
```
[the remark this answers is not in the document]
fzyzcjy: Already present in the current head: `removedParentEnvKeys` is converted to uppercase and compared against `entry.key.toUpperCase()`. This is covered by `runCommand removes selected parent environment keys case-insensitively`.

fzyzcjy:
I reviewed every `frb_codegen` change against the current `master`, specifically asking whether it is required for the dart2wasm support or can be dropped.

Conclusion: all 12 changed files are justified. I do not see a `frb_codegen` change that can be removed without either breaking a dart2wasm representation boundary, losing typed-array correctness, or leaving a checked-in generated template stale.

Per-file assessment:

- `assets/integration_template/shared/shared/lib/src/rust/frb_generated.dart` — necessary generated-output sync. The template must use `dcoDecodeString` and `dcoDecodePrimitiveInt` for the same JS/Dart representation differences handled by the generator. This is not a separate behavioral change and should remain generation-derived.
- `.../dart/.../codec/cst/encoder/ty/primitive_list.rs` — necessary. Signed and unsigned 64-bit lists need distinct BigInt typed arrays (`BigInt64Array` versus `BigUint64Array`); generic `jsify()` does not preserve the required representation.
- `.../dart/.../codec/dco/decoder/ty/boxed.rs` — necessary. Boxed 8/16/32-bit integers must delegate to the primitive decoder so integral JS numbers represented as Dart `double` under dart2wasm are normalized instead of failing a direct cast.
- `.../dart/.../codec/dco/decoder/ty/delegate.rs` — necessary. Arrays/lists, strings/backtraces, chars, primitive enums, JSON, anyhow messages, and big primitives all cross representation-sensitive boundaries. Routing them through `dcoDecodeList`, `dcoDecodeString`, or `dcoDecodePrimitiveInt` is the compatibility fix. The String and Backtrace arms could be stylistically combined, but neither conversion can be removed.
- `.../dart/.../codec/dco/decoder/ty/enumeration.rs` — necessary. Both the enum payload container and discriminant need normalization before indexing/decoding.
- `.../dart/.../codec/dco/decoder/ty/general_list.rs` — necessary. A JS array is not guaranteed to arrive as `List<dynamic>` until `dartify()` is applied by `dcoDecodeList`.
- `.../dart/.../codec/dco/decoder/ty/primitive.rs` — necessary. This is the central normalization for 8/16/32-bit integers that may arrive as integral doubles under dart2wasm.
- `.../dart/.../codec/dco/decoder/ty/record.rs` — necessary. Records use a list-shaped wire payload and need the same JS-array normalization before length checks and element decoding.
- `.../dart/.../codec/dco/decoder/ty/rust_opaque.rs` — necessary. Opaque metadata is list-shaped, so passing an unnormalized JS array to `frbInternalDcoDecode` is unsafe under dart2wasm.
- `.../dart/.../codec/dco/decoder/ty/structure.rs` — necessary for the same reason as records: normalize the JS array before field access and validation.
- `.../dart/.../misc/ty/dart_fn.rs` — necessary. Callback call IDs can arrive as an integral double, so the generated closure must accept `dynamic` and normalize at the native delivery boundary. Keeping `int` would fail before conversion can run.
- `.../rust/.../codec/cst/decoder/ty/primitive_list.rs` — necessary correctness fix. A BigInt typed-array view may have a non-zero byte offset; converting the whole backing buffer reads unrelated bytes. Preserving `byte_offset` and `byte_length` is required for subarray/view inputs. The assertion also checks the 8-byte element invariant.

These changes also form a coherent set: the helpers in `frb_dart` define the platform-specific normalization, while these generator changes ensure every generated representation-sensitive call site actually uses them. The generated-code tests plus the pure-Dart and PDE dart2wasm E2E runs cover the resulting paths.

tools/frb_internal/lib/src/makefile_dart/test.dart
```diff
@@ -759,4 +792,29 @@ Future<void> _runFlutterDoctor() async {
   await exec('flutter doctor -v');
 }
 
+Future<String> _flutterWebChromeBinaryArg() async {
+  final chromeBinary = Platform.environment['CHROME_EXECUTABLE'];
+  if (chromeBinary != null && chromeBinary.isNotEmpty) {
+    return '--chrome-binary=$chromeBinary ';
```
fzyzcjy: **[P2] Quote the configured Chrome executable path**

When `CHROME_EXECUTABLE` contains spaces, as in standard macOS or Windows Chrome paths, the raw value is interpolated into a command that `SimpleExecutor` runs through a shell. The shell splits it into multiple arguments, so `flutter drive` cannot launch Chrome. Escape it for the active shell or invoke Flutter with a structured argument list.

frb_utils/lib/src/dart_web_test_utils/static_content.dart
```diff
@@ -82,6 +98,26 @@ const kTestEntrypointHtmlContent = r'''
...
 </html>
 ''';
+
+const kJsEntrypointScript = r'''
+  <script src="main.dart.js" async></script>
+''';
+
+const kWasmEntrypointScript = r'''
+  <script type="module">
+    import { compile } from "./main.dart.mjs";
```
fzyzcjy: **[P2] Put the module import inside the try block**

When `main.dart.mjs` is missing, fails to parse, or does not export `compile`, ES module loading rejects this static import before the module body executes. The `try`/`catch` therefore never sends `false`, so the runner waits for its ten-minute timeout instead of failing fast. Load the module dynamically inside the protected block or install an equivalent module-error handler.

frb_dart/lib/src/cli/build_web/executor.dart
```diff
@@ -193,6 +254,7 @@ Future<void> _executeWasmPack(
       'RUSTFLAGS': rustflagsResolution.rustflags,
       if (stdout.supportsAnsiEscapes) 'CARGO_TERM_COLOR': 'always',
     },
+    removedParentEnvKeys: _cargoLlvmCovEnvKeys,
```
fzyzcjy: **[P2] Preserve non-coverage Rust compiler wrappers**

When a normal `build-web` invocation sets `RUSTC_WRAPPER`, for example to `sccache`, without running under cargo-llvm-cov, this unconditional removal still deletes the variable before `wasm-pack`. That silently disables the configured wrapper and can break wrapper-dependent builds. Only strip these keys when cargo-llvm-cov markers are present, or retain non-coverage wrappers.

fzyzcjy (commented):
Found three P2 correctness issues; see the inline comments.

frb_utils/lib/src/dart_web_test_utils/static_content.dart
```diff
@@ -82,6 +98,25 @@ const kTestEntrypointHtmlContent = r'''
...
+const kWasmEntrypointScript = r'''
+  <script type="module">
+    try {
+      const { compile } = await import("./main.dart.mjs");
+      const response = await fetch("./main.dart.wasm");
+      const app = await compile(await response.arrayBuffer());
+      const instance = await app.instantiate({});
+      instance.invokeMain();
+    } catch (error) {
+      console.error(error);
```
fzyzcjy: **[P2] Keep failure reporting independent of the log socket**

Moving the import into the `try` block is correct, but this path can still miss `sendTestResult(false)`. The log WebSocket is created without waiting for `open`, and this `console.error` calls `channel.send`; if the socket is still `CONNECTING`, that throws before the result is sent. `sendTestResult` also starts with the overridden `console.log`, which has the same failure mode. A fast missing or invalid module can therefore still reach the ten-minute timeout.

Please make log forwarding safe when the socket is not open, use the captured native logger (or no logger) inside `sendTestResult`, and ensure the result socket is opened and sent regardless of logging failures. Add a browser-level regression case with a missing or invalid `main.dart.mjs` that verifies the runner receives `false` promptly.

frb_dart/test/build_web_executor_test.dart
```diff
@@ -76,4 +80,326 @@ void main() {
...
+        wasmPack.arguments,
+        containsAllInOrder(['--out-name', 'demo_crate']),
+      );
+      expect(wasmPack.arguments, containsAll(['--features', 'demo_feature']));
+      expect(wasmPack.env, containsPair('RUSTUP_TOOLCHAIN', 'nightly'));
+      expect(
+        wasmPack.env,
+        containsPair('RUSTFLAGS', buildWebDefaultWasmPackRustflags),
+      );
+      expect(wasmPack.removedParentEnvKeys, contains('LLVM_PROFILE_FILE'));
```
fzyzcjy: **[P1] Cover both coverage states instead of keeping this assertion unconditional**

With the new condition in `_executeWasmPack`, a normal test process without `CARGO_LLVM_COV=1` correctly passes an empty `removedParentEnvKeys` list. This assertion is now stale: the focused Docker test fails here with `Expected: contains 'LLVM_PROFILE_FILE'` and `Actual: []`.

Please make the parent environment injectable (or extract the decision into a pure helper) and add deterministic tests for both states:

- No coverage marker with `RUSTC_WRAPPER=sccache`: retain the wrapper and remove no parent keys.
- `CARGO_LLVM_COV=1`: remove the cargo-llvm-cov keys, including its `RUSTC_WRAPPER`.

This both fixes the current test failure and protects the intended non-coverage behavior.

fzyzcjy (commented):
Two follow-ups remain after the latest fixes; the inline comments describe the concrete changes.

fzyzcjy (commented):
LGTM

fzyzcjy:
that force push is a rebase stack

PR #3236, opened 2026-06-14
docs: add HarmonyOS setup guide

star4277:
## Changes

docs: add HarmonyOS setup guide

## Optional tools

* It is recommended to use the `/frb-develop-feature` skill, which contains detailed guidance for the agent to develop a feature.
* There is a devcontainer and Dockerfile at `.devcontainer`, which optionally helps creating a standardized development environment.


fzyzcjy:
nit: would be great to be in `misc` folder

PR #3251, opened 2026-06-15
Restore sanitizer CI coverage

fzyzcjy:
## Summary

- Restore ASAN, MSAN, LSAN, and TSAN coverage for three normal Dart packages and deliberate-bad controls, using version-matched, checksum-verified Dart SDKs.
- Require normal entrypoints to exit zero and finish their test suites; require failing controls to emit the expected sanitizer diagnostic.
- Release serializer buffers when Dart input or callback-result encoding throws before ownership transfer, preserving the original allocation length and preventing double disposal.
- Defer the recursive CST disposed-input prevalidation fix at the maintainer's request. Preserve existing CST encoding behavior and explicitly suppress its known container leaks instead.
- Keep the three requested contributing documentation files unchanged relative to the PR base.

## Known CST leak exception

The [baseline LSAN job](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/33966816816/job/101308395334) at `ed0fe3a159` completed 4,386 tests, then reported 560 bytes in 35 allocations from generated CST container allocators. Encoding an already-disposed opaque inside a container can throw after allocation but before ownership transfers to Rust. This PR intentionally leaves that bug unfixed.

- `tools/dart_lsan_cst.supp` names the 18 observed allocator symbols with full-name anchors. It is enabled only for the `pure_dart` ASAN/LSAN entrypoint, including ASAN's leak detector.
- Suppression statistics must use known rules and remain within a total of 36 allocations and 576 bytes. Unknown or duplicated rules, malformed summaries, and exceeded budgets fail validation.
- Normal entrypoints still require exit zero and the test-success marker. Unrelated unsuppressed leaks remain failures. Deliberate-bad controls receive none of these suppressions.
- Valgrind retains the four previously recorded allocator aliases for the same CST leak. The obsolete serializer-buffer suppression remains removed.
- These rules identify allocation sites, not the precise disposed-input call path. A new leak at the same allocation site may be hidden within the existing budget; the Valgrind aliases do not have the LSAN count/byte budget.

## CST budget regression evidence

- Baseline `97c5b8c755`: [LSAN job 103503812368](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/34674602504/job/103503812368) passed all 4,386 tests and reported only the known 18 allocator sites, but the 35/560 budget rejected the complete 36-allocation, 576-byte workload.
- Each of six CST variants has an array and a vec disposed-input call; each allocates a backing buffer and a wrapper. The enum and nested calls add one box each: six allocations per variant, 36 total. The historical report counted one fewer RustAsync list allocation; its precise reachability cause is unproven.
- A fixture copied from the real CI summary fails against the old budget and passes after correction. Tests still reject 37 allocations, 577 bytes, unknown rules, duplicate rules, and unsuppressed failures.
- Docker internal tool tests passed (219); full lint passed. Independent correctness and test-weakening reviews found no actionable issues with the correction.

## Serializer reproduction evidence

- Baseline `03a5674889`: the temporary disposed-opaque reproducer throws as expected but fails the release assertion, expecting allocation length 8 and observing null.
- Command: `cd frb_example/dart_minimal && timeout 300 dart test test/rust_vec_dispose_test.dart` in the repository Docker environment.
- Related baseline `f5aaff6c59`: `RustVecU8.dispose()` passes length zero instead of the allocated length eight.
- Final coverage lives in `pure_dart/test/serializer_dispose_test.dart` and its PDE counterpart. The temporary minimal fixture has been removed.

## Scope and validation

- Stack: `master` → #3383 → #3419 → this PR. The independent Web receiver/StreamSink changes remain excluded, and #3441 remains closed. Existing Web flakes may need retries.
- Local Docker validation: codegen tests passed (665 library tests, 13 binary tests, and 1 doctest); 219 internal tool tests passed and cover suppression scope, exact rules, aggregate budgets, and rejection of unrelated failures.
- Full repository lint passed; a second owning code-generation run produced no changes.
- A real local LeakSanitizer probe failed for the known allocation without suppression, passed with suppression, and failed again when an unrelated allocation leaked alongside it.
- Full current-head sanitizer, Valgrind, browser, and macOS CI results are not implied by these local checks.
- Serializer cleanup releases its byte buffer; it does not promise arbitrary opaque-resource rollback. Existing TSAN exceptions remain package-scoped, limited to one known thread-creation report per entrypoint.

## SDK cache lock contention reproduction

- Baseline `8df01cc3eb45b8693d8a6b8cac666b868681a17f` uses `FileLock.exclusive` for the shared release cache.
- In the FRB Linux Docker environment, hold an append-opened `.lock` with `FileLock.exclusive` in one Dart process, then have a second Dart process open that same path and call `await lock.lock(FileLock.exclusive)`. The contender immediately throws `FileSystemException: lock failed` with `Resource temporarily unavailable, errno = 11` and exits 255.
- Repeat with `FileLock.blockingExclusive` in the contender: it stays pending while the first process holds the lock, then acquires it and exits zero after release. The probe bounds startup/acquisition to five seconds and the entire run to twenty seconds.
- Fix `845941f12d` changes only the cache lock mode to `blockingExclusive`, preserving the critical section and cleanup. This validates cross-process lock contention, not a complete concurrent SDK download. POSIX process-level locks do not serialize isolates within the same process.


fzyzcjy (commented):
LGTM

frb_example/pure_dart/test/dart_sanitizer_runtime_shutdown_io.dart
```diff
@@ -0,0 +1,56 @@
...
+      if (pendingDropCount() == 0) {
+        break;
+      }
+    }
+    if (DateTime.now().isAfter(deadline)) {
+      throw StateError('Timed out draining pending DartOpaque drops');
+    }
+    await Future<void>.delayed(Duration.zero);
+  }
+  drainLeakedVecs();
```
fzyzcjy: P0: Do not unconditionally drain these registries before LSan observes process exit. They contain every outstanding bridge Vec, Box, and persistent handle, so a regression where Dart/finalizers fail to return ownership is force-freed here and the sanitizer run can turn green. The deliberate_bad sentinel does not exercise these bridge registries. Please make non-empty outstanding registries fail (or limit cleanup to explicitly proven runtime-owned allocations) instead of sweeping them. Since this file is generated, the generator needs the corresponding fix too.

tools/frb_internal/lib/src/misc/dart_sanitizer_tester/sanitizer.dart
```diff
@@ -1,7 +1,20 @@
...
 extension SanitizerMetadata on Sanitizer {
-  String get rustflagValue {
+  String? get rustflags {
+    final value = rustflagValue;
+    if (value == null) return null;
+
+    final sanitizerFlag = '-Zsanitizer=$value';
+    if (this == Sanitizer.msan) return sanitizerFlag;
+
+    return '$sanitizerFlag --cfg frb_sanitize_runtime_shutdown';
```
fzyzcjy: P0: This cfg is also enabled for TSAN, but it injects global Mutex<HashMap/HashSet> bookkeeping into the production FFI pointer and DartOpaque paths. Those locks add happens-before edges and serialize ownership handoffs, which can hide races that exist without the test-only instrumentation. Please keep TSAN on production synchronization semantics, or use teardown tracking that does not introduce synchronization into the tested hot paths.

frb_rust/src/for_generated/pointer.rs
```diff
@@ -33,21 +90,73 @@ pub fn from_vec_to_array<T, const N: usize>(v: Vec<T>) -> [T; N] {
 // ref: doc of [Box::into_raw]
 pub fn new_leak_box_ptr<T>(t: T) -> *mut T {
     let x: Box<T> = Box::new(t);
+    #[cfg(frb_sanitize_runtime_shutdown)]
+    {
+        let ptr = Box::into_raw(x);
+        let previous = LEAKED_BOXES.lock().unwrap().insert(
```
fzyzcjy: Sanitizer-only correctness regression: Box::into_raw for zero-sized types may return the same dangling address for multiple simultaneously live boxes. Keying this registry only by address makes the second insert replace the first and trip previous.is_none(). Please cover multiple outstanding Box<ZST> values and use tracking that supports duplicate raw addresses.

fzyzcjy (commented):
Independent chain audit result: changes are required. The sanitizer matrix and previously skipped opaque tests are stronger, but the inline findings below weaken sanitizer fidelity or introduce sanitizer-only failures compared with main.

fzyzcjy:
Test/sanitizer weakening audit of the full chain (`master` → #3383 → #3251): **FAIL — one weakening remains.**

- #3383 is a mechanical split of the sanitizer tester/metadata; I found no lost checks or reduced coverage there.
- The cumulative chain expands the sanitizer matrix from ASAN/LSAN normal lanes to ASAN/MSAN/LSAN/TSAN normal lanes plus deliberate-bad sentinels, and otherwise does not delete tests, add skips, shrink inputs, or weaken existing assertions.
- Blocker: `tools/frb_internal/lib/src/misc/dart_sanitizer_tester/runner.dart:317-354`. `isOnlyAllowedLeakForTesting` accepts any non-zero exit when stderr contains exactly one `SUMMARY:` line equal to the allowlist. It does not reject additional non-`SUMMARY:` fatal/runtime/sanitizer diagnostics. Therefore an exact known leak summary plus an unrelated teardown exception, fatal diagnostic, or other failure can still make the job pass. This is broader than suppressing only the known leak and is weaker than the base behavior, which required exit code 0 for normal lanes.
- The new unit test only rejects a different summary or multiple summaries; it does not cover an allowlisted summary combined with an unrelated fatal diagnostic.

Verdict: I cannot confirm “no test weakening” on the current head `23b72e6c`. The suppression needs to validate the complete allowed failure shape (and have a negative regression case for allowed summary + extra fatal/error output) before this gate passes. This audit intentionally does not require fixing the underlying sanitizer/runtime bug.

fzyzcjy:
Re-audit of the updated full chain (`master` → #3383 → #3251), superseding my previous FAIL comment: **PASS — no test or sanitizer weakening found at head `76affc2e`.**

- #3383 remains a mechanical split of the sanitizer tester/metadata; no checks or coverage were lost.
- The cumulative chain preserves the original ASAN/LSAN normal lanes and expands coverage to ASAN/MSAN/LSAN/TSAN across the three normal packages, plus a deliberate-bad sentinel for each sanitizer.
- I found no deleted tests, new skips/ignores, reduced inputs, relaxed assertions, smaller CI matrix, or `continue-on-error`-style bypass.
- The previous blocker is fixed. The known-failure suppression now requires the sanitizer-specific exit code, exactly one matching `SUMMARY:`, exactly one matching fatal diagnostic header, and the expected trailing report shape. Extra sanitizer reports, `DEADLYSIGNAL`, Dart unhandled exceptions, fatal/runtime diagnostics, wrong exit codes, or wrong trailing output are rejected. Normal package runs must also emit the explicit test-success marker.
- The ASAN Dart-only UAF sentinel changed from expected success to expected failure, which strengthens detection.
- Focused validation: `cd tools/frb_internal && dart test test/makefile_dart_test.dart` passed 70/70 in the FRB Docker environment.

Verdict for the requested gate: the updated PR chain does not weaken tests or sanitizer coverage relative to main. This verdict intentionally does not require fixing the underlying known sanitizer/runtime leaks; the remaining suppressions are narrowly shaped.

fzyzcjy:
The CST prevalidation in `frb_codegen/src/library/codegen/generator/wire/dart/spec_generator/codec/cst/validate.rs` addresses an observed native-memory leak.

- Before the fix, the [LSAN job at `ed0fe3a159`](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/33966816816/job/101308395334) completed 4,386 tests successfully, then reported **560 bytes leaked in 35 allocations**. Allocation stacks point to generated CST list, enum, and nested-opaque container allocators.
- The old encoder allocates a native container before encoding its contents. If a contained opaque handle is already disposed, encoding throws before the FFI call transfers ownership to Rust, leaving the container allocation without cleanup. The same ordering problem can occur across arguments when an earlier argument allocates before a later argument throws.
- Prevalidation traverses opaque-bearing inputs before encoding any argument, rejecting already-disposed handles before those allocations occur.
- The [subsequent LSAN job at `222bc594f1`](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/33969751235/job/101317208120) completed the same 4,386 tests and passed without the leak report. These are historical CI logs that were inspected for this explanation, not a fresh sanitizer run on the rebased head.

The current implementation has a cost and a limited guarantee:

- Containers containing opaque values get an additional traversal; getters may be read again during encoding. No performance measurement is claimed here.
- It rejects already-disposed inputs; it does not provide general rollback for throwing getters or repeated moves of the same owned handle.

This implementation can be replaced with another correct cleanup strategy, or extracted into a separate PR to reduce scope. Simply removing it would restore the known leak path; retrying CI would not fix that. My recommendation is to retain the fix or split it out, rather than drop it as unnecessary defensive code.

tools/frb_internal/lib/src/misc/dart_sanitizer_tester/dart_sdk.dart
```diff
@@ -8,55 +8,249 @@ import 'package:flutter_rust_bridge_internal/src/makefile_dart/test.dart';
...
+  final releaseName = sanitizedDartReleaseName();
+  final pathCacheRoot = path.join(
+    Directory.systemTemp.path,
+    'frb_sanitized_dart',
+    releaseName,
+  );
+  await Directory(pathCacheRoot).create(recursive: true);
+  final lock = await File(path.join(pathCacheRoot, '.lock'))
+      .open(mode: FileMode.append);
+  await lock.lock(FileLock.exclusive);
```
fzyzcjy: **[P2] Wait for the shared SDK cache lock instead of failing concurrent runs**

- **Trigger**: Two sanitizer commands running concurrently on the same machine or in the same container contend for the same release directory's `.lock`, even when they select different sanitizers.
- **Impact**: `FileLock.exclusive` is non-blocking: contention immediately throws a `FileSystemException`; `await` does not wait for the lock to become available. Since this lock is held throughout download, checksum verification, and extraction, the second invocation fails before its tests begin.
- **Fix**: Use `FileLock.blockingExclusive` or bounded retries to handle normal lock contention.

fzyzcjy (commented):
The shared SDK cache lock can cause concurrent sanitizer runs to fail before testing begins.

- **Scope**: Static review of the local PR branch against its stacked base.
- **Validation**: Tests were not run and live CI was not verified; network and Docker access were restricted during the original review.

fzyzcjy (commented):
The cumulative changes preserve or strengthen existing assertions and the CI matrix, but the three inline findings weaken existing leak-detection failure conditions. They do not satisfy the requirement that checks must remain unchanged or become stronger.

- Scope: local `master` (`71b50c786c`) → #3383 → #3419 → #3251 (`8df01cc3eb`).
- Publication check: the GitHub PR chain and #3251 head were confirmed before submitting this review.
- Validation: static review only; tests were not rerun.

tools/dart_valgrind.supp
```diff
@@ -16,10 +16,30 @@
    obj:*dart_valgrind_test_entrypoint
 }
 {
-   frb_sse_serializer_error_rust_vec_u8_leak
+   pure_dart_speed_twin_box_process_exit_leak
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:malloc
+   fun:frbgen_frb_example_pure_dart_cst_new_box_speed_twin_sync
```
fzyzcjy, line 23: [P1] Restore Valgrind leak detection for CST allocators

The existing `pure_dart` Valgrind test loads these four new suppressions. Each matches only `malloc`/`calloc`, an allocator symbol, and a leak kind, so new leaks matching the corresponding definite/indirect kind can pass without any caller-path, allocation-count, or byte-count limit.

- The two rules on the reviewed `master` baseline do not suppress these allocator stacks.
- Removing the RustVec suppression strengthens a different check and does not offset this regression under the requirement that checks only remain unchanged or become stronger.
- Please fix the leaks and remove these allocator suppressions.

tools/frb_internal/lib/src/misc/dart_sanitizer_tester/runner.dart
```diff
@@ -30,44 +28,171 @@ Future<void> run(TestDartSanitizerConfig config) async {
...
+  for (final row in tables.single.group(1)!.trim().split('\n')) {
+    final fields = row.trim().split(RegExp(r'\s+'));
+    if (fields.length != 3 || !rules.remove(fields[2])) {
+      throw Exception('Unexpected LSAN suppression rule');
+    }
+    count += int.parse(fields[0]);
+    bytes += int.parse(fields[1]);
+  }
+  if (count > 36 || bytes > 576) {
+    throw Exception('Known CST leak suppression exceeded its budget');
```
fzyzcjy, line 117: [P1] Do not use a new LSAN budget to accept previously failing leaks

The `pure_dart` ASAN/LSAN entrypoint adds 18 allocator suppressions, and this check accepts up to 36 suppressed allocations totaling 576 bytes. These leaks were not exempted on the reviewed `master` baseline.

- The aggregate budget also lets allocator counts compensate for one another: the fixture contains only 4 `HideDataTwinMoi` allocations, while the new unit test explicitly accepts all 36 allocations from that single allocator.
- Keeping the nonzero-exit-code check does not preserve detection, because suppression can make the sanitizer exit successfully before this check runs.
- Please fix cleanup when encoding fails and remove the new leak exemptions and budget to preserve the existing failure conditions.

tools/frb_internal/lib/src/misc/dart_sanitizer_tester/runner.dart
```diff
@@ -30,44 +28,171 @@ Future<void> run(TestDartSanitizerConfig config) async {
...
+  if (suppressionFile == null || sanitizer != Sanitizer.tsan) {
+    return {};
+  }
+  final existing = environment['TSAN_OPTIONS'];
+  return {
+    'TSAN_OPTIONS': [
+      if (existing != null && existing.isNotEmpty) existing,
+      'report_thread_leaks=1',
+      'print_suppressions=1',
+      'suppressions=../../tools/$suppressionFile',
```
fzyzcjy, line 86: [P2] Preserve thread-leak failures in the existing TSAN command

Running the existing `test-dart-sanitizer --sanitizer tsan` command for `pure_dart` or `pure_dart_pde` now suppresses thread leaks whose creation stack contains `simple_use_async_spawn_blocking::`, and the summary checker accepts one match. The reviewed `master` command has no such exemption, so an unreclaimed thread on this path no longer fails the check.

- `report_thread_leaks=1` does not preserve the original failure condition when the report is suppressed.
- Adding a TSAN CI matrix strengthens coverage elsewhere but does not offset weakening an existing command.
- Please fix thread cleanup and remove this exemption.

fzyzcjy (commented):
LGTM

fzyzcjy:
Clarification of the CST leak exceptions, why main is green, and the decision to keep the current approach.

### Why main passes without these CST exceptions

- Main (`71b50c786c`) really ran and passed [pure_dart Valgrind](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/34674304152/job/103503950641) and [pure_dart LSAN](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/34674304152/job/103503950800). They were not skipped.
- This is not simply an unavoidable SDK-version difference. Main's Valgrind job already uses Dart 3.13.0. This PR also changes the test entrypoint from forced `exit(...)` to `RustLib.dispose()` followed by normal VM shutdown.
- I isolated the shutdown behavior using one Dart 3.13 AOT executable, one Rust library, the same complete test suite, and no CST suppressions. Both variants passed 4,386 tests with the same six existing skips.

| Shutdown behavior | Valgrind | Fixed-runtime LSAN |
| --- | --- | --- |
| `RustLib.dispose()` followed by forced exit | Zero errors, exit zero; the 576 bytes in 36 CST allocations remain **still reachable** | Exit zero, no leak report |
| `RustLib.dispose()` followed by normal shutdown | Four CST leak contexts: 384 bytes directly lost plus 192 bytes indirectly lost | Reports 576 bytes leaked in 36 allocations |

- Normal shutdown removes the reachability that previously kept these unfreed native allocations classified as reachable. Calling `RustLib.dispose()` alone is not sufficient to trigger the difference: the forced-exit control also calls it.
- Allocation addresses were checked: different allocator names selected by Valgrind and LSAN are aliases at the same addresses in the optimized library. This attribution is not based only on matching byte totals.
- Experiment limits: these were local Linux ARM64 runs. LSAN used a fixed system GCC runtime injected into the AOT process, not CI's x86_64 instrumented Dart SDK. Both final Valgrind variants used the same `--max-stackframe=16777216` setting to avoid ARM Dart stack-switching false positives; neither used CST suppressions. These experiments establish the shutdown effect, not current-head full CI approval.

### Current implementation and accepted tradeoff

The maintainer has selected **normal shutdown + the current known-CST-leak exceptions**, deferring the recursive CST prevalidation fix from this PR.

- Keep normal shutdown: it exercises more complete teardown than main's forced exit.
- ASAN/LSAN: enable 18 fully anchored allocator rules only for the `pure_dart` ASAN/LSAN entrypoint. Validate aggregate suppression statistics against a maximum of **36 allocations / 576 bytes**. The budget is shared across allocator rules; unknown or duplicate rules, malformed summaries, and exceeded budgets fail.
- Valgrind: retain four CST allocation-site rules. These have **no allocation-count or byte budget**.
- Normal test entrypoints must still finish successfully and exit zero. Unrelated unsuppressed leaks remain failures. The obsolete serializer-buffer suppression stays removed; the serializer cleanup fix remains included.
- These are allocation-site exceptions, not exact disposed-input-path exceptions. A different leak at a covered site may be hidden; for ASAN/LSAN it must fit within the shared budget, whereas Valgrind's four exceptions are unbounded.

### Correction to the earlier framing

- It was too strong to say that matching main's detection strength necessarily requires fixing CST or accepting failing CI. Restoring main's forced-exit behavior is another route, and the local controls demonstrate why it can pass without CST exceptions.
- However, that would give up the additional normal-shutdown coverage. The chosen approach retains that coverage while explicitly accepting the exceptions above.
- This is an accepted scoped compromise, **not a claim that every aspect of detection is at least as strong as main**. Additional shutdown coverage does not cancel out the possibility that an allocator exception hides another leak main could have detected.
- No further code change is requested for this tradeoff. The current implementation at `845941f12da98df230f1693026d043765c55e0f8` already follows it. Final CI status still needs to be evaluated separately.


### CI snapshot at this head (2026-09-12)

- In [the full CI run](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/34678211221), all **16 Dart sanitizer jobs** and all **three Valgrind jobs** have passed with the current exceptions and normal shutdown.
- The full workflow is not green yet: macOS jobs are still queued, and [pure_dart dart2wasm](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/34678211221/job/103513026249) failed one stream test. Its displayed name is `func_stream_add_value_and_error_twin_normal`: the expected two data events followed by an error were reduced to the error alone; the job reported 4,177 passed, 70 skipped, and one failed.
- This resembles the previously discussed Web stream delivery problem. A same-head retry is needed before classifying this occurrence as a flake; sanitizer success does not establish that conclusion.
- GitHub rejected rerunning the individual job while the full workflow remains unfinished. A [focused same-SHA dart2wasm retry](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/34681341148) has been dispatched and is currently queued. It is additional evidence, not a replacement for full CI completion.
- The maintainer accepts retrying existing Web flakes. The independent Web stream fix remains outside this PR; #3441 stays closed.

**Decision:** keep the current code for the shutdown/suppression tradeoff. The known CST leak remains deferred and explicitly suppressed; it is not fixed by this PR. CI completion is a separate outstanding check.

fzyzcjy:
CI update for `845941f12da98df230f1693026d043765c55e0f8`:

- The [same-SHA pure_dart dart2wasm retry](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/34681341148/job/103521470868) passed: **4,178 tests passed, 70 skipped**, with no source change. The focused workflow also completed successfully. This provides evidence that the earlier stream failure is intermittent.
- The earlier failing display name `func_stream_add_value_and_error_twin_normal` is reused across generated variants. Log ordering identifies the actual failing call as `funcStreamAddValueAndErrorTwinRustAsync()` in `test/api/pseudo_manual/stream_twin_rust_async_test.dart`; the ordinary TwinNormal variant passed in that same original run.
- The [original full workflow](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/34678211221) still has macOS/iOS jobs running or queued. Its original failed Web job has not yet been rerun successfully, so this update does **not** claim that full CI is green.
- The chosen normal-shutdown behavior and current sanitizer exceptions remain unchanged.

fzyzcjy:
image

PR #3270, opened 2026-06-26
Support pre-generation cleanup

### LINKED ISSUE #3263: Support Cleanup of Stale Generated Files in flutter_rust_bridge_codegen

When using `flutter_rust_bridge_codegen generate`, generated output files that are no longer part of the API graph are not removed automatically.

A common example is when a Rust module is removed from `rust_ffi/src/api/mod.rs`.

Before:

```rust
pub mod user_api;
pub mod settings_api;
pub mod device_api;
```

This generates:

```text
lib/native_files/rust_ffi/
├── user_api.dart
├── settings_api.dart
├── device_api.dart
```

Suppose `device_api` is later removed:

```rust
pub mod user_api;
pub mod settings_api;
```

Running:

```bash
flutter_rust_bridge_codegen generate
```

correctly stops generating `device_api.dart`, but the previously generated file remains on disk:

```text
lib/native_files/rust_ffi/
├── user_api.dart
├── settings_api.dart
├── device_api.dart   <-- stale file remains
```

This stale file can still be imported by Flutter code, causing confusion and potentially hiding API removal mistakes.

Currently, I work around this by manually deleting the generated output directory before every code generation run:

```bash
RUST_DIR="$project_root/lib/native_files/rust_ffi"

echo "Cleaning generated outputs..."

if [ -d "$RUST_DIR" ]; then
    rm -rf "$RUST_DIR"
    echo "Removed: $RUST_DIR"
fi

flutter_rust_bridge_codegen generate
```

I would like FRB to provide a built-in mechanism for cleaning stale generated files.

One option would be an optional configuration field:

```yaml
pre_generation_cleanup: true
```

Default:

```yaml
pre_generation_cleanup: false
```

When enabled, FRB would remove generated output directories before generation begins.

Another option would be automatic stale file detection. The generator could track the complete generation graph and remove outputs that are no longer produced.

For example:

1. Build the generation graph from the current Rust API surface.
2. Determine the set of files that should exist after generation.
3. Compare that set against the existing generated output directory.
4. Remove generated files that are no longer present in the graph.

Pseudo-flow:

```text
existing generated files
          │
          ▼
current generation graph
          │
          ▼
expected output files
          │
          ▼
delete(existing - expected)
          │
          ▼
generate current outputs
```

The current workaround is maintaining a separate cleanup script and ensuring it is run before every FRB generation step.

While this works, it requires additional tooling outside FRB and can easily be forgotten by contributors, resulting in stale generated artifacts remaining in the repository or local workspace.

**Additional context**

The issue becomes more noticeable during API refactoring, module renames, module removals, or when splitting and reorganizing generated bindings. In these situations, FRB correctly generates the new outputs but does not remove obsolete ones that were generated during previous runs.

fzyzcjy:
## Summary
- add an opt-in pre_generation_cleanup codegen setting and CLI flag
- clean stale Dart, Rust, and C generated output paths before generation when enabled
- document the setting and cover parser/internal cleanup behavior with tests

## Tests
- cargo fmt --manifest-path frb_codegen/Cargo.toml
- ./frb_internal test-rust-package --package frb_codegen
- ./frb_internal generate-internal-book-help
- ./frb_internal generate-internal-rust
- ./frb_internal precommit-generate
- ./frb_internal lint --fix

Close #3263


ganeshrvel:
@fzyzcjy 

I completely agree with `@gemini-code-assist` here. Since this feature performs destructive operations, I think we should be extremely cautious. An incorrectly configured `dart_output` could end up deleting a user's actual source code, which would be a worst-case scenario.

To make this as safe as possible, I think we should add a few intentional layers of protection before allowing cleanup to run.

**1. Double Opt-In (CLI + Config)**

Require both a dedicated CLI flag and the configuration option.

For example:

```yaml
pre_generation_cleanup: true
```

would not be sufficient by itself. The user would also have to invoke code generation with something like:

```bash
flutter_rust_bridge_codegen generate --enable-pre-generation-cleanup
```

Cleanup would only run if both are present. This prevents someone from accidentally enabling destructive behavior simply by pulling a configuration change.

**2. Marker File Verification**

I also think we should require a marker file that explicitly declares a directory as managed by FRB.

For example, if:

```yaml
dart_output: lib/native/saya_ffi
```

then the parent directory could contain a marker file such as:

```text
lib/native/.frb_cleanup.toml
```

whose contents explicitly specify the managed output directory:

```toml
dart_output = "lib/native/saya_ffi"
```

Before performing cleanup, FRB would:

1. Check that the marker file exists.
2. Parse the marker file.
3. Verify that the path stored in the marker file exactly matches the `dart_output` configured in `flutter_rust_bridge.yaml`.

If any of these checks fail, cleanup is aborted with a clear error message.

This gives users one more opportunity to explicitly acknowledge that a particular directory is managed by FRB and may be recursively cleaned. It also makes accidental misconfiguration much less likely to result in deleting unrelated project files.

I realize this introduces a little more setup, but given that this feature recursively deletes directories, I think it's worth favoring safety over convenience.

fzyzcjy:
by agent:

Thanks, agreed on making the recursive Dart cleanup recoverable. I updated the implementation so existing Dart output contents are moved to `<dart_root>/.dart_tool/flutter_rust_bridge/pre_generation_cleanup/<timestamp>-...` and the backup path is logged, while retaining the unsafe path guards. I did not add a double opt-in/marker-file gate in this PR because the backup plus path guards avoids irreversible recursive deletion while keeping the cleanup usable.

ganeshrvel:
Is there any rotation for the accumulated recycled backup directory? Having 100+ backups wouldn’t look very good.

Maybe let users configure the recycled deleted files retention or rotation max count value in the configuration file.

PR #3271, opened 2026-06-26
Fail fast for locked Rust opaque sync calls on web

### LINKED ISSUE #3266: Sync call while async is running might cause freeze on wasm
First, thank you, great work! 

There is a deadlock when calling a `#[fr(sync)]` function that mutates data while an async call is running. This happens only on wasm. I am not sure if this should be considered a bug or a limitation of wasm. If this cannot be fixed, I suggest updating the documentation.

## How to reproduce 

I have setup a [minimal project](https://github.com/Julien5/frb-sync-async-wasm/) to reproduce the freeze.

Version: 
- flutter_rust_bridge_codegen 2.12.0
- flutter 3.44.1

### Build 

Run:
```
./scripts/build-web.sh
```
- unpack /tmp/webapp.tgz and start the webserver to serve `<path>/app/build/web/`

### Reproduce 

- Open Firefox (tested with firefox 152.0).
- Click on `realAsyncConst` button, which calls:
```
    pub async fn realAsyncConst(&self) {
        log::trace!("[1]");
        sleep(std::time::Duration::from_millis(1000)).await;
        log::trace!("[2]");
    }
```
- and quickly after on the `realSyncMut` button, which calls
```
    #[frb(sync)]
    pub fn realSyncMut(&mut self) -> String {
        self.count = 1;
        format!("count={}", self.count)
    }
```
The application freezes. 
  
Other button combinations do not freeze the application.

With firefox, the application freezes. With chrome, the application does not freeze, it reports the `OnAtomics.wait cannot be called in this context` error.

## Hypothesis

After reading [this](https://github.com/fzyzcjy/flutter_rust_bridge/issues/1910) and [this](https://github.com/fzyzcjy/flutter_rust_bridge/issues/1917) related issues, I am not sure if this should be considered a bug or a limitation of wasm. I suggest updating the [documentation](https://cjycode.com/flutter_rust_bridge/manual/miscellaneous/wasm-limitations) accordingly, like:

    Calling a sync function while an async function is running might cause a deadlock. 
    To prevent this, avoid synchronous functions on `&mut` with `#[frb(sync)]`. 
    Use the frb async default instead, even when the body is non-async, for example:
    ```
     pub async fn pseudoSyncMut(&mut self) -> String {
		 // write data 
         self.count = 1;
         format!("count={}", self.count)
     }
    ```	
    On the dart side, the caller might have to `await` the result.

fzyzcjy:
Fixes #3266.

## Summary

- Make browser event-loop Web synchronous Rust opaque object lock acquisition fail fast when the object is already locked, instead of blocking the event loop.
- Keep native/non-Web behavior unchanged, and keep Web dedicated-worker contexts on the existing blocking lock path.
- Cover both `RustOpaque` and `RustAutoOpaque` for three shapes: browser main-thread sync contention fails fast, worker-pool sync contention can wait, and async contention can wait.
- Document the Web limitation and the intended workaround: do not use a synchronous API when it may need to wait for an async Rust opaque borrow; use an async API instead.
- Keep the existing pure_dart generator marker handling unchanged from upstream; the new regression APIs live in dedicated `rust_opaque_web_locking` files whose file-level generator comments request the needed SSE/PDE variants without creating the sync/async duplicate churn.

## What Was Broken

The failing pattern is:

1. Dart starts an async Rust call that takes a mutable borrow of a Rust opaque object.
2. Before that async call completes, Dart calls a `#[frb(sync)]` Rust API that also needs the same object lock.
3. The generated sync decode path tries to synchronously acquire the generated lock.

Before this PR, that last step used a blocking wait for every Web sync lock path. If the async call can only resume after the browser event loop makes progress, the sync call blocks the same event loop that is needed to release the lock. Firefox can freeze, while Chrome may report an `Atomics.wait cannot be called in this context`-style error depending on the exact path and browser behavior.

The bug is not that Rust mutual exclusion exists. The bug is that a browser event-loop Web synchronous call tried to wait for a lock whose release depends on browser event-loop progress.

## Why Non-Web Is Usually Fine But Web Is Not

On native/non-Web targets, blocking a synchronous caller on a lock is a normal operation. The blocked caller and the async task that eventually releases the lock can be scheduled by an OS/runtime with real threads available, so waiting does not inherently stop the whole process from making progress.

Web/Wasm is different for this sync-call path. The Dart/JS/Wasm synchronous call is running on the browser event-loop thread. If it enters a blocking wait, it prevents the event loop from running the continuation that would let the earlier async Rust call finish and release the mutable borrow.

So the Web failure is a self-dependency:

```text
sync Rust call waits for the Rust opaque object lock
  lock release requires the earlier async Rust call to resume
    async resume requires browser event-loop progress
      event-loop progress is blocked by the sync Rust call
```

That is why the same API combination can appear merely blocking or recoverable on non-Web, but freeze the page when the Web sync call is running on the browser event loop. Web dedicated-worker contexts are different: when the current global scope is a dedicated worker, this PR preserves the previous blocking lock behavior.

## Fix

For Web browser event-loop calls, synchronous Rust opaque object lock acquisition now uses non-blocking `try_read` / `try_write` and throws a JS exception with an explanatory message when the lock is already held.

Native/non-Web targets still use the previous blocking lock behavior. Web dedicated-worker contexts also keep blocking behavior, since blocking a worker does not block the browser event loop that drives the main-page async continuation.

This is intentionally a fail-fast fix rather than a serialization fix for the browser event-loop path. A synchronous Web call on that path cannot safely wait for an async borrow that needs the browser event loop to release the lock. Returning an immediate error preserves page responsiveness and tells the caller to use an async API for that operation.

### #3266

#3266 reports the exact bug fixed here: start an async call, quickly call a sync mutable method on the same Rust-owned object, and observe that Web freezes. The report also notes that other button combinations do not freeze and suggests documenting the limitation if it cannot be fixed.

My conclusion agrees with the issue's diagnosis in substance: this is a Web-specific sync/async locking problem, not a general native problem. I also agree that users should avoid `#[frb(sync)]` for operations that may contend with async mutable borrows on the browser event-loop path.

The refinement in this PR is that FRB cannot make a browser event-loop synchronous Web call safely block until the async borrow completes, but it can avoid freezing the browser by detecting the locked state and failing immediately. Dedicated-worker Web contexts still use the normal blocking path, so legitimate worker-side blocking waits are not removed.

### Prior analysis mentioned from #3266

The prior analyses mentioned around this issue were directionally right in the important way: Web main-thread code must not execute a path that may block on WebAssembly atomics or equivalent synchronization, because the browser event loop itself is needed to make progress.

The safe conclusion is narrower than "Web can never use synchronized Rust opaque data." Async FRB APIs can still be valid when they do not require a synchronous main-thread wait, and dedicated workers can still use the blocking path. The problematic operation fixed here is specifically browser event-loop synchronous Web lock acquisition that would need to wait for an already-held Rust opaque object lock.

## Regression Coverage

The new regression coverage checks both `RustOpaque` and `RustAutoOpaque`:

1. browser main-thread Web sync call while an async borrow is in flight fails fast with a Dart-visible error instead of freezing,
2. two worker-pool sync calls can contend for the same object and complete normally,
3. two async calls can contend for the same object and complete normally.

The coverage exists in both pure_dart and pure_dart_pde generated variants.

## Verification

After restoring `tools/frb_internal/lib/src/frb_example_pure_dart_generator/utils/generator_utils.dart` to match upstream and regenerating:

- `./frb_internal generate-internal-frb-example-pure-dart`
- `cd frb_example/pure_dart && cargo run --manifest-path ../../frb_codegen/Cargo.toml -- generate`
- `cd frb_example/pure_dart_pde && cargo run --manifest-path ../../frb_codegen/Cargo.toml -- generate`
- `cd frb_example/pure_dart && timeout 300 dart run test test/api/rust_opaque_web_locking_test.dart`
- `cd frb_example/pure_dart_pde && timeout 300 dart run test test/api/rust_opaque_web_locking_test.dart`
- `./frb_internal test-dart-web --package frb_example/pure_dart`
- `./frb_internal test-dart-web --package frb_example/pure_dart_pde`

Earlier broader local verification on this branch also included native pure_dart/PDE runs, `frb_rust` Rust tests, Dart lint, and generation checks. The broad native pure_dart run passed its test body but previously hung during an unrelated VM-service teardown, so the final post-merge validation above focuses on this fix plus the full Web suites that exercise the regression.

## Review

- Correctness subagent review: one PDE generated-output finding found and fixed in follow-up commit `3f31328005`; re-review found no unresolved actionable findings.
- Test-weakening subagent review: no actionable findings.
- Gemini review: actionable feedback about a too-weak test helper was addressed by removing the helper and covering the real sync Dart -> Rust path instead.


fzyzcjy:
Investigated the latest Gemini suggestion about stripping duplicator tags in `handleDisabledPdeContent`. I tried the literal change and regenerated pure_dart/PDE, but it changes the generation order semantics: PDE copy happens before duplicator, so removing duplicator markers in `handleDisabledPdeContent` makes the later duplicator create extra `event_listener_twin_rust_async` APIs/tests while the generated Dart bindings do not include the matching method. `./frb_internal test-dart-native --package frb_example/pure_dart_pde` then fails with `Method not found: createEventSyncTwinRustAsync`. So I reverted that experiment and am leaving the current marker behavior unchanged.

fzyzcjy:
Follow-up on the existing `rust_auto_opaque*_test.dart` invalid-borrow cases: I initially skipped the same-object borrow+mut-borrow cases on Web too broadly, but that was unnecessary. Those tests already assert failure, and the new Web fail-fast behavior still satisfies that intent. I restored the existing `rust_auto_opaque` / generated twin tests so they have no final PR diff, then reran both Web suites successfully:

- `./frb_internal test-dart-web --package frb_example/pure_dart`
- `./frb_internal test-dart-web --package frb_example/pure_dart_pde`

The new `rust_opaque_web_locking_test.dart` files remain responsible for asserting the specific locked-object fail-fast error and the worker-pool/async waiting cases.

fzyzcjy (commented):
LGTM

PR #3279, opened 2026-06-27
Ignore generated BigInt parse analyzer lint

### LINKED ISSUE #3273: Generated `BigInt.parse(dynamic)` triggers `argument_type_not_assignable` lint
## Describe the bug

The generated Dart bindings call `BigInt.parse` with a `dynamic` value:

```dart
@protected
BigInt dco_decode_U128(dynamic raw) {
  return BigInt.parse(raw);
}
```

With strict analysis enabled, this produces:

```
The argument type 'dynamic' can't be assigned to the parameter type 'String'.
```

The generated code forces users to either suppress the lint or manually edit generated files after every code generation.

## Steps to reproduce

1. Generate bindings for a Rust function using a `u128` (or another type that generates `BigInt.parse(raw)`).
2. Run `dart analyze` with strict analysis enabled.
3. Observe the analyzer error on the generated file.

## Logs

No code generation errors.

The issue appears in the generated Dart code during static analysis.

## Expected behavior

The generated code should satisfy the Dart analyzer without requiring users to suppress lints.

For example:

```dart
return BigInt.parse(raw as String);
```

or another type-safe approach if `raw` is guaranteed to be a `String`.

If there are cases where `raw` may not always be a `String`, the generated code should perform whatever conversion is appropriate instead of passing `dynamic` directly to `BigInt.parse`.

## Generated binding code

```dart
@protected
BigInt dco_decode_U128(dynamic raw) {
  return BigInt.parse(raw);
}
```

## Development Environment

**OS:** macOS

**Version of `flutter_rust_bridge_codegen`:** 2.13.0-beta.1

**Version of `flutter_rust_bridge`:** 2.13.0-beta.1

## Additional context

This isn't a runtime issue, but it does cause analyzer errors in generated code. Generated bindings should ideally be analyzer-clean without requiring users to ignore `argument_type_not_assignable` or modify generated files after each regeneration.

fzyzcjy:
## Summary
- Add `argument_type_not_assignable` to generated Dart `ignore_for_file` headers
- Regenerate example bindings so DCO `BigInt.parse(dynamic)` output is analyzer-suppressed

## Validation
- `./frb_internal precommit-generate`
- `./frb_internal generate-internal-rust`
- `./frb_internal lint --fix`
  - Note: `frb_hooks` pana reported a dartdoc `RangeError` and scored 150/160, while static analysis remained 50/50 and the command exited 0.

Close #3273


fzyzcjy:
Gemini suggested reverting the global generated-file ignore and changing the `BigInt.parse(raw)` call site instead. I am not taking that suggestion here because this PR intentionally follows the maintainer direction from issue #3273: generated Dart files already centralize several analyzer suppressions in the common `ignore_for_file` header, and this change adds `argument_type_not_assignable` to that generated-only list. The follow-up commit did accept Gemini's actionable cleanup about removing the now-redundant web-only ignore.

ganeshrvel:
Thanks for the quick fix!

I don't think this should be fixed by suppressing the analyzer warning. The underlying issue is that the generated code isn't type-safe:

```dart
return BigInt.parse(raw);
```

Since `BigInt.parse` expects a `String`, perhaps the generator could emit a type-safe conversion instead, if that's valid for this code path.

Adding `argument_type_not_assignable` to the global `ignore_for_file` list hides the warning GLOBALLY, but the generated code itself still has the same issue.

cc @fzyzcjy

fzyzcjy:
>  the generated code isn't type-safe

iirc Dart itself is type-safe - when casting to wrong type it will just throw; but yes we may make it more typed

ganeshrvel:
@fzyzcjy I see you have closed this. Do we have a better way to fix the typing issue?

fzyzcjy:
I was letting agent create a new pr for it

PR #3287, opened 2026-06-29
Fix CST codegen for casted primitive i64 DTOs

Samiisd:
## Summary
- treat type_64bit_int casted primitives like primitive wire values for CST list allocation
- wrap optional casted primitives like other primitive-like optionals so Option<i64> uses *mut i64 instead of *mut Option<i64>
- add a native-only DTO regression fixture covering Vec<i64> and Option<i64> with full_dep/type_64bit_int

## Validation
- cargo test --manifest-path /tmp/flutter_rust_bridge-upstream/frb_codegen/Cargo.toml test_native_only_dto_i64_type_64bit_int -- --nocapture
- cargo test --manifest-path /tmp/flutter_rust_bridge-upstream/frb_codegen/Cargo.toml casted_primitive_
- minimal repro: upstream master generated code fails with *mut Option<i64> and <i64>::new_with_null_ptr(); patched branch generated code passes cargo check


fzyzcjy:
Hmm seems the PR is closed, feel free to ping me if you have any questions!

EDIT: oh I see the new PR

PR #3288, opened 2026-06-29
Fix CST codegen for casted primitive i64 DTOs

### LINKED ISSUE #3289: CST codegen fails for i64 DTO fields with type_64bit_int
When using the native/CST DTO path with `type_64bit_int: true`, DTO fields containing `Vec<i64>` and `Option<i64>` generate Rust code that does not compile.

Minimal shape:

```rust
pub struct NativeOnlyDto {
    values: Vec<i64>,
    maybe_value: Option<i64>,
}

#[flutter_rust_bridge_macros::frb(semi_serialize)]
pub fn accept_native_only_dto(arg: NativeOnlyDto) {
    drop(arg);
}
```

With `full_dep: true`, `web: false`, and `type_64bit_int: true`, current master emits CST code such as:

```rust
new_leak_vec_ptr(<i64>::new_with_null_ptr(), len)
maybe_value: *mut Option<i64>
```

This fails to compile because `i64` does not implement `NewWithNullPtr`, and the optional wire type recursively tries to decode `*mut Option<i64>` as `Option<_>`.

samiisd:
## Changes

Fixes #3289

## Optional tools

* It is recommended to use the `/frb-develop-feature` skill, which contains detailed guidance for the agent to develop a feature.
* There is a devcontainer and Dockerfile at `.devcontainer`, which optionally helps creating a standardized development environment.


fzyzcjy:
Great job! Could you please run https://github.com/fzyzcjy/flutter_rust_bridge/tree/master/.claude/skills/frb-issue-to-green-pr (which tells the agent to add tests etc) and have a review on the code, and I will review it ASAP (hopefully in weekends)

fzyzcjy (changes requested):
This review was performed by Codex, an AI coding agent, with independent semantic, architecture, performance, adversarial, and test-validity passes.

Requesting changes for two Web codegen blockers:

1. **[P1] `delegate.rs:146-151,174-177` emits `CastedPrimitive` as Common code with an IO-only signature.** Converting the formatted body with `.into()` stores it in `Acc::common`; Common signatures are then generated using `Target::Io`. The resulting `cst_encode_CastedPrimitive_*` therefore returns `int`, while the delegated 64-bit primitive encoder returns `JSAny` on Web. For `I64`/`Isize`, there is an additional input mismatch: the casted API exposes `int`, but `cst_encode_i_64` takes `PlatformInt64`, which is `BigInt` on Web. Please emit target-specific IO/Web bodies and signatures (including `BigInt.from(raw)` for signed Web inputs), and cover this with a Web-enabled Dart generation/analyzer test.

2. **[P1] the new boxed/optional path still classifies `CastedPrimitive` as non-primitive on Web.** `parser/mir/parser/ty/optional.rs:43-44` wraps it in `Boxed`, but the Web branches in Dart `encoder/ty/boxed.rs:55-73` and Rust `decoder/ty/boxed.rs:112-126` still rely on `inner.is_primitive()`, which is false for `CastedPrimitive`. On Dart this declares the boxed encoder as returning an `int /* *JsValue */` while its body returns `JSAny`. On Rust, Optional can generate `CstDecode<Option<i64>> for *mut JsValue` with a body calling `self.map(...)`, which does not compile for a raw pointer. Please centralize the wire-primitive classification and use it consistently in the allocator, wire-type, pointer, and Optional decisions.

The current regression coverage cannot detect either failure: `native_only_dto_i64/flutter_rust_bridge.yaml` sets `web: false`, and `wire/rust/mod.rs:57-95` only invokes the Rust generator and checks substrings. It never generates or analyzes Dart or compiles the complete generated binding. Please route this fixture through the normal full codegen/compile path and add a Web-enabled variant. Also, `optional.rs:63-96` tests a `#[cfg(test)]` copy of the production match table; make that helper the production classifier or test the parser end-to-end so the test fails when production behavior regresses.

Review process: the existing Gemini review was inspected; its two `matches!` suggestions are already reflected in the current code and are not actionable. No performance blocker was found. No fixes were made because this was a review-only pass. The remaining Codecov patch/project failures are consistent with the untested Dart paths above.

fzyzcjy:

PR #3294, opened 2026-07-03
Fix nondeterministic CargoKit JNI packaging for Android library plugins

### LINKED ISSUE #3293: fix(cargokit): fix non-deterministic native library packaging for Android library plugins
### Problem
When using CargoKit with modern versions of the Android Gradle Plugin (AGP), compiled Rust shared libraries (`.so` files) can be missing from the packaged Android library project outputs (AAR / final APK debug builds). 

This happens because the plugin's `plugin.gradle` only registers task dependencies for `merge[Variant]NativeLibs` (e.g. `mergeDebugNativeLibs`), which is an application-level task and does not exist in Android library plugin projects (which apply `'com.android.library'`). JNI libraries for Android library modules are copied/merged by different tasks, such as:
- `merge[Variant]JniLibFolders`
- `copy[Variant]JniLibsProjectOnly`
- `copy[Variant]JniLibsProjectAndLocalJars`

Because no task dependency was declared for these library-specific JNI compilation tasks, CargoKit's compile task was run concurrently or was not guaranteed to complete before packaging, resulting in a packaging race condition where `.so` files were omitted from the build.

### Solution
This PR extends the CargoKit task dependency injection to include these library project JNI tasks:
- `merge${buildType.capitalize()}JniLibFolders`
- `copy${buildType.capitalize()}JniLibsProjectOnly`
- `copy${buildType.capitalize()}JniLibsProjectAndLocalJars`

This ensures that the Android build system waits for Cargo to compile the library under all configurations (including debug/release and all target ABIs) before assembling the library assets.

samiisd:
## Problem

Modern Android Gradle Plugin library projects do not use only the application-oriented `merge<Variant>NativeLibs` task when packaging JNI libraries. CargoKit therefore did not guarantee that its Rust build completed before every Android library JNI copy/merge task, which could omit the generated `.so` from an AAR or final APK.

## Solution

- Register each CargoKit build task as a dependency of the Android application and library JNI consumers:
  - `merge<Variant>NativeLibs`
  - `merge<Variant>JniLibFolders`
  - `copy<Variant>JniLibsProjectOnly`
  - `copy<Variant>JniLibsProjectAndLocalJars`
- Maintain the implementation in the canonical CargoKit repository through fzyzcjy/cargokit#12.
- Update both CargoKit template submodule pointers to the same canonical commit and regenerate all checked-in CargoKit copies.
- Keep the CargoKit Android fixture compatible with current Flutter stable by using its minimum supported Gradle 8.7, Android Gradle Plugin 8.6, and Kotlin Gradle Plugin 2.0 versions.

## Regression coverage

- Assert the direct Gradle task dependency graph for debug, profile, and release Android library variants.
- Build clean Android APKs and assert that `lib/arm64-v8a/libhello_rust_ffi_plugin.so` is present in the packaged archive.
- Cover Ubuntu, macOS, and Windows through the CargoKit build matrix.

## Local verification

- `./frb_internal precommit-integrate`
- `./frb_internal sync-cargokit-copies`, followed by a clean-tree check
- `./frb_internal lint --fix`
- Clean debug APK build with `lib/arm64-v8a/libflutter_package.so` present
- Controlled negative gate: the dependency regression fails with the previous merge-only wiring and passes with this change

Closes #3293


fzyzcjy:
(same as another PR - would be great to execute that skill)

fzyzcjy (changes requested):
**Agent-authored review:** this review was performed by Codex using independent correctness, test-weakening/doc-compliance, and adversarial code-quality passes.

### [P1] Apply the fix to the canonical CargoKit source

Representative location: `frb_example/flutter_package/cargokit/gradle/plugin.gradle:178` (the same issue applies to all seven changed files).

These files are downstream copies, not the CargoKit implementation shipped by FRB templates. Both canonical template submodules remain pinned to `6f7144d192b04075acd73af2656cdebb1d3e4055`, whose `plugin.gradle` still wires only `merge...NativeLibs`. `sync-cargokit-copies` overwrites four of these files, while integrate generation recreates the other three. The current CI confirms this exact failure: the integrate jobs regenerate `plugin.gradle` back to the canonical version and fail the clean-tree check, and `Generate :: Internal` also fails.

Consequently, newly created or integrated user projects still receive the broken task dependency, and normal generation will overwrite this patch. Please implement and test the change in `fzyzcjy/cargokit`, update both integration-template submodule pointers, then run `./frb_internal precommit-integrate` (and the normal CargoKit-copy sync/generation checks) to regenerate the checked-in copies.

### [P2] Add a packaging regression that checks the native artifact

This fix depends on AGP-internal task names and ordering, but the PR adds no regression proving that the Rust `.so` reaches the package. A Gradle/Flutter build can succeed while omitting the library, which is the reported failure mode. In the canonical CargoKit change, please add a clean Android library packaging test that assembles an AAR or consuming APK and asserts the expected `lib/<abi>/lib<name>.so` archive entry, covering the supported AGP path(s).

Test-weakening gate: no tests or assertions were weakened. No file-size or concrete performance regression was found. A fresh `/gemini review` was requested, but Gemini reports that its GitHub review service has been sunset; the older Gemini review is anchored to a stale commit and discusses code no longer present in this PR.

fzyzcjy:

Issue #3296, opened 2026-07-05
Native Assets build hook runs for every `flutter test` file, causing massive slowdown


ganeshrvel:
## Version

* `flutter_rust_bridge: 2.13.0-beta.4`

After migrating from Cargokit to Native Assets, I noticed a very large increase in `flutter test` execution time.

The tests themselves haven't changed. The only change was moving to the Native Assets integration.

## Build hook

My build hook is essentially just:

```dart
import 'package:flutter_rust_bridge_hooks/flutter_rust_bridge_hooks.dart';

Future<void> main(List<String> args) async {
  await build(args, (input, output) async {
    await const FlutterRustBridgeNativeAssetsBuilder(
      cratePath: '<my_rust_crate>',
    ).run(
      input: input,
      output: output,
    );
  });
}
```

No additional work is performed.

## Test initialization

Each test initializes the already-built Rust library directly:

```dart
Future<void> initializeRustLibraryForTests() async {
  await RustLib.init(
    externalLibrary: ExternalLibrary.open(
      rustLibraryPath,
    ),
  );
}
```

A typical test looks like:

```dart
void main() {
  setUpAll(() async {
    await initializeRustLibraryForTests();
  });

  tearDownAll(RustLib.dispose);

  test('example', () {
    // test body
  });
}
```

## Test layout

The project has many small test files, for example:

```text
test/
  codec_a_test.dart
  codec_b_test.dart
  codec_c_test.dart
  helper_a_test.dart
  helper_b_test.dart
  ...
```

Each test file is lightweight and executes quickly.

## Observed behavior

It appears the Native Assets build hook is executed for every individual test file.

For projects with many test files, the startup cost becomes highly noticeable because every test process waits for the build hook before executing.

The Rust library is already built and the tests simply load it via `RustLib.init(...)`, so the repeated build hook invocation becomes the dominant cost.

I also noticed that this happens even for test files which does not import `flutter_rust_bridge`, does not call `RustLib.init(...)`, and does not use the Rust library at all. Those tests still appear to wait for the Native Assets build hook before executing, adding unnecessary overhead to the entire test suite.

## Expected behavior

Ideally one of the following would happen:

* the Native Assets build hook is executed once per `flutter test` invocation rather than once per test file, or
* the hook framework detects that nothing has changed and skips almost immediately, making the per-file overhead negligible.

fzyzcjy:
(in a ddl and will reply as soon as it finished)

[...]

PR #3305, opened 2026-07-10
Add manual reproduction for Native Assets test hook overhead

fzyzcjy:
## Intentional reproduction

This PR adds a mechanical manual procedure for #3296. It is not a runtime fix and does not close the issue.

## Result on the current environment

The expanded procedure covers cold single-file, explicit-two-file, no-argument default discovery, and an immediately repeated identical warm invocation.

| Fixture | Cold single | Cold explicit two | Cold default discovery | Identical warm repeat |
| --- | ---: | ---: | ---: | ---: |
| FRB `2.13.0-beta.6` | 1 hook / 12.867s | 1 hook / 9.409s | 1 hook / 8.921s | `1 -> 1` / 1.039s |
| `native_toolchain_rust` `aeda048` | 1 hook / 14.247s | 1 hook / 10.931s | 1 hook / 6.218s | `1 -> 2` / 7.326s |

All selected tests passed. Neither fixture multiplied hook processes by the number of test files. FRB skipped the hook for the identical warm invocation; vanilla reran it and reported `File modified during build. Build must be rerun.` Timings describe this environment only and are not a cross-fixture Rust build benchmark.

Flutter 3.44 invokes its Native Assets builder once before dispatching the collected test files to the test runner. Hook discovery follows the package dependency graph, so whether an individual test imports FRB does not control package-level Native Assets preparation.

The reported per-file behavior is therefore not reproducible on this revision, and there is no evidence-backed FRB runtime change to make. A remaining reproduction would need the reporter's exact Flutter revision, command, and cache state.

## Validation

- Full manual matrix rerun in the per-worktree FRB Docker environment.
- `./frb_internal lint --fix`
- Independent correctness and test-weakening reviews: no findings after fixes.
- Mechanical procedure: `tools/manual_tests/2026-07-10-native-assets-test-hook-overhead.md`
- Initial explicit-two-file evidence: https://gist.github.com/fzyzcjy/bff5d36467bcc53b56e0abaeab66f109


fzyzcjy:
Addressed the review findings in 9723ecf3b0: the procedure now compares clean one-file and two-file runs, reports raw count deltas without claiming root cause, pins the vanilla upstream checkout and fixture path, and records actual Dart process IDs.

fzyzcjy:
Addressed the latest Gemini feedback in 85dbdf9dc2: configuration snippets are now YAML/TOML blocks rather than shell commands, and the results list matches the generated marker files.

fzyzcjy:
Addressed the final Gemini review in the new commit: both comparison loops now fail explicitly when the marker is absent, and the pinned vanilla fixture has executable marker, test-file, and comparison commands.

fzyzcjy:
Full review-corrected manual execution evidence, with local paths and machine identifiers sanitized: https://gist.github.com/fzyzcjy/20a002cad09dda0ac8e7411cd3717b33

Issue #2978, opened 2026-02-13
Cancellation Notifier for `StreamSink`


Shadowcat650:
## Description

From #2917:
> Currently, [StreamSink](https://cjycode.com/flutter_rust_bridge/guides/types/translatable/stream) in flutter_rust_bridge supports sending events from Rust to Dart via streams, but there’s no mechanism for Rust to be notified when the Dart-side subscription is cancelled — for example, when a user calls subscription.cancel() or when the Dart stream is disposed.
> 
>This limitation can lead to resource leaks if the Rust side holds long-lived resources such as sockets, threads, or file handles that should be cleaned up once the Dart listener stops.

## Proposed Solution

Add a `CancelNotifier` future that can be obtained by calling `.cancel_token()` on the `StreamSink`. This object would be similar to a [`tokio::sync::Notify`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html), except that it can be set once and never again.

```rust
struct MySubscriber {
    sink: StreamSink<MyObject>,
}

impl MySubscriber {
    async fn subscribe(self, mut rx: tokio::sync::mpsc::Receiver<MyObject>) {
        let cancel = self.sink.cancel_token();
        loop {
            tokio::select! {
                val = rx.recv() => {
                    match val {
                        Some(v) => {
                            let _res = self.sink.add(v); // This result theoretically could be used to detect sink closure (on Err).
                        }
                        None => break, // channel (not the sink) closed
                    }
                }
                _ = cancel.notified() => {
                    break;
                }
            }
        }
    }
}
```

## Motivation

I am building a normalized cache for my Flutter app that feeds fresh data to the UI. When the UI stops listening to the stream, I don't have a way to clean it up (I *think* I can detect this once something fails to send with `StreamSink::add()`, but I leak memory indefinitely if the watched object never updates again).

fzyzcjy:
Yes, that looks pretty reasonable and feel free to PR for it! (I plan to handle all pending prs recently btw)

Issue #1836, opened 2024-03-26
memory leak of ReceivePort when using create_stream


uditrugman:
### Describe the bug

the current implementation closes the receive port when a "done" is received on the stream, but if the dart side cancelled the subscription before receiving "done", then the receive port will remain open on the dart side. Then, even if closing the stream, the receive port will never be released.

an alternative would be the close outside the for loop, but that would cause another issue.
if a client (dart side) closes the port, there could be a race where rust can still send through the sendPort and that would lead to an error.

see the code below.

```dart
  Stream<S> _executeStreamInner<S, E extends Object>(
      StreamTask<S, E>? task) async* {
    final portName =
        ExecuteStreamPortGenerator.create(task!.constMeta.debugName);
    final receivePort = broadcastPort(portName);

    task.callFfi(receivePort.sendPort.nativePort);

    final codec = task.codec;
    task = null;

    await for (final raw in receivePort) {
      try {
        yield codec.decodeObject(raw);
      } on CloseStreamException {

        // BUG: we only close when "done" is received.
        receivePort.close();
        break;
      }
    }
  }
```

### Steps to reproduce

// the below code demonstrates how the leak can happen. all is needed to cancel the subscription and the receive port will remain open

```
  print("creating");
  final s = createLogStream();
  print("listening");
  
  final subscription = s.listen((event) {
    print("event: ${event.msg}");
  }, onDone: () {
    print("done");
  });

  print("sending 1");
  sendEvent(msg: "msg 1");
  print("sending 2");
  sendEvent(msg: "msg 2");
  print("done");

  await Future<void>.delayed(const Duration(seconds: 1));
  
  // the cancel below causes the for loop to end without receiving the "done" event on the stream
  subscription.cancel();
```

### Logs

```shell
there are no logs.
```

fzyzcjy:
Good observation! What if we close the port using a `finally` block outside the for (instead of the catch block inside for)?

uditrugman:
i commented on that as well. it's not valid because then the receive port will be closed while the rust side "sendPort" is still open and rust might send events.
there is a race condition here that unless the dart side knows how to "close" the stream, then the two sides are not in sync.

I'd suggest that the generated code will know how to close the stream and when a client cancells subscription, the streams should be closed as well

uditrugman:
[...]

PR #3311, opened 2026-07-15
Fix StreamSubscription.cancel() hanging for an idle RustStreamSink

dbsxdbsx:
## Changes

`StreamSubscription.cancel()` on a `RustStreamSink` stream can hang forever when the Rust side is idle (never sends another message and never closes the stream). This is a distinct bug from #1836 (which fixed eventual `ReceivePort` release): here the cancel *Future itself* never completes.

### Root cause

The Dart side wrapped the receive port in an `async*` generator that did `await for (receivePort)`. A generator suspended in `await for` cannot be interrupted by cancelling its subscription 鈥?`cancel()` only completes once the generator unwinds through its `finally`, and the `finally` (which closes the port) is exactly what can never run while the generator is parked on an idle port. Closing the port first only helps on native, where `ReceivePort.close()` delivers a done event; on web, `BroadcastChannel.close()` delivers nothing and the generator stays parked.

### Fix

Listen to the receive port directly (manual `StreamController` + subscription) instead of going through an `async*` generator. Cancelling a plain subscription completes immediately and identically on native and web. The previous semantics are preserved: early messages are still buffered before the consumer listens, and a decoded error/panic still terminates the stream after the error event. The now-unused `async` dependency is removed.

Because the port is now closed promptly on cancel, `sink.add(...)` can legitimately return `Err` afterwards, so the `event_listener` example fixtures no longer `.unwrap()` that result, and the related regression test now `await`s the cancellation with a timeout instead of leaving it unawaited.

### Reproduction (before this PR)

- Baseline commit: `1d5348baae` (current `master`)
- Steps: create a `RustStreamSink<int>`, `setupAndSerialize`, `listen`, send nothing from Rust, then `await subscription.cancel().timeout(Duration(seconds: 1))`
- Observed: `TimeoutException after 0:00:01.000000: Future not completed` 鈥?on both `dart test` (VM) and `dart test -p chrome` (web)
- Expected: `cancel()` completes promptly regardless of platform or Rust activity

### Testing

- `frb_dart`: new cancellation regression tests pass on VM and Chrome; full `frb_dart` suite green.
- `frb_example/pure_dart` and `frb_example/pure_dart_pde`: `event_listener` native integration tests pass (the sync send-after-cancel test previously panicked via `.unwrap()`; now green).
- Generated twins / `pure_dart_pde` copies regenerated via `./frb_internal generate-internal-frb-example-pure-dart` (idempotent, no drift). `cargo fmt`/`clippy`, `dart analyze`/`format` all clean.
- Not run locally (Rust/wasm env): web (wasm) integration tests for the examples and the full example suites 鈥?left to CI.

### Scope

This fixes only the Dart-side idle-cancel deadlock. The separate feature of notifying Rust when a Dart subscription is cancelled (a `cancel_token()` on the Rust `StreamSink`, discussed in #2978) is intentionally **not** included here and should be designed/aligned separately.

## Checklist

- [x] New tests are added to ensure the fix works
- [x] `./frb_internal precommit` run in a clean environment (codegen for the example was run via `generate-internal-frb-example-pure-dart`; full precommit left to CI)
- [x] Reproduction report included above


frb_dart/lib/src/stream/stream_sink.dart
```diff
@@ -40,21 +39,59 @@ _State<T> _setup<T>(BaseCodec<T, dynamic, dynamic> codec) {
...
 
-  final stream = rawStream.listenAndBuffer();
+  controller
+    ..onPause = portSubscription.pause
+    ..onResume = portSubscription.resume
+    ..onCancel = () {
+      terminated = true;
+      receivePort.close();
+      return portSubscription.cancel();
+    };
```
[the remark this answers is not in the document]
dbsxdbsx: Good catch — adopted in 5b1d23a. Both `onPause` and `onResume` are now guarded with `!terminated` to prevent a `StateError` when the port subscription has already been cancelled by `terminate()`.

frb_dart/lib/src/stream/stream_sink.dart
```diff
@@ -40,21 +39,59 @@ _State<T> _setup<T>(BaseCodec<T, dynamic, dynamic> codec) {
...
+  // Listen to the port directly instead of wrapping it in an `async*` generator
+  // that does `await for (receivePort)`. A generator suspended in `await for`
+  // cannot be interrupted by cancelling its subscription, so if Rust stays idle
+  // (never sends another message and never closes the stream) then
+  // `StreamSubscription.cancel()` would hang forever. Closing the port only
+  // wakes such a generator on native (where `ReceivePort.close()` delivers a
+  // done event) but not on web (where closing a `BroadcastChannel` delivers
+  // nothing), so `await for` is fundamentally unsafe here. A plain subscription
+  // can always be cancelled immediately and identically on every platform.
+  final controller = StreamController<T>(sync: true);
```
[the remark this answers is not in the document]
dbsxdbsx: Intentionally keeping `sync: true` — the previous implementation used `listenAndBuffer()` from `package:async`, which also creates a `StreamController(sync: true)`. This is important for two reasons:

1. **Event ordering**: an async controller delivers events in the next microtask, which could reorder early-arriving messages relative to when the consumer starts listening — exactly the scenario that `listenAndBuffer()` (and this replacement) is designed to handle correctly.

2. **Behavioural parity**: changing to an async controller would be a semantic change from the original implementation, not just a refactor. Since this PR is a targeted bug fix (cancel deadlock), minimizing unrelated behavioral changes is preferable.

The concern about listener exceptions propagating back is valid in general, but the `onData` callback here (`controller.add`) is our own code (not user code), so synchronous delivery is safe in this context.

dbsxdbsx:
**CI note:** The single failing job `Test :: Rust (ubuntu-latest, nightly)` is an ICE (Internal Compiler Error) in the nightly rustc while compiling `frb_codegen` with `cargo-llvm-cov` coverage instrumentation:

```
error: the compiler unexpectedly panicked. This is a bug
note: compiler flags: -C embed-bitcode=no -C debuginfo=2 -C instrument-coverage
```

All Rust tests themselves passed (`0 failed`). The crash is in `frb_codegen` — unrelated to any files changed in this PR (`frb_dart` + `frb_example` event_listener). The same job passed on `master` at `1d5348baae`, so this is a rolling-nightly regression in the compiler, not something introduced by this change.

All other 165 jobs (Generate, Lint, Test Dart/Flutter native/web, Build, Bench) are green.

fzyzcjy:
Great job! I will have a check as soon as having time.

frb_dart/lib/src/stream/stream_sink.dart
```diff
@@ -15,6 +15,10 @@ class RustStreamSink<T> {
     return serializeNativePort(_state!.receivePort.sendPort.nativePort);
   }
 
+  /// {@macro flutter_rust_bridge.internal}
+  @visibleForTesting
+  dynamic get debugSendPort => _state?.receivePort.sendPort;
```
fzyzcjy: [P2] Keep the raw transport out of the public API

`RustStreamSink` is re-exported from `flutter_rust_bridge.dart`, so `@visibleForTesting` does not keep this getter internal; it adds a consumer-visible member that returns a platform-specific `dynamic` port and exposes the private `[action, payload]` wire format. Its only caller is a VM-only coverage test, and that test already demonstrates that the API is not cross-platform by skipping web. Please remove this getter and use a non-exported/injected bridge harness (or an existing generated Rust-to-Dart integration stream) for these tests instead of expanding the production API for coverage.

frb_dart/lib/src/stream/stream_sink.dart
```diff
@@ -40,21 +44,65 @@ _State<T> _setup<T>(BaseCodec<T, dynamic, dynamic> codec) {
...
+  portSubscription = receivePort.listen(
+    (raw) {
+      final T decoded;
+      try {
+        decoded = codec.decodeObject(raw);
+      } on CloseStreamException {
+        terminate();
+        return;
+        // coverage:ignore-start
+      } catch (error, stackTrace) {
```
fzyzcjy, line 86: [P2] Test this reachable error path instead of excluding it

This branch does not require native FFI: a test source can feed malformed DCO input such as `[0, 'not-an-int']` to the existing `raw as int` decoder and assert that the returned stream emits the decode error and then closes. That is an important behavior newly implemented by this PR; if `addError` or `terminate` regresses, the current suite still passes while Codecov reports all modified lines covered. Please remove this ignore and cover error-then-done through a non-public injection seam. The separate port-level `onError` ignore can remain if that transport error truly cannot be injected.

frb_dart/test/stream_sink_test.dart
```diff
@@ -17,4 +29,73 @@ void main() {
...
+
+    // DcoCodec wire format: [action, payload]. Action 0 = success.
+    (sink.debugSendPort as dynamic).send([0, 42]);
+    await Future<void>.delayed(const Duration(milliseconds: 50));
+
+    expect(items, [42]);
+    await subscription.cancel().timeout(const Duration(seconds: 1));
+  });
+
+  test('pause then resume then cancel completes', () async {
```
fzyzcjy: [P2] Verify upstream pause/resume propagation, not just callback execution

This test only calls `pause()`, `resume()`, and then `cancel()`; it still passes if both forwarding callbacks in `stream_sink.dart` are replaced with no-ops. Because this PR now owns the forwarding that prevents the controller from buffering an unbounded Rust event stream while a consumer is paused, the regression test should use an injectable source with observable `onPause`/`onResume` signals and assert that the upstream subscription actually pauses and resumes. That harness can also replace the public `debugSendPort` seam above.

fzyzcjy (changes requested):
Agent review performed by Codex (not a human reviewer).

I reviewed the base-to-head diff with independent passes for semantics/contracts, lifecycle/concurrency, architecture, performance/resources, test weakening, and an adversarial integration check. The core direct-subscription fix looks correct: eager buffering and synchronous ordering match the old `listenAndBuffer()` implementation, cancellation now reaches the receive-port subscription directly, lifecycle guards are sound, and the event path is neutral-to-better in overhead. All 176 current CI checks are green.

I am requesting changes for the three inline findings: a coverage-only test seam leaking raw transport into the public API, a reachable decode-error branch hidden from coverage, and a pause/resume test that does not verify upstream propagation/backpressure.

Gemini status: its earlier review was checked; the pause/resume-after-termination issue was fixed in `5b1d23a`, while the `sync: true` concern is not actionable because the previous helper also used a synchronous controller and Dart stream callbacks are zone-guarded. I requested a fresh Gemini pass on the latest head, but the bot reported that its GitHub review service has been sunset. No code was changed by this agent review.

fzyzcjy:

frb_dart/lib/src/stream/stream_sink.dart
dbsxdbsx: Addressed in d8b057a.

Removed `debugSendPort` from `RustStreamSink`. Tests now inject a raw event source via a top-level `rustStreamSinkWithRawSourceForTest` helper (imported only through `package:flutter_rust_bridge/src/...`). Both package entrypoints now `export ... show RustStreamSink`, so that helper stays out of the public API.

frb_dart/lib/src/stream/stream_sink.dart
dbsxdbsx: Addressed in d8b057a.

Removed the `coverage:ignore` around the decode `catch`. The new injected-source test feeds `[0, 'not-an-int']` and asserts the stream emits the decode error and then closes. Left the port-level `onError` ignore in place, since that transport path still cannot be injected.

frb_dart/test/stream_sink_test.dart
dbsxdbsx: Addressed in d8b057a.

Replaced the old pause/resume/cancel smoke test with one that uses an injectable `StreamController` with observable `onPause`/`onResume` callbacks and asserts the upstream source actually pauses and resumes. Same harness as the other two points.

dbsxdbsx:
@fzyzcjy , is it now all ok?   I have other things I want to PR, so I don't want to bother you too much. I would like to get this existing PR completely resolved first.

fzyzcjy:
given this changes the core logic about stream binding, I may need to do a thorough review (by me, not by agent), and to be honest I am a bit worried about whether we miss some edge case etc about such stream binding since it is easy to miss. too busy recently, and I will try to squeeze time next weekend to check this!

dbsxdbsx:
(Frankly, the below is all built by AI, but triggered by I with what @fzyzcjy  you said, since there maybe some corner case hard to find.)

Pushed one more commit before you look at this. While hunting for edge cases myself I found two latent problems in my own previous commit and a missing test, so I'd rather you review the fixed version.

### On the edge-case worry

Instead of arguing from reading the diff, I ran the old binding (`async*` + `listenAndBuffer()`) and the new one against the same 11 contracts: pre-listen buffering, event ordering, close-stream, decoded error, source done, `cancelOnError`, single-subscription, pause/resume backpressure, cancel from inside `onData`, never-listened, and idle cancel.

New: 11/11. Old: 7/11 — and all four failures are the same idle-`cancel()` hang this PR fixes, not semantic differences. So the behavioural change really is confined to cancellation. Happy to push that A/B harness into the repo as a test file if you'd find it useful.

### Why the extra commit

1. Last round's test seam made `_State.receivePort` nullable and added a private `RustStreamSink._withState`. That is the same smell you flagged, just less visible — `setupAndSerialize()` on an injected sink dereferences a null port. Only reachable from tests, so nothing shipped was broken, but the seam belongs on the stream-building function instead. `RustStreamSink` and `_State` are now byte-identical to master, so the production diff is only the binding itself.

2. `terminate()` reached the subscription through a `late final`. A `Stream` is allowed to report done or error before `listen` returns, and that read then throws `LateInitializationError`. Receive ports never do this, so again nothing shipped was broken, but the function now takes an arbitrary `Stream`, so the reference is nullable and gets cancelled after `listen` if that window was hit. Regression test added.

3. The main thing `listenAndBuffer()` gave us — events arriving before the consumer listens are buffered and replayed in order — had no test at all. Added that, plus close-stream, drop-after-termination, pause-buffering-with-replay, and source-level errors (which also removed the last `coverage:ignore` in this file).

### Two things I left out of this PR on purpose

- Rust-side `sink.add()` now returns `Err` once Dart has cancelled, where before it kept returning `Ok` forever. That is the intended consequence of closing the port promptly — it is why the `event_listener` fixtures had to drop `.unwrap()` — but it is user-visible, so it may deserve a line in the release notes or on the `Stream` doc page. Tell me if you want that here or in a follow-up.
- Incidentally, this also restores the only signal Rust currently has for "Dart stopped listening", which is exactly the complaint in #2978. Just noting it — not proposing anything in this PR.

fzyzcjy:
looks great! (btw would be great to let agent use skills to make ci pass)

dbsxdbsx:
[  @fzyzcjy  Still, the below are all produced by AI and after heavy check , roughly speaking, all these CI failures are not due to my PR. ]

### CI note

Short version: four of the five failures in run [30806590936](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/30806590936) are environment issues, and the fifth — `Test :: Dart :: Web (frb_example--pure_dart)` — is a pre-existing flake that `master` reproduces. While digging into it I did find one genuine test bug, which is what 2c9fa22217 fixes.

The four environment failures, none of which touch this PR:

| Job | Error |
|---|---|
| `Test :: Rust (nightly)` | `cargo-llvm-cov`: `failed to collect object files` (the Rust tests themselves were 36 passed / 0 failed) |
| `Generate :: Command Integrate` | crates.io returned HTTP 500 three times for `parking_lot` |
| `Build :: Flutter (macos-15-intel)` | `Error running pod install` |
| `Test :: Flutter :: Native :: macos` | `Download failed! Please check passed arguments.` |

#### 2c9fa22217: three example tests were never awaited

`handle_stream_sink_at_{1,2,3}` call the async `testHandleStream` helper without `async`/`await`:

```dart
test('dart call handle_stream_sink_at_2', () {
  testHandleStream(handleStreamSinkAt2TwinNormal);   // not awaited
});
```

The test passes instantly, the stream keeps running in the background, and `expect(cnt, max)` fires during some later unrelated test — hence the original `handle_stream_sink_at_2` failure being reported as *"This test failed after it had already completed"*. It also let the three streams race each other. This predates the PR (abca19c697, Nov 2023).

2c9fa22217 adds the missing `async`/`await`; twins and the `pure_dart_pde` copies were regenerated with `./frb_internal generate-internal-frb-example-pure-dart` (idempotent — a second run adds no drift). The four changed `pure_dart` test files pass natively, 40/40. Happy to drop this commit if you would rather keep the PR limited to `frb_dart`.

#### The remaining web failures also happen on `master`

`gh run rerun` needs admin rights, so I ran the job on my fork instead — `test_dart_web[package=frb_example--pure_dart]` on upstream `master` (1d5348ba) and on this branch's head, 9 runs each.

The controlled part is three load-matched pairs, one `master` run and one PR run dispatched together so both see the same runner contention:

| Pair | `master` | this PR |
|---|---|---|
| 1 | failed (`stream_sink_inside_struct_twin_normal`) | passed |
| 2 | failed (`stream sink` ×2) | failed (`stream sink` ×2) |
| 3 | passed | failed (`stream sink`) |

2/3 versus 2/3, and in pair 2 both arms failed the *same two tests* while running side by side. Unmodified `master` also produced `ConcatenateWith static stream sink test` with `Expected: '0' / Actual: '4'` — the leading events of a stream were dropped.

Across all 18 runs it is 4/9 for `master` and 7/9 for this PR, but that gap comes entirely from my first two rounds, where I dispatched six runs at once and starved the runners: those took 65–133 s instead of the usual ~50 s and failed with timeouts. Fisher's exact on 4/9 vs 7/9 gives p ≈ 0.17.

Two things I checked that might be useful:

- This branch is not slower — 49 s / 4192 tests at d8b057afed, against 48/49/50/52/55 s for the last five `master` runs.
- `pure_dart` is not special, just the largest web suite: 4191 tests / 54 s, versus 1989 / 21 s for `pure_dart_pde`. That is why it is the one that keeps catching this, and why `pure_dart_pde` passing in the same run is weaker evidence than it looks.

The symptoms fit `BroadcastChannel` not buffering messages posted before a listener attaches: losing data events shows up as `Expected '0' / Actual '4'`, losing the close-stream message shows up as a 30 s timeout. I have not tried to fix that here — it looks orthogonal to this PR and probably deserves its own issue.

One caveat I should be upfront about: on upstream CI this job has been green on `master` for the last five runs, while this PR has hit it twice in three. At n = 3 that is not something counting can settle (p ≈ 0.11), so the argument above rests on the load-matched pairs and on `master` reproducing the identical failures, not on the raw upstream tally. If you want more confidence, re-running the job on this PR a few times would settle it quickly.

fzyzcjy:
looks great, I triggered a rerun as well

fzyzcjy (approved):
stream_sink.dart change LGTM; wait for my agent to add more tests

fzyzcjy:
proof that it reproduces w/o fix: https://github.com/fzyzcjy/flutter_rust_bridge/pull/3347

fzyzcjy (approved):
whole pr lgtm

blocker before merge: agent is fixing ci

fzyzcjy:
web ci err

image

fzyzcjy:
Investigated this specific Web CI failure. Conclusion: it is the same pre-existing Web/BroadcastChannel stream flake, not a regression introduced by this PR.

- **Failed attempt**: the Wasm build, web server, and browser startup all succeeded. The existing `rust_auto_opaque_test.dart` test named `stream sink` then stalled at `(await stream.toList()).single` while waiting for the stream close event and hit the 30-second test timeout. [Failed job](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/31307241345/job/93229452049)
- **Pre-PR master evidence**: master push SHA `576ea02e11423196b4ee7309210f505547ba286c`, using the old async-generator StreamSink implementation, failed the same existing `stream sink` test with the same 30-second timeout. This rules out the controller refactor in this PR as the necessary cause. [Master baseline job](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/31252039186/job/93089799154)
- **Same-head rerun**: rerunning the job on the unchanged PR head `d62ccbf6e5f465496c915e2b27f41f96d0279d9c` passed. [Successful rerun](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/31307241345/job/94972712998)

The likely underlying mechanism is the existing Web `BroadcastChannel` readiness race: an initial data event or the final close event can occasionally be lost. No PR-specific code change is warranted for this failure; rerunning was the correct action. The current check rollup is fully green and the PR merge state is `CLEAN`.

fzyzcjy:
@dbsxdbsx seems ci is red

e.g.
* https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/31873052987/job/94986827417?pr=3359
* https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/31873052987/job/94986827513?pr=3359

could you please have a quick check on it (I will temporarily revert the pr and re-merge after it is solved)

dbsxdbsx:
@fzyzcjy sorry for the late reply — I only just saw the ping.

I checked the two Web jobs on #3359. They look like the same pre-existing BroadcastChannel flake, not an idle-`cancel()` regression and not something #3359's doc change introduced:

- `ConcatenateWith static stream sink` / `handle_stream_sink_at_1` dropping leading events (`Expected: 0` / `Expected: 5, Actual: 4`)
- `stream sink` hitting the 30s timeout (lost close)
- `event listener subscription cancels inside onData` timing out on `cancelled.future` — more likely the `first` event never arrived, so `onData` never ran

#3360 looks like the right place to fix that transport race, and I will not open a competing PR. Happy to help re-land #3311 once #3360 is in.

fzyzcjy:
@dbsxdbsx unreverting this in https://github.com/fzyzcjy/flutter_rust_bridge/pull/3366, and letting agent to check the flaky test in https://github.com/fzyzcjy/flutter_rust_bridge/pull/3360.

btw do you have interest in reviewing #3360 (which I have not reviewed yet) :)

PR #3315, opened 2026-07-19
Support Swift Package Manager

normalllll:
## Changes

Support Swift Package Manager

### DIFF
--- a/frb_codegen/assets/integration_template/cargokit/shared/REPLACE_ME_CARGOKIT_HOOK_DIR/build.dart
+++ b/frb_codegen/assets/integration_template/cargokit/shared/REPLACE_ME_CARGOKIT_HOOK_DIR/build.dart
@@ -1,0 +1,267 @@
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:code_assets/code_assets.dart';
+import 'package:flutter_rust_bridge_hooks/flutter_rust_bridge_hooks.dart';
+
+const _cratePath = 'REPLACE_ME_CARGOKIT_CRATE_PREFIXREPLACE_ME_RUST_CRATE_DIR';
+const _temporaryToolchainMarker =
+    '# Generated temporarily by flutter_rust_bridge Apple build hook.';
+
+void main(List<String> args) async {
+  await build(args, (input, output) async {
+    if (!input.config.buildCodeAssets) {
+      return;
+    }
+
+    final codeConfig = input.config.code;
+    if (!_usesSwiftPackageManagerForApplePlatform(
+      input.outputDirectoryShared,
+      input.packageName,
+      codeConfig.targetOS,
+      output.dependencies.add,
+    )) {
+      return;
+    }
+
+    await _withTemporaryRustToolchain(
+      packageRoot: input.packageRoot,
+      outputDirectoryShared: input.outputDirectoryShared,
+      cratePath: _cratePath,
+      targetTriple: _targetTriple(codeConfig),
+      addDependency: output.dependencies.add,
+      action: () => const FlutterRustBridgeNativeAssetsBuilder(
+        assetName: 'REPLACE_ME_RUST_CRATE_NAME.dylib',
+        cratePath: _cratePath,
+      ).run(input: input, output: output),
+    );
+  });
+}
+
+String _targetTriple(CodeConfig config) {
+  return switch ((config.targetOS, config.targetArchitecture)) {
+    (OS.macOS, Architecture.arm64) => 'aarch64-apple-darwin',
+    (OS.macOS, Architecture.x64) => 'x86_64-apple-darwin',
+    (OS.iOS, Architecture.arm64)
+        when config.iOS.targetSdk == IOSSdk.iPhoneSimulator =>
+      'aarch64-apple-ios-sim',
+    (OS.iOS, Architecture.arm64) => 'aarch64-apple-ios',
+    (OS.iOS, Architecture.x64) => 'x86_64-apple-ios',
+    _ => throw UnsupportedError(
+      'Unsupported Apple target: ${config.targetOS} '
+      'on ${config.targetArchitecture}',
+    ),
+  };
+}
+
+bool _usesSwiftPackageManagerForApplePlatform(
+  Uri outputDirectory,
+  String packageName,
+  OS targetOS,
+  void Function(Uri) addDependency,
+) {
+  final platformDirectory = switch (targetOS) {
+    OS.iOS => 'ios',
+    OS.macOS => 'macos',
+    _ => null,
+  };
+  if (platformDirectory == null) {
+    return false;
+  }
+
+  final manifest = _findFlutterSwiftPackageManifest(
+    outputDirectory,
+    platformDirectory,
+    addDependency,
+  );
+  return manifest != null &&
+      manifest.readAsStringSync().contains('name: "$packageName"');
+}
+
+File? _findFlutterSwiftPackageManifest(
+  Uri outputDirectory,
+  String platformDirectory,
+  void Function(Uri) addDependency,
+) {
+  var directory = Directory.fromUri(outputDirectory);
+  while (directory.parent.path != directory.path) {
+    final pluginDependencies = File.fromUri(
+      directory.uri.resolve('.flutter-plugins-dependencies'),
+    );
+    if (pluginDependencies.existsSync()) {
+      addDependency(pluginDependencies.uri);
+    }
+    final candidate = File.fromUri(
+      directory.uri.resolve(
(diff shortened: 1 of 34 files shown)


fzyzcjy:
Great job! Could you please let your agent follow the skills in the repo about updating cargokit to generate things and let it pass test CI

normalllll:
Since I only have a Linux environment, I am temporarily unable to get these CI tests to pass.

fzyzcjy:
Hmm could you please let your agent debug using the CI (and there is a feature to run ci manually and partially), and as soon as having time I will also let my agent have a check (but may not be very soon)

frb_codegen/src/library/integration/integrator/pubspec.rs
```diff
@@ -15,13 +15,28 @@ pub(super) fn pub_add_dependencies(
...
     // frb-coverage:ignore-end
     if integration_backend == IntegrationBackend::Cargokit {
-        match template {
-            Template::App => flutter_pub_add(
+        let (dependency_working_directory, hooks_local_path) = match template {
+            Template::App => (Some(Path::new("rust_builder")), "../../../frb_hooks"),
+            Template::Plugin => (None, "../../frb_hooks"),
+        };
+        flutter_pub_add(
+            &["code_assets:^1.0.0"],
```
fzyzcjy, line 23: [P1] This changes CargoKit's SDK contract for every platform. `flutter_rust_bridge_hooks` requires `hooks ^2.0.2`, whose SDK lower bound is Dart 3.10, while FRB still supports Dart 3.9.2 for OHOS and the CargoKit docs explicitly position this backend for older Flutter/Dart SDKs. Because these dependencies are added unconditionally, Android/OHOS-only and CocoaPods users can now fail during `pub get` before the macOS early return is ever reached. Please keep build-hook dependencies out of the legacy CargoKit path unless an SPM-capable Apple consumer actually needs them, and add resolver coverage at the lowest supported SDK.

frb_codegen/assets/integration_template/cargokit/shared/REPLACE_ME_CARGOKIT_HOOK_DIR/build.dart
```diff
@@ -0,0 +1,285 @@
...
+  );
+  final lockFile = File.fromUri(
+    outputDirectoryShared.resolve('frb_rust_toolchain.lock'),
+  );
+  await lockFile.create(recursive: true);
+  final lock = await lockFile.open(mode: FileMode.write);
+
+  try {
+    await lock.lock(FileLock.exclusive);
+    if (await toolchainFile.exists()) {
```
fzyzcjy, line 160: [P1] This branch makes existing CargoKit toolchain files opt into `native_toolchain_rust`'s stricter contract. A valid/common CargoKit `rust-toolchain.toml` such as `channel = "stable"` (or one without an explicit `targets` list) reaches the builder unchanged, but `native_toolchain_rust 1.0.4` rejects it before Cargo runs because it requires an exact channel and the current Apple triple. The no-toolchain fixtures only exercise the generated happy path. Please preserve the active CargoKit toolchain semantics while supplying the builder with an effective pinned channel/target, and add tests for existing stable and target-less toolchain files.

frb_codegen/assets/integration_template/cargokit/shared/REPLACE_ME_CARGOKIT_HOOK_DIR/build.dart
```diff
@@ -0,0 +1,285 @@
...
+  required String cratePath,
+  required String targetTriple,
+  required void Function(Uri) addDependency,
+  required Future<void> Function() action,
+}) async {
+  final crateDirectory = Directory.fromUri(packageRoot.resolve('$cratePath/'));
+  final toolchainFile = File.fromUri(
+    crateDirectory.uri.resolve('rust-toolchain.toml'),
+  );
+  final lockFile = File.fromUri(
```
fzyzcjy, line 152: [P1] The state being protected is `packageRoot.resolve(cratePath)/rust-toolchain.toml`, but the lock is keyed by this consumer-specific output directory. For a published/path plugin, two Flutter apps can build the same package or pub-cache crate with different `outputDirectoryShared` values, acquire different locks, and then overwrite/delete the same target-specific toolchain file. A read-only package cache fails immediately, and a crash during the in-place write can leave shared source truncated or polluted because cleanup relies on `finally`. Build hooks should treat `packageRoot` as read-only: please move the effective toolchain configuration into staging/output or extend the builder to accept it without mutating the crate. A crate-keyed lock plus atomic replacement would address only part of this problem.

frb_codegen/assets/integration_template/cargokit/shared/REPLACE_ME_CARGOKIT_HOOK_DIR/build.dart
```diff
@@ -0,0 +1,285 @@
...
+import 'dart:io';
+
+import 'package:code_assets/code_assets.dart';
+import 'package:flutter_rust_bridge_hooks/flutter_rust_bridge_hooks.dart';
+
+const _cratePath = 'REPLACE_ME_CARGOKIT_CRATE_PREFIXREPLACE_ME_RUST_CRATE_DIR';
+const _temporaryToolchainMarker =
+    '# Generated temporarily by flutter_rust_bridge Apple build hook.';
+
+void main(List<String> args) async {
```
fzyzcjy, line 11: [P2] This copies roughly 275 lines of generic Apple/SPM detection and Rust-toolchain orchestration into every generated project, even though the actual build is already owned by `flutter_rust_bridge_hooks`. These copies are effectively frozen: the overlay skips an existing `build.dart`, so fixes cannot reach users through a hooks-package upgrade or a later integrate run. The follow-up commits in this PR already had to patch the template plus all generated copies. Please move this orchestration behind a public builder/helper in `frb_hooks` and keep the template as a small `assetName`/`cratePath` adapter; that centralizes tests and removes most of this duplicated implementation.

fzyzcjy (changes requested):
AI-agent disclosure: this review was performed and submitted by an AI coding agent on Tom's behalf.

I reviewed the full PR delta with separate correctness/contract, lifecycle/concurrency, architecture/maintainability, adversarial integration, and build-time performance passes. I am requesting changes for the four inline findings below. The generated example files were treated as downstream copies of the integration templates, and no separate material build-time performance regression was found.

fzyzcjy:

fzyzcjy:
AI-agent disclosure: this review comment was prepared and posted by an AI coding agent on Tom’s behalf.

I performed a read-only static review of the current PR head and inspected the existing CI results. I did not run tests. The overall direction is reasonable, and the generated `Package.swift` linkage plus the SPM detection logic broadly match Flutter conventions. However, I do not think the PR is ready because the following issues remain:

1. **The build-hook dependencies are added to every CargoKit project.** `code_assets` and `flutter_rust_bridge_hooks` are added regardless of target platforms or whether the consumer uses SPM. This changes the SDK/dependency contract for Android-only, OHOS, older Dart, and CocoaPods-only users before the non-macOS hook guard can run. The existing Windows CargoKit CI already demonstrates a concrete `hooks ^1` versus `hooks ^2` solver failure.

2. **Existing CargoKit toolchain files are not compatible with the new builder contract.** When a crate already has `rust-toolchain.toml`, the hook passes it unchanged to `native_toolchain_rust`. Common CargoKit configurations such as `channel = "stable"` or a toolchain without an explicit current Apple target are valid for Cargo/rustup but are rejected by `native_toolchain_rust` before Cargo runs. The implementation should preserve the active CargoKit toolchain semantics while supplying an effective pinned channel and target to the SPM build path.

3. **The hook mutates package source with the wrong lock scope.** It creates or deletes `packageRoot/<crate>/rust-toolchain.toml`, while its lock is stored in consumer-specific `outputDirectoryShared`. Two Flutter projects consuming the same path or pub-cache package therefore take different locks but modify the same source file. Read-only package caches fail immediately, concurrent consumers can race, and abrupt termination can leave a truncated or stale toolchain file. Build hooks should treat `packageRoot` as read-only and put effective toolchain state in hook-owned staging/output.

4. **Too much orchestration is frozen into generated projects.** Roughly 285 lines of generic SPM detection and toolchain handling are copied into every generated hook. Existing hooks are skipped on later integration runs, so fixes cannot reach users through a `flutter_rust_bridge_hooks` upgrade. This logic should live behind a public helper/builder in `frb_hooks`, leaving a small generated adapter for `assetName` and `cratePath`.

The added overlay test verifies generated files and strings, but it does not exercise dependency resolution, existing toolchain files, read-only/shared package roots, or the actual hook behavior. These contract and ownership issues should be resolved before treating the PR as correct or merge-ready.

Issue #3067, opened 2026-05-07
Codegen-emitted .unwrap() panics are silent process kills under panic = "abort"


stepangfil:
## Summary

  The codegen-emitted `frb_generated.rs` uses `.unwrap()` on small but
  real failure paths (UTF-8 deserialization, integer parsing). Under
  default `panic = "unwind"` this is benign — FRB's own `catch_unwind`
  wrapper translates the panic into a Dart exception, which is the
  documented contract (see #1543).

  For projects that pin **`panic = "abort"`** in their release profile
  — a common setting for security-critical Rust applications following
  RustCrypto best practices — the `catch_unwind` safety net cannot fire
  because `abort` terminates the process before any unwind handler runs.
  The codegen `.unwrap()` calls then become silent process kills on
  malformed FFI input.

  This is not a defect of FRB under default settings. It is an
  incompatibility between two reasonable choices: FRB's design relies
  on unwinding; security-conscious downstream projects opt out of
  unwinding for their own reasons.

  ## Reproduction

  `flutter_rust_bridge = "2.12.0"`, `flutter_rust_bridge_codegen = "2.12.0"`.

  ```toml
  # Cargo.toml
  [profile.release]
  panic = "abort"
  ```

  After running `flutter_rust_bridge_codegen generate`, two `.unwrap()`
  sites appear in the generated FFI deserialization paths:

  ```rust
  // frb_generated.rs:222 — String UTF-8 deserialization
  impl SseDecode for String {
      fn sse_decode(deserializer: ...) -> Self {
          let mut inner = <Vec<u8>>::sse_decode(deserializer);
          return String::from_utf8(inner).unwrap();
      }
  }

  // frb_generated.rs:230 — integer parsing for u128
  impl SseDecode for u128 {
      fn sse_decode(deserializer: ...) -> Self {
          let mut inner = <String>::sse_decode(deserializer);
          return inner.parse().unwrap();
      }
  }
  ```

  If the Dart side ever passes malformed bytes (codegen bug, manual
  edit on one side without the other, FFI surface attack), the unwrap
  panics → `panic = "abort"` terminates the process immediately. Under
  default `panic = "unwind"` the existing FRB `catch_unwind` wrapper
  catches the panic and surfaces it as a Dart exception, so the
  problem is invisible.

  ## Why projects pin `panic = "abort"`

  A short list of reasons, all relevant to security and crypto code:

[...]

PR #3330, opened 2026-08-07
docs: note `panic = "abort"` behavior of generated code

stepangfil:
## Changes

Documents the `panic = "abort"` behavior of the generated code, per the discussion in #3067. Docs only, no code changes.

The "Return Types & Exceptions" page states that a Rust panic surfaces as a `PanicException` in Dart. That holds under Rust's default `panic = "unwind"`; a project that sets `panic = "abort"` opts out of the unwinding that makes it possible, so the process terminates at the panic site instead and nothing reaches the Dart side.

Two additions to that page:

* a short note next to the existing `PanicException` bullet, pointing at the details;
* a section at the end explaining the behavior, why the generated decoder `.unwrap()`s are **not** reachable in normal use (the generated encoder and decoder are two halves of the same codec, and Dart's own type invariants hold before anything is encoded), and the abnormal situations that do reach them — a codegen bug, hand-editing one side of the bridge without regenerating the other, or a caller invoking the native symbols directly and bypassing the generated Dart code.

Kept version-agnostic on purpose: no file line numbers, since those shift on every codegen run.

Addresses #3067, which was auto-closed as stale in the meantime.

### DIFF
--- a/website/docs/guides/types/translatable/return.md
+++ b/website/docs/guides/types/translatable/return.md
@@ -10,6 +10,12 @@ In addition, Rust has `panic` in addition to Result error, thus:
 
 * When Rust panic occurs, a `PanicException` will be thrown in Dart.
 
+:::note
+This is the behavior under Rust's default `panic = "unwind"`, which is what almost all projects use.
+If your project opts into `panic = "abort"` instead, panics cannot be turned into Dart exceptions —
+see [below](#panic-abort).
+:::
+
 If you want to see stack traces (backtraces), [this doc page](../../how-to/stack-trace) discusses how to configure it.
 
 ## Example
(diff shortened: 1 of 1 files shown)


website/docs/guides/types/translatable/return.md
```diff
@@ -10,6 +10,12 @@ In addition, Rust has `panic` in addition to Result error, thus:
 
 * When Rust panic occurs, a `PanicException` will be thrown in Dart.
 
+:::note
+This is the behavior under Rust's default `panic = "unwind"`, which is what almost all projects use.
+If your project opts into `panic = "abort"` instead, panics cannot be turned into Dart exceptions —
+see [below](#panic-abort).
+:::
+
```
fzyzcjy: nit: remove this diff

website/docs/guides/types/translatable/return.md
```diff
@@ -75,3 +81,36 @@ pub enum CustomStructError {
 ```
 
 As for how to fill it in or use it, you can refer to `thiserror` crate for some hints.
+
+## Projects using `panic = "abort"` {#panic-abort}
```
fzyzcjy: nit: make text concise

fzyzcjy (commented):
good job! just a bit of nits

fzyzcjy:
@all-contributors please add stepangfil for code

PR #3344, opened 2026-08-08
Limit automatic build runner outputs

fzyzcjy:
Close #3299

## Summary

- Restrict the automatic `build_runner` invocation to `.freezed.dart` and `.g.dart` outputs beneath the configured `dart_output` directory.
- Require `build_runner` 1.7.0 or newer because output filters were introduced in that release.
- Document that unrelated Dart builders must be run separately.

## Reproduction

- Baseline commit: `576ea02e11423196b4ee7309210f505547ba286c`
- Reproduction PR: #3342
- Baseline evidence: https://gist.github.com/fzyzcjy/497aede299e9e81de7648b6a15e31ace
- Steps:
  - Add an FRB complex enum that requires Freezed under `frb_example/dart_minimal`.
  - Add an unrelated JSON-serializable Dart model outside the configured `lib/src/rust` output directory.
  - Run FRB code generation.
- Observed before the fix:
  - The required `minimal.freezed.dart` output was generated.
  - The unrelated `unrelated_model.g.dart` output was also generated, proving that FRB ran builders across the whole Dart package.
  - Direct comparison in the same Docker environment took 20.160 seconds for an unfiltered cold run, 16.931 seconds for an unfiltered cached run, and 1.867 seconds for a filtered cached run.
- Expected after the fix:
  - Required Freezed/JSON outputs beneath `dart_output` are generated.
  - Outputs from unrelated builders outside `dart_output` are not generated by FRB code generation.
- Fix execution evidence: https://gist.github.com/fzyzcjy/1ad1e8593bea4485917227c8b9bc00af

The timing reported in #3299 remains environment- and project-dependent. This change removes the deterministic amplification from unrelated package builders; it does not eliminate `build_runner` cold-start cost.

## Validation

- Manual regression through the real FRB code-generation command: required Freezed output present; unrelated JSON output absent.
- Focused Rust regression: 4 passed, 0 failed.
- `./frb_internal test-rust-package --package frb_codegen`: 181 library tests, 13 binary tests, and 1 doc test passed.
- `./frb_internal test-dart-native --package frb_example/pure_dart`: all tests passed.
- `./frb_internal test-dart-native --package frb_example/pure_dart_pde`: all tests passed.
- `./frb_internal precommit-generate`: passed with no generated drift.
- `./frb_internal generate-internal-rust`: passed with no generated drift.
- `./frb_internal lint --fix`: passed; final worktree remained clean.
- Independent correctness review: no unresolved actionable findings.
[...]

Issue #2097, opened 2024-06-17
Run Dart build_runner never finishes running (internal loop?)


Larpoux:
### Describe the bug

when I code this : 
```rust
use flutter_rust_bridge::frb;
use web_audio_api::node::AudioNode as XAudioNode;
pub trait AudioNode
{
        #[frb(sync)]
        fn get_x_node(&self) -> &dyn web_audio_api::node::AudioNode ;

        #[frb(sync)]
        fn connect(&self, dest:&dyn AudioNode) ->&dyn AudioNode
        {
                let toto = self.get_x_node();
                let titi = dest.get_x_node();
                toto.connect(titi);
                dest
        }
}
```
FRB enters an infinite internal loop (never finishes).

[toto.log](https://github.com/user-attachments/files/15864867/toto.log)

### Steps to reproduce

-

### Logs

```shell
See attachment
```

### Generated binding code

```shell
-
```

### Version of `flutter_rust_bridge_codegen`

flutter_rust_bridge_codegen 2.0.0-dev.40

fzyzcjy:
Hmm that's weird. I guess firstly just execute that build_runner command manually in a terminal, and see whether it terminates or stuck forever. Another attempt may be wait for a few minutes to see whether it finishes (but is just very slow). It even takes >10 minutes for large Flutter projects IIRC.

fzyzcjy:
Oh I realize another thing (though seems unrelated to build_runner): Currently https://cjycode.com/flutter_rust_bridge/guides/lifetimes/overview is experimental and needs to use a flag to be enabled. (I think web-audio-api mostly does not need this but will double check.) Otherwise, its return type is translated to void.

Related: https://github.com/fzyzcjy/flutter_rust_bridge/issues/2083

Larpoux:
I closed most of my other tasks (IDE, ...) because sometimes I am low on memory.
```sh
larpoux@Mac-mini:~/projmac/tau$ RUST_LOG=debug flutter_rust_bridge_codegen generate >/tmp/toto.log
[1.3s] Parse  
  └── [1.3s] Cargo expand & syn parse  
  └── [0.0s] Parse HIR  
  └── [0.0s] Parse MIR  
[0.0s] Generate  
[526.7s] Polish ⠈
  └── [526.7s] Run Dart build_runner 
```                                                                                                                                                                                                                                                                                                                                                           
[...]

PR #3353, opened 2026-08-13
Stream build_runner output during codegen polish

dbsxdbsx:
## Changes

Polish used to buffer `build_runner` stdout/stderr until the child exited (`Command::output()`). A hung or slow run then only showed a spinner, even when `build_runner` had already printed the useful line (`freezed on N inputs`, an analyzer exception, etc.).

This streams that output line-by-line above the codegen progress bar, and still keeps the collected `Output` so a failed run can `bail!` with both stdout and stderr.

Only `dart_build_runner` opts in. `cargo expand` / `dart fix` / format stay buffered so they do not flood the console.

Related: #2097 (streaming this output was already suggested there).
Orthogonal to #3344: that PR scopes `--build-filter`; this one does not change what `build_runner` scans.
The two stack badly when both are missing — a long unscoped run then looks like a deadlock — but each fix stands on its own.

## Test plan

- [x] `cargo +nightly-2025-02-01 fmt --check` in `frb_codegen`
- [x] `cargo clippy -- -D warnings` in `frb_codegen` (same invocation as CI `lint-rust`)
- [x] `cargo test -p flutter_rust_bridge_codegen` (lib + bin + doc), including real `cmd`/`echo` spawn tests for `stream_output`
- [x] `cargo run -p flutter_rust_bridge_codegen -- generate` in `frb_example/dart_minimal`


fzyzcjy:
Good point! could you plz let agent follow the skills and make ci green

dbsxdbsx:
Pushed a follow-up for the Codecov patch miss (3 lines). Those were the format-on-fail arms of `assert!` in the streaming spawn tests, not the product path. The spawn/tee coverage itself was already there.

I kept the custom failure messages instead of collapsing those checks to a bare `assert!(cond)`. A passing run never takes the panic arm, but the message still matters when a check fails: on Windows, `cmd /c echo` can yield `
` or extra spaces, and the dump shows the actual stdout/stderr/error. Only those arms are marked `frb-coverage:ignore`.

`Lint :: Dart :: Primary` is a real CI break, not from this polish change: `pana` 0.23.18 stopped following package-root symlinks, so the isolated `frb_dart` copy no longer sees `README.md` / `CHANGELOG.md` (150/160, threshold 0). I considered two fixes: materialize those files before scoring, or pin the tool. Pinning `pana` to 0.23.17 looks like the safer one for now — it restores the last known-good score without changing package layout. Opened #3354 for the durable fix; this PR only keeps the pin.

The other first-run reds stay untouched:
- `Test :: Dart :: Web (pure_dart)`: existing web `RustStreamSink` completion hang (same family as #3311)
- Android quickstart smoke: emulator/adb + Gradle network
- macOS generate: rustup DNS while installing 1.93.1

New push will rerun the matrix.

fzyzcjy:
## Review findings

- **[P1] Preserve closed stdin in the streaming path** (`frb_codegen/src/library/commands/command_runner.rs:239-243`): `Command::output()` gives the child a closed stdin, while `Command::spawn()` inherits parent stdin by default. A `build_runner` / FVM / Dart subprocess that tries to read input can now consume terminal input or block indefinitely, which is a lifecycle regression directly opposed to the anti-hang goal. Set `cmd.stdin(Stdio::null())` before `spawn()`, and add a regression test whose child reads stdin and must terminate on EOF.

- **[P1] Do not drop live output when the progress target is hidden** (`frb_codegen/src/library/utils/console.rs:38-39`): indicatif 0.18 documents that `MultiProgress::println()` does nothing when its draw target is hidden. The default stderr target is hidden when stderr is not a terminal, so CI, redirected logs, and many IDE/task runners still receive no live `build_runner` diagnostics; a hung successful child never reaches the later buffered logging. Fall back to a direct stderr write when `MULTI_PROGRESS.is_hidden()` (or use another path that always writes), and cover the hidden-target case.

- **[P2] Remove the stale `pana` changelog entry** (`CHANGELOG.md:11`): the net PR diff says CI pins `pana` to 0.23.17, but `tools/frb_internal/lib/src/makefile_dart/lint.dart` now matches `master` and activates unversioned `pana` after follow-up commit `5a903bf0f7`. The release note no longer describes the shipped behavior. Remove this line, or restore the pin as a separate complete change.

## Review conclusion

- **Independent passes**: semantics/tests, lifecycle/concurrency, architecture, performance, and an adversarial end-to-end pass.
- **Dismissed candidates**: no substantive thread or memory overhead; partial-line buffering and stdout/stderr interleaving match the stated line-oriented scope.
- **Changes made during review**: none.
- **Status**: not ready until the findings above are resolved; current CI checks are also still pending.

dbsxdbsx:
Thanks for the review. All three findings are addressed:

- **[P1] Closed stdin on the streaming path:** `0255f8d`
- **[P1] Live output when the progress target is hidden:** `0255f8d`
- **[P2] Stale pana changelog line:** `490b5c2`

frb_codegen/src/library/commands/dart_build_runner.rs
```diff
@@ -16,6 +16,7 @@ pub fn dart_build_runner(dart_root: &Path, fvm_install_mode: FvmInstallMode) ->
     let out = command_run!(
         call_shell[Some(dart_root), Some(ExecuteCommandOptions {
             envs: Some(dart_run_extra_env()),
+            stream_output: true,
```
fzyzcjy: (human comment)

@dbsxdbsx wondering maybe we can do things like

1. in normal mode, do not stream logs; if it runs too long, just let the spinner have a message saying sth like "if you want to see logs, use verbose mode"
2. in verbose mode (--verbose), stream all command outputs, not only this one

dbsxdbsx: - Agree: default stays quiet; `--verbose` streams every child command, not only `build_runner`.
- Hint on the spinner from the start, e.g. `Run Dart build_runner (use --verbose for logs)` — no timeout. A fixed delay is hard to pick, and a late hint still cannot show logs for the current run.
- No new flag, and no need to thread `verbose` through the library. `--verbose` / `RUST_LOG=debug` already raise the log level; `execute_command` can stream iff `log::max_level() >= Debug`, and we can drop the hardcoded `stream_output: true` here.

What do you think?  @fzyzcjy

fzyzcjy: LGTM

dbsxdbsx: Thanks, implementing this next.

fzyzcjy:
(feel free to ping me when done)

dbsxdbsx:
@fzyzcjy  it is done.

fzyzcjy:
looks great! will have a checkprobably this weekend

fzyzcjy (approved):
LGTM, todo(fzyzcjy): add realistic UT

fzyzcjy (approved):

fzyzcjy (approved):

Issue #2917, opened 2025-11-03
Support `onCancel` callback for `StreamSink`.


623637646:
### 📝 Description

Currently, [`StreamSink`](https://cjycode.com/flutter_rust_bridge/guides/types/translatable/stream) in `flutter_rust_bridge` supports sending events from Rust to Dart via streams, but there’s no mechanism for Rust to be notified when the Dart-side subscription is cancelled — for example, when a user calls `subscription.cancel()` or when the Dart stream is disposed.

This limitation can lead to **resource leaks** if the Rust side holds long-lived resources such as sockets, threads, or file handles that should be cleaned up once the Dart listener stops.

---

### 🚀 Proposed Feature

Add support for an **`onCancel` callback** in the `StreamSink` type.  
When the Dart subscription is cancelled, the Rust side’s callback should be invoked automatically.

623637646:
Just tried this example code from https://cjycode.com/flutter_rust_bridge/guides/types/translatable/stream

```rust
use anyhow::Result;
use std::{thread::sleep, time::Duration};

use crate::frb_generated::StreamSink;

const ONE_SECOND: Duration = Duration::from_secs(1);

// can't omit the return type yet, this is a bug
pub fn tick(sink: StreamSink<i32>) -> Result<()> {
    let mut ticks = 0;
    loop { // <<<<< This loop never stop
        sink.add(ticks);
        sleep(ONE_SECOND);
        if ticks == i32::MAX {
            break;
        }
        ticks += 1;
    }
    Ok(())
}
```

The `loop` never stop even if the `MyHomePage` (The widget which using `tick`) is gone. 
Without the `onCancel callback for StreamSink` feature, How do we handle this case?

rkudiyarov:
We have faced the same issue! Looking forward to way to know that the stream is closed on the rust side.

fzyzcjy:
> Without the onCancel callback for StreamSink feature, How do we handle this case?

One potential way may be explicitly call a callback, e.g.

```
pub struct MyTicker { is_running: AtomicBool }

impl MyTicker {
  pub fn create(sink) {
    while self.is_running { sink.add }
  }

  pub fn cancel() { self.is_running = false; }
}
```

Issue #2339, opened 2024-10-05
StreamSink impl async fn cancelled()


modeckrus:
I use spawn to send something in stream via StreamSink, and I want to close this green thread when dart stream is closed. So I suggest to add method for StreamSink cancelled like in https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html

fzyzcjy:
Looks reasonable and feel free to PR for this new feature!

As a workaround before it is implemented, you can explicitly call a Rust function marking it is closed when that happens.

PR #3360, opened 2026-08-15
Preserve Web StreamSink ordering across workers

fzyzcjy:
base #3366

## Summary

- Preserve Rust-to-Dart Web `StreamSink` ordering when data and the final close originate from different workers.
- Release worker-local `BroadcastChannel` caches only after Dart has drained the ordered close, with a bounded fallback for subscriptions canceled before close delivery.
- Keep the cancellation, pause, and resume behavior introduced by #3311.

## Root cause

`BroadcastChannel` preserves order for one sender, but not across different sender objects. A worker that dropped the final `StreamSink` could therefore deliver close before data already posted by another worker. Dart closed the stream and discarded the late data.

The transport also needed a lifecycle protocol spanning every participating worker. Immediate Rust-side cleanup could race with queued delivery, while acknowledgement-only cleanup leaked worker-local channels when Dart canceled before receiving close.

## Fix

- Allocate a shared wasm sequence for every data, error, and close message.
- Reorder messages in Dart before decoding; carry failed sequence numbers so a failed post cannot create a permanent gap.
- Retry a failed final close through a fresh channel.
- Mark final close for release and let Dart acknowledge only after all preceding messages have drained.
- Broadcast cache release to every worker after acknowledgement.
- Schedule a generation-separated fallback release after final drop so canceled streams are cleaned up without shortening any stream's delivery grace period.
- Use a page-session nonce in channel names to avoid same-origin page collisions.
- Implement ordering with a synchronous transformer so cancel, pause, and resume propagate directly.

## Regression evidence

- Intentional-red reproduction: #3356
- Focused failing run: https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/31874840364
- Failing job: https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/31874840364/job/94990448752
- The regression runs 100 streams with 1000 exact ordered values each. The old implementation truncates generated variants, including first mismatches at indices 996 and 911 in CI and 908 and 985 locally.

## Local validation

- `cargo test -p flutter_rust_bridge`: 41 unit tests and 3 integration tests passed.
- `cd frb_dart && dart test`: 84 tests passed.
- `cargo check -p flutter_rust_bridge --target wasm32-unknown-unknown`: passed.
- Full `pure_dart` Web suite: 4220 passed.
- Full `pure_dart_pde` Web suite: 2003 passed.
- `./frb_internal precommit-generate`: passed with no generated drift.
- `./frb_internal lint --fix`: format, Rust fix/clippy, Dart format/fix, and static analysis passed with no drift. The final pana package score remains 150/160 only because its temporary package lacks `README.md` and `CHANGELOG.md`; static analysis is 50/50.

## Focused CI

`test_dart_web[package=frb_example--pure_dart]`


dbsxdbsx:
(I don't personally feel qualified to review this at this depth, but since you asked: I ran it through three different AI agents and then edited the result. What follows is that write-up.)

I took a look at this (as you asked on #3311). The bug is real, and numbering + reordering on the Dart side is the right *direction* for cross-worker BroadcastChannel.

I would not land this shape, though. The PR turned an ordering fix into a reliable-delivery protocol (readiness handshake, skipped-seq carry, close retry on a fresh channel, ACK-only cache release). That is a lot of machinery, and a few holes are still open:

- After cancel on Web, the consumer future completes, but the underlying channels stay up until Rust sends a release-marked close. For an idle StreamSink (the #3311 case) that close never comes, so the transport can leak.
- Until ready, posts are structured-cloned into an unbounded `pending_messages` queue and reported as success. A sync burst (e.g. 1000 adds) piles up before the JS timer can run.
- A sequence can be allocated and then never become DATA / SKIP / ABORT (queued-but-never-flushed, or both close attempts fail). Dart then waits forever for `next`.
- The description still mentions a bounded/generation fallback; that path is gone from HEAD, while the readiness protocol is not described.

A narrower cut would keep sequence+reorder, give every allocated seq a terminal state, bound or eliminate the ready queue, and let Dart cancel *abort* the transport instead of waiting for Rust to drop.

One related constraint, not a design proposal: #2339 / #2917 / #2978 still want Dart cancel to notify Rust (whatever the public API). I am not asking to build that here. I only flag that the current “cancel, but keep the pipe open until Rust closes” path will make that feature harder, and it is the same idle-cancel situation as #3311.

Sketch of the narrower data flow I have in mind (solid = this PR; dashed = later, API TBD). This is not a cancel-token design.

```mermaid
flowchart LR
  WA[Worker A<br/>add] --> SEQ[Shared seq]
  WB[Worker B<br/>close] --> SEQ
  SEQ -->|data / error / close + seq| CH[BroadcastChannel]
  CH --> ORD[Dart reorder]
  ORD -->|in-order| US[User Stream]

  US -->|cancel| AB[Abort]
  AB -->|stop waiting<br/>drop local ports| ORD
  AB -.->|later: tell Rust to stop<br/>#2978 API TBD| RT[Rust task]

  ORD -->|after close drained<br/>ACK| CACHE[Worker channel caches]
```

Happy to go into any of the bullets in more detail if useful.

fzyzcjy:
looks reasonable! will have a deeper check asap hopefully this weekend

frb_dart/lib/src/generalized_isolate/_web.dart
```diff
@@ -40,27 +47,201 @@ class ReceivePort extends Stream<dynamic> {
...
+          onDone: () {
+            if (!consumerCanceled) controller.close();
+            scheduleMicrotask(closeSource);
+          },
+        );
+
+    controller
+      ..onPause = sourceSubscription.pause
+      ..onResume = sourceSubscription.resume
+      ..onCancel = () {
```
fzyzcjy, line 122: Cancellation now only detaches the consumer; it keeps the ordered browser subscription alive and continues sequencing every later message until Rust eventually sends a release-marked close. A valid long-lived producer can retain the StreamSink or ignore cancellation, so that close may never arrive. Each canceled stream then permanently retains its Dart controller/listener and BroadcastChannels, while continuing to spend CPU on messages that are discarded. The new tests always inject a close after cancellation, so they do not cover this lifetime. Can we make cancellation an explicit Rust/Dart release handshake (with a terminal acknowledgement) rather than depending on eventual StreamSink drop?

fzyzcjy (commented):
agent

---

Found one blocking lifecycle/resource issue in the cancellation path; see the inline comment.

PR #3367, opened 2026-08-16
feat(codegen): generate Dart operators from Rust traits

star4277:
## Summary

- Generate Dart operator overloads from Rust standard operator traits.
- Route generated operators through Rust bridge calls, including `PartialEq` and `PartialOrd`.
- Preserve existing Dart equality behavior when `PartialEq` is not implemented, while using Rust equality when it is implemented.
- Add operator mapping documentation, generated examples, and regression tests.

## Verification

- `cargo check -p flutter_rust_bridge_codegen`
- `cargo clippy -p flutter_rust_bridge_codegen -- -D warnings`
- `cargo fmt --all -- --check`
- `cargo test -p flutter_rust_bridge_codegen --lib -- --test-threads=1` (172 passed)
- Dart operator tests and `dart analyze` for the generated examples
- Regenerated and checked `D:\Projects\myapp`


fzyzcjy:
good point! could you plz let agent pass all ci tests

PR #3374, opened 2026-08-22
Make Dart Valgrind failures a hard gate

fzyzcjy:
## What

- Preserve Valgrind’s nonzero exit status instead of opting out of command failure checks.
- Validate both stdout and stderr, where Valgrind writes its diagnostics.
- Keep the existing test-completion marker and definite/indirect leak checks unchanged.
- Add regression coverage for clean output, stderr-only leaks, nonzero exit, and incomplete tests.

## Reproduction

The intentional-red reproduction is #3370. Its focused CI job failed on the stderr-only definite leak as expected after the regression assertion was added:

- https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/32558387754/job/96996100045

Before this fix, the production wrapper passed only stdout to the leak checker and used `checkExitCode: false`, so the same Valgrind failure could be reported as green.

## Validation

- `dart test test/makefile_dart_test.dart`: 56 tests passed.
- `dart analyze`: passed for `tools/frb_internal`.
- `dart format`: passed for the changed Dart files.
- Real `test-dart-valgrind` now fails on the repository’s existing Valgrind diagnostics instead of silently accepting them; this is the intended hard-gate behavior and exposes follow-up memory cleanup work.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -931,7 +931,7 @@ jobs:
           architecture: x64
       - run: sudo apt-get update
 
-      # execute
+      # execute: Valgrind exit status and both output streams are hard gates
       - run: ./frb_internal test-dart-valgrind --package ${{ matrix.package }}
 
   test_dart_sanitizer:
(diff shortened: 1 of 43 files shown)


fzyzcjy (commented):
LGTM but need to pass ci

fzyzcjy (commented):
LGTM

fzyzcjy:
diff LGTM

PR #3395, opened 2026-08-23
Use Dart pub for CargoKit build tools

fzyzcjy:
## Changes

- Run nested CargoKit build-tool packages with Dart pub instead of inheriting the parent package mode.
- Keep the change independent of the Flutter 3.47 upgrade.
- This is the first predecessor in the stack.

## Validation

- `dart analyze`
- `dart test test/makefile_dart_test.dart`

## CI

- `ci-manual-dispatch` is intentional because focused local coverage is sufficient for this one-line mode correction.


fzyzcjy (commented):
LGTM

PR #3396, opened 2026-08-23
Complete from-scratch generation

fzyzcjy:
## Changes

- Remove all checked-in Dart generation outputs before regenerating examples from scratch.
- Cover builder outputs as well as the existing generated bridge outputs.
- This PR is stacked on #3395 and remains independent of the Flutter 3.47 upgrade.

## Validation

- `dart analyze`
- focused `makefile_dart_test.dart` coverage

## CI

- `ci-manual-dispatch` is intentional because this predecessor has focused local coverage.

PR #3397, opened 2026-08-23
Restore build_cli state around internal generation

fzyzcjy:
## Changes

- Validate all target pubspecs before changing any of them.
- Enable `build_cli` only for internal generation, then restore the exact original contents on success or failure.
- Format each generated package in the same lifecycle.
- Keep the lifecycle helper in a dedicated module without production `ForTesting` entry points.
- This PR is stacked on #3396 and remains independent of the Flutter 3.47 upgrade.

## Validation

- `dart analyze`
- `dart test test/makefile_dart_test.dart` (60 tests)
- full `tools/frb_internal` test suite
- `generate-internal --set-exit-if-changed` reached the repository contributor-count check with a clean worktree; that unrelated existing check rejected the current contributor data.

## CI

- `ci-manual-dispatch` is intentional because this predecessor has focused local coverage.


fzyzcjy (commented):
LGTM

PR #3383, opened 2026-08-23
Split Dart sanitizer tester

fzyzcjy:
## Summary

- Split the Dart sanitizer tester into SDK, runner, and metadata modules while preserving its behavior.
- Activate the tested macOS quickstart package's absolute Debug app bundle, including the native-assets example.
- Preserve screenshot OCR, readiness, timeout, and failure checks. Cover both package names and paths containing spaces.

## Reproduction

- Baseline: `60b3f2ff2b70d33802e16ceb30af808e311307f8` on the macOS Intel CI runner.
- Command: `./frb_internal test-flutter-quickstart-smoke --package frb_example--flutter_via_create_native_assets --target desktop --device-id macos`.
- [The failed job](https://github.com/fzyzcjy/flutter_rust_bridge/actions/runs/33958642030/job/101290129655) built `flutter_via_create_native_assets.app` but tried to activate `flutter_via_create`; the screenshot OCR did not contain `Hello, Tom`.
- Expected: activate the app actually built, then require the same screenshot and OCR evidence.

## Validation and stack

- Head: `f6ba0db01f2deb03e925aac97136625c97f1658c`, targeting `master`.
- Removed the independent Web prerequisite and the requested contributing-overview paragraph. No smoke implementation or assertions were removed.
- Per-commit restack audit preserved all original changes except that explicitly requested paragraph removal. Independent correctness and test-scope reviews found no unintended changes.
- On descendant stack tip `4b36907e1f4187331d62f835b867bf848e79eac2`, Docker `lint --fix` passed; Dart runtime passed 155 tests and internal tools passed 215 tests. These are combined-stack results, not a separate execution of this PR head.
- Actual macOS screenshot/OCR validation remains outside this local run. No full current-head CI pass is claimed.

PR #3419, opened 2026-08-30
Close response ports when native calls throw

fzyzcjy:
## Change

- A normal call creates its response port before invoking `task.callFfi`. If that invocation throws synchronously, no reply arrives to trigger the port's usual cleanup.
- Return the port's explicit close callback, close on synchronous invocation failure, and rethrow the original exception.
- Preserve first-response completion and type-error propagation; cover closing before any response.

## Validation and stack

- Head: `ca8597dc2ce2dfeec0f546a7eb6ecbc48a58fded`, based on PR #3383.
- The independent Web receiver/StreamSink prerequisite has been removed. This change uses the existing ordinary response-port API.
- The PR's added and removed content is identical before and after restacking. Independent correctness and test-scope reviews found no unintended changes.
- On descendant stack tip `4b36907e1f4187331d62f835b867bf848e79eac2`, Docker `lint --fix` passed; the Dart runtime suite passed 155 tests and internal tools passed 215 tests. These are combined-stack results, not a separate execution of this PR head.
- Full current-head CI is not yet certified. This change does not add a timeout for calls that neither throw nor reply.


fzyzcjy:
The added cleanup does not normally close the response port twice:

- On success, the existing reply handler closes the port before completing the result.
- If `task.callFfi(...)` throws synchronously, the new `catch` closes the port and rethrows the original error.
- Even if Rust has already posted a reply, the asynchronous Dart port handler runs after the current synchronous call unwinds. Closing the native port first prevents that handler from receiving the reply: [RawReceivePort.close documentation](https://api.dart.dev/dart-isolate/RawReceivePort/close.html).

Repeated closure is also safe if it does occur:

- Dart VM explicitly returns without doing anything when the receive port is already closed: [Isolate::CloseReceivePort](https://github.com/dart-lang/sdk/blob/main/runtime/vm/isolate.cc#L3609-L3623).
- The Web implementation of this response port uses `MessagePort.close()`. Its specified steps set the detached flag and disentangle the port only if it is still entangled, so repeating the operation is safe: [HTML Standard](https://html.spec.whatwg.org/multipage/web-messaging.html#dom-messageport-close).

An additional `closed` flag is therefore unnecessary for this cleanup. Losing subsequent replies after the synchronous exception is intentional: the call has already failed, and its unused response port should be released.

fzyzcjy (commented):
LGTM

PR #3431, opened 2026-09-05
Support Dart Wasm output in build-web

fzyzcjy:
## Summary

- Add Dart Wasm output support to `build-web`.
- Update wasm-bindgen, worker, gallery bootstrap, CLI help, and focused build-web tests.

## Validation

### DIFF
--- a/frb_dart/lib/src/cli/build_web/executor.dart
+++ b/frb_dart/lib/src/cli/build_web/executor.dart
@@ -7,6 +7,34 @@ import 'package:flutter_rust_bridge/src/cli/cli_utils.dart';
 import 'package:flutter_rust_bridge/src/cli/run_command.dart';
 import 'package:meta/meta.dart';
 
+const _cargoLlvmCovEnvKeys = [
+  'LLVM_PROFILE_FILE',
+  '__CARGO_LLVM_COV_RUSTC_WRAPPER',
+  '__CARGO_LLVM_COV_RUSTC_WRAPPER_RUSTFLAGS',
+  '__CARGO_LLVM_COV_RUSTC_WRAPPER_CRATE_NAMES',
+  'RUSTC_WRAPPER',
+  'CARGO_LLVM_COV',
+  'CARGO_LLVM_COV_SHOW_ENV',
+  'CARGO_LLVM_COV_TARGET_DIR',
+  'CARGO_LLVM_COV_BUILD_DIR',
+];
+
+/// Command runner used by build-web, injectable by tests.
+@visibleForTesting
+typedef BuildWebCommandRunner =
+    Future<RunCommandOutput> Function(
+      String command,
+      List<String> arguments, {
+      String? pwd,
+      Map<String, String>? env,
+      bool shell,
+      bool silent,
+      bool? checkExitCode,
+      bool printCommandInStderr,
+      List<String> removedParentEnvKeys,
+      Duration? timeout,
+    });
+
 /// {@macro flutter_rust_bridge.cli}
 class BuildWebArgs {
   /// {@macro flutter_rust_bridge.cli}
@@ -37,6 +65,9 @@ class BuildWebArgs {
   final String? dartCompileJsEntrypoint;
 
   /// {@macro flutter_rust_bridge.cli}
+  final String? dartCompileWasmEntrypoint;
+
+  /// {@macro flutter_rust_bridge.cli}
   final List<String> features;
 
   /// {@macro flutter_rust_bridge.cli}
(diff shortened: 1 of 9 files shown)


frb_example/gallery/web/index.html
fzyzcjy, line 1: give evidence

frb_dart/lib/src/cli/build_web/executor.dart
```diff
@@ -50,6 +81,7 @@ class BuildWebArgs {
     required this.wasmPackRustupToolchain,
     required this.wasmPackRustflags,
     required this.dartCompileJsEntrypoint,
+    this.dartCompileWasmEntrypoint,
```
fzyzcjy: make it required?

frb_example/gallery/web/index.html
```diff
@@ -31,29 +31,8 @@
...
-        },
-        onEntrypointLoaded: function(engineInitializer) {
-          engineInitializer.initializeEngine().then(function(appRunner) {
-            appRunner.runApp();
-          });
-        }
-      });
-    });
-  </script>
+  <script src="flutter_bootstrap.js" async></script>
```
fzyzcjy, line 36: This matches the current Flutter web scaffold. Flutter’s official `index.html.tmpl` uses exactly `<script src="flutter_bootstrap.js" async></script>` and links to the official initialization guide for customization: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/templates/app/web/index.html.tmpl. The guide documents that `flutter build web` generates `flutter_bootstrap.js`, which initializes `_flutter.loader` and calls its `load` API: https://docs.flutter.dev/platform-integration/web/initialization. Using the generated bootstrap instead of the old hand-written `loadEntrypoint` flow therefore follows Flutter’s current template and lets Flutter select the artifacts produced by the chosen web compiler.

fzyzcjy (commented):
Evidence for the bootstrap update.

frb_dart/lib/src/cli/build_web/executor.dart
```diff
@@ -60,47 +92,71 @@ extension on BuildWebArgs {
...
+
+  String get outputDartWasm => '$output/main.dart.wasm';
 }
 
 /// {@macro flutter_rust_bridge.cli}
-Future<void> executeBuildWeb(BuildWebArgs args) async {
-  await _sanityChecks(args);
+Future<void> executeBuildWeb(
+  BuildWebArgs args, {
+  @visibleForTesting BuildWebCommandRunner runCommandImpl = runCommand,
```
fzyzcjy, line 102: if runCommandImpl is only for testing, rm it to simp code

fzyzcjy: I would keep this injection point. `executeBuildWeb` orchestrates several external commands (`which`, `cargo read-manifest`, `wasm-pack`, optional `wasm-bindgen`, and Dart compilation). The focused executor tests use this runner to capture and assert the exact command sequence without installing or executing those tools. Removing it would either make the unit tests environment-dependent integration tests or require a broader command-runner abstraction, which would add more code than this single optional seam.

frb_dart/lib/src/cli/build_web/executor.dart
```diff
@@ -193,6 +254,9 @@ Future<void> _executeWasmPack(
       'RUSTFLAGS': rustflagsResolution.rustflags,
       if (stdout.supportsAnsiEscapes) 'CARGO_TERM_COLOR': 'always',
     },
+    removedParentEnvKeys: Platform.environment['CARGO_LLVM_COV'] == '1'
+        ? _cargoLlvmCovEnvKeys
+        : const [],
```
fzyzcjy, line 259: why this?

fzyzcjy: `cargo llvm-cov` runs the outer test process with coverage-specific Cargo variables such as `LLVM_PROFILE_FILE`, `__CARGO_LLVM_COV_RUSTC_WRAPPER*`, and `RUSTC_WRAPPER`. `wasm-pack` then launches a nested Cargo build for `wasm32-unknown-unknown`; inheriting the native coverage wrapper/profile configuration contaminates that unrelated Wasm build and can make it fail. We remove the coverage variables only when the explicit `CARGO_LLVM_COV=1` marker is present. The conditional is important: without that marker, user configuration such as `RUSTC_WRAPPER=sccache` remains untouched. The follow-up on #3432 extracts this decision into a pure helper and covers both cases deterministically.

fzyzcjy (commented):
LGTM

PR #3432, opened 2026-09-05
Isolate wasm-pack from cargo llvm-cov

fzyzcjy:
## Summary

- Remove cargo-llvm-cov wrapper variables before invoking wasm-pack.
- Preserve ordinary compiler wrappers such as sccache outside cargo-llvm-cov runs.

## Validation


fzyzcjy (commented):
LGTM

PR #3453, opened 2026-09-12
Expand macros in build.rs using cbindgen-inspired isolation and recursion guards

fzyzcjy:
- `build.rs` previously parsed unexpanded source, so APIs declared by Rust macros disappeared from generated bindings. Use real compiler expansion for both CLI and build-script generation.
- **Reference: cbindgen.** Adapt cbindgen's [isolated expansion target directory](https://docs.rs/cbindgen/latest/src/cbindgen/bindgen/cargo/cargo_expand.rs.html#75-91) and [recursive-generation guard](https://github.com/mozilla/cbindgen/blob/main/src/bindgen/builder.rs): use `OUT_DIR/frb-expand` for child Cargo and `_FRB_CODEGEN_IS_RUNNING` to stop nested generation before configuration parsing or writes. Log the early return. Normal CLI generation retains its existing target directory.
- Preserve encoded Rust flags, append `frb_expand`, and forward the build-script target. For the build script's own package, inherit its active Cargo features and disabled-default state. Explicit `rust_features` retains precedence; other packages do not inherit the caller's features. Read original names from `CARGO_CFG_FEATURE`, with a manifest-aware fallback for older Cargo and an error for ambiguous legacy names.
- **Validation:** 14 cargo-expansion tests passed locally in Docker, including actual Rust compiler expansion with default, optional, and disabled features. The real `dart_build_rs` logging e2e still passes. Earlier validation also covered the full codegen tests, clean Cargo build, bootstrap generation, encoded flags, recursion guard, and CLI/build.rs convergence.
- **Review:** independent correctness and test-weakening reviews found no remaining actionable issue in feature inheritance.
- **Chain:** first of two PRs; #3454 supplies end-to-end logging macro coverage. The example is unchanged in this predecessor. Supersedes #3414. `ci-manual-dispatch` intentionally avoids running the full matrix on this predecessor until it is next to land.
- **Final local verification:** all 660 codegen library tests, 13 CLI tests, and 1 doc test passed. CLI generation of `dart_minimal` and `pure_dart` converged with no file changes. Repository native/wasm Rust lint passed. Disabling feature inheritance makes the compiler-expansion regression test fail on the unexpected default API.


fzyzcjy (commented):
LGTM

PR #3465, opened 2026-09-13
Add security audit workflow

0rzech:
## Changes

Adds security audit github workflow and updates and/or replaces packages reported by `cargo deny check advisories`.


fzyzcjy:
good job! could you please let agent follow the skills in repo to make ci green, and also maybe your workflow can be triggered per commit in ci.yaml

0rzech:
The workflow will run for each commit push that changes the dependencies. It will run scheduled on master too, because advisories may appear with time without dependency changes.

On the other hand, it doesn't make much sense to run it for each and every commit and fail unrelated ones because of that. Also, triggering it from another workflow would complicate things due to Cargo.lock and Cargo.toml filters.

But if you still want to do that, just let me know.

What should I do with regards to the agent?

0rzech:
I've fixed the errors, but it looks like the tests are flaky. [Here](https://github.com/0rzech/flutter_rust_bridge/actions/runs/34776355834/job/103775142862) and [here](https://github.com/0rzech/flutter_rust_bridge/actions/runs/34776620329) you can see successful runs for the same commit.

fzyzcjy:
hmm e.g. I do not see security audits triggered here (maybe due to not changing related things)

and I am worried nobody will realize it fails hidden in the workflow if it is not on main / every commit

given this looks lightweight, I still guess it is good to make it per-commit into ci.yaml

image

fzyzcjy:
no worries about the flaky ones, and btw could you plz let agent list a table about all deps you changed, especially e.g.  changing deps, w/ reasons

0rzech:
You do not see security audit being run, because it must be first merged into master. It will run and be visible once it is merged into master.

The audit _will_ run for every commit, but only if the dependencies (`Cargo.lock`, `Cargo.toml`) were changed or if the audit was triggered by a schedule.

I genuinely don't know what I am supposed to do to allow the agent to work.

0rzech:
Btw. I forgot to add `**/deny.toml` to push trigger paths in the workflow. Are you ok. with me fixing that and triggering CI workflow again?

0rzech:
So, the current workflow is as follows:
- If a developer or agent changes `Cargo.lock`, `Cargo.toml` or `deny.toml` on any branch, the workflow will run on that commit and check whether everything is still fine and in accordance to `deny.toml`.
- On 01:30 every day, the workflow will run on master branch's tip and check whether everything is still fine (i.e. no vulnerabilities were discovered for existing dependencies).
- If the workflow fails, it will be obvious why and it won't be lost in the long list of all the jobs.
- If a user does not change `Cargo.lock`, `Cargo.toml` or `deny.toml`, the workflow will not run, because it would fail unrelated commits and force unrelated PRs to include dependency changes, and bumping dependencies is not exactly trivial in this project. Given, that the workflow will run each day on master, you will be notified about new RUSTSEC CVEs anyway, and you can always create a dedicated PR to fix those.

If you still want to run the workflow for each commit and embed it into CI workflow:
- Scheduled workflow file should stay anyway to be run each day in search for new vulnerabilities of existing dependencies.
- I'd have to add yet another job to an already pretty huge and complex CI workflow for possibly no benefit other than prefixing the name with `CI`, but I may as well be missing some context here.
- I will do it, because it's your project and your choice after all. 🙂

fzyzcjy:
ok no worries, then I will check this PR in more details and merge later! (maybe next weekend if having time)

0rzech:
Alright, I've added security audit job to CI workflow. 🙂

0rzech:
Could you rerun the failed job? I have run the `CI` workflow with `test_dart_sanitizer[package=frb_example--pure_dart]` filter on the same commit and it went fine thrice:
* https://github.com/0rzech/flutter_rust_bridge/actions/runs/35732603334
* https://github.com/0rzech/flutter_rust_bridge/actions/runs/35734230505
* https://github.com/0rzech/flutter_rust_bridge/actions/runs/35739441942

In general, the CI seems to be rather flaky overall. For instance, I shouldn't have to commit 5600e3a6512b984350d604df53db0a8f5cfbb898, because the only Dart code I touched was in `frb_internal` and wihout changing any dependencies, and yet the CI failed, because the job bumps dependencies and diffs them against the PR which. Also, random jobs seem to fail randomly. This is not to diss the project - it's a humble suggestion on what is problematic from contributor's point of view.

fzyzcjy:
looks like the CI is green (except for the one which I also think is flaky) :)

yes the CI is sometimes flaky, it is best if it could be enhanced, and feel free to PR if you have interest!

0rzech:
Interestingly enough, the flaky one failed twice in a row when run in the context of your repo. 😂 Could you try just one more time?

As for the general CI reliability, thanks for the invitation! I can't promise I'll be able to fix it, but for sure I don't want to attempt to do this in the current PR. 😉

PR #3467, opened 2026-09-18
fix(codegen): custom ser/des is ignored for types with a built-in delegate

### LINKED ISSUE #2212: Add Rust Ipv4Addr, Ipv6Addr to Dart InternetAddressType class
I found, that there is no default rust to dart conversion for Network types Such as Ipv4Addr, Ipv6Addr to dart InternetAddressType since frb converts it to opaque struct. Can you add this conversion by default in frb?
[Rust IpAddr](https://doc.rust-lang.org/std/net/enum.IpAddr.html)
[Dart InternetAddressType](https://api.flutter.dev/flutter/dart-io/InternetAddressType-class.html)

boxdot:
compute_matcher_types renders the Rust API type fully qualified (uuid::Uuid),
but the caller matches it against the bare path segment (Uuid), so a custom
rust2dart/dart2rust registered for such a type never matches and the built-in
delegate silently wins. Match the last segment too.

Closes #2212, where this surfaced as Ipv4Addr and was closed undiagnosed.


fzyzcjy:
Great job! could you please let your agent follow the skills in this repo to make CI green :)

boxdot:
I am not sure how to fix CI. And I am a human.

